PR #9, opened 2021-01-26
Automate reference.txt -> md with reference_to_md.py

ysthakur:


ysthakur:
I would suggest squashing and merging since the second commit is pretty trivial.

PR #13, opened 2021-04-07
Destroy the legends

ysthakur:


ysthakur:
Remove the legends part from the elements.md thing

PR #44, opened 2021-05-03
Fix innerHTML -> value

hyper-neutrino:
the markdown format buttons at the top of the page don't work because they are trying to set the innerHTML of a textarea


hyper-neutrino:
never mind, this version is behind

PR #60, opened 2021-05-12
fixed offline integer parsing bug

AMiller42:
Fixed bug where a number with multiple decimal points would only push the number up to the first decimal, e.g. 
```
123,          || 123       // Expected
123.456,      || 123.456   // Expected
123.456.,     || 123       // Not Expected
123.456.789,  || 123       // Not Expected
```
This bug only affects the offline interpreter, though I don't understand why.


Lyxal:
I can only assume it affects the offline interpreter because of how the safe eval regex works online

PR #61, opened 2021-05-12
strings and structures decided to get along

AMiller42:
There was a parsing bug where a string at the end of an integer or a variable name would be parsed as part of it, e.g.
```
123.456`hello`789
```
This program tries to convert ```123.456`hello`789``` to a float, which errors. 

```
42→x`world`
```
This program tries to assign `42` to a variable named ``` x`world` ``` which also errors.

Now, the first program pushes `123.456`, `hello`, and `789` to the stack. The second program assigns `42` to a variable named `x` and pushes `world` to the stack.

### DIFF
--- a/VyParse.py
+++ b/VyParse.py
@@ -250,9 +250,31 @@ def Tokenise(source: str):
             continue
 
         elif type(char) is list:
-            if structure != NO_STMT:
+            if structure not in [NO_STMT, INTEGER, VARIABLE_GET, VARIABLE_SET]:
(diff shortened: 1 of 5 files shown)


AMiller42:
I intended to make the `ÞS` addition a seperate PR, but I accidentally put them together.
¯\\\_(ツ)_/¯

AMiller42:
I intended to make the `ÞS` addition into its own PR, but I accidentally combined them.
¯\\\_(ツ)_/¯

PR #68, opened 2021-05-30
Use template literals because why not

ysthakur:
Totally unnecessary but I was bored


Lyxal:
Probably the  most useless pr yet, but eh

PR #71, opened 2021-05-31
Do some formatting

ysthakur:


ysthakur:
Ideally, all files would be formatted with black every time changes are made so that there'd be fewer conflicts, but it'd mess up Vyxal.py so much that I decided against it :/.
Btw, I'd suggest squashing and merging this, as you don't want all these commits polluting the commit history (I would have suggested this for my earlier PRs too but forgot to, sorry).

utilities.py
```diff
@@ -49,9 +58,10 @@ def from_ten(number, custom_base):
         interesting_part, temp = divmod(temp, base_exponent ** power)
         result = append(custom_base[interesting_part])
         power -= 1
-    
-    if temp == 0: result
-    
+
+    if temp == 0:
```
ysthakur, line 62: idk what this is for

PR #72, opened 2021-06-02
Fixed palindroming

chunkybanana:
Palindroming was using the vectorising add instead of the non-vectorising join, leading to unexpected results when palindroming lists.  Fixed.


Lyxal:
that's a no from me.

the irony is that this change breaks backwards compatibility for the behaviour on numbers.

PR #76, opened 2021-06-06
`Jo` flags

chunkybanana:
J prints the stack joined by newlines.

o forces implicit output, even when something has already been outputted. Works with other printing flags :)


Lyxal:
Who's Jo?

PR #84, opened 2021-06-11
Only run tests in run-tests/python-package.yml

ysthakur:
Also only use Python 3.8 for testing.


ysthakur:
This is a pretty minor change, but it'll be one less thing to take care of manually. I'm like 80% sure it'll work once you merge into master, it's just failing right now because there's a merge.

PR #86, opened 2021-06-12
Update experimental with changes from master

ysthakur:
Update the experimental branch with the new changes that just happened in master so it isn't lagging behind. I went through most of it, and as far as I can tell, it doesn't erase any of the changes already in the experimental branch, since there were only new files for a codepage and elements so far.


Lyxal:
I mean, there's not much point in doing so, as I'll be rewriting most files. Doesn't hurt though

PR #87, opened 2021-06-14
Add some functions

ysthakur:
This is untested, and certainly incomplete,  which is why it's just a draft PR (hopefully that doesn't ping anyone).


ysthakur:
I'm an idiot, the cartesian power thing multiplies a matrix instead of doing cartesian product.

PR #97, opened 2021-06-20
Improved dictionary compression

chunkybanana:
Dictionary compression is now fully optimised. The algorithm works by iterating through the string and finding the longest compressible string at each point.


chunkybanana:
This hopefully doesn't break backwards compatibility in any way.

PR #101, opened 2021-06-23
Make Q print with flag o (#1)

ysthakur:


ysthakur:
This should fix https://github.com/Lyxal/Vyxal/issues/99, but I'll let Lyxal decide if `o` should print even with `Q`.

Lyxal:
Sounds good to me

Lyxal:
w..why did you reference #1 ?

ysthakur:
@Lyxal Short story long: I created a branch ysthakur-patch-q (the default was ysthakur-patch-1 but I name it ysthakur-patch-q because it involved Q). It had a lot of commits, and since you always merge directly instead of squashing and merging ( >( ), I decided to delete my fork of Vyxal, make a new fork with the ysthakur-patch-q branch, delete the ysthakur-patch-q branch over here, and work on it on my fork. Then I squashed and merged it into my fork's master so that it would be a single commit when merged here. Over there, it was the first and only issue, so the `#1` got added to the title. Then it got merged into this master branch, where `#1` is a completely different issue. It also got a #101 added to the end this time because it's actually the 101th issue here. Blame GitHub for being stupid.

PR #140, opened 2021-07-09
Pull functions and global variables out of interpreter

ysthakur:
I pulled most of the builtins into builtins.py, pulled global variables such as `stack` and settings/flags into globals.py, utility classes and functions into utilities.py, and lowercased VY_* functions.

Edit: Also made an array_builtins.py, let me know if there are any more categories we can separate functions into.


ysthakur:
I have no idea why, but the FizzBuzz test didn't work. When I discard the `''`/`'Fizz'`/`'Buzz'`/`'FizzBuzz'` string using `_`, all I get is a list of `[43, 107, 43, 107...]` 100 times instead of the numbers themselves. Hopefully someone can figure out why.

ysthakur:
As Lyxal said, it was a problem with global settings. I've moved them back into interpreter.py, but I feel they need to be inside their own object, and that object should be passed around instead of the settings being global. This would also allow multiple programs to be run at once (and would make running multiple programs, one after another, easier, including the tests).

Lyxal (commented):
good

Lyxal:
But I'll need to put the stuff from the most recent commit into these files before merging

ysthakur:
@Lyxal I've tried to keep builtins.py up-to-date by going through the most recent commits and copying stuff from interpreter.py, but yeah, it'd be a good idea to check if recent bugs/features are fixed/working here too.

Issue #146, opened 2021-07-10
FR: Modify Cartesian Product to work better with infinite generators.


hyper-neutrino:
Currently, Cartesian Product returns pairs firstly in order of the first list and secondly in order of the second list. However, if the first list is an infinite sequence, this takes infinitely many steps to reach the second row, and thus is not a bijection to the natural numbers and doesn't reach every element in finite time.

If the first list is infinite and the second isn't, transposing the output (without changing the order of elements in the pair) will work; for example, `[1, 2, 3, ...] × [a, b, c]` can give `[[1, a], [1, b], [1, c], [2, a], [2, b], [2, c], ...]` rather than `[[1, a], [2, a], [3, a], ..., [1, b], ...]` (where the `[1, b]` will never be reached).

If both lists are infinite, the output can be done diagonally. For example, to pair `[1, 2, 3, ...], × [1, 2, 3, ...]`, we can return `[[1, 1], [1, 2], [2, 1], [1, 3], [2, 2], [3, 1], [1, 4], [2, 3], [3, 2], [4, 1], ...]`, which is an enumerable ordering and will reach every element in finite time.

ysthakur:
Dibs on implementing this feature if and when [this](https://github.com/Vyxal/Vyxal/pull/140) PR is merged.

PR #155, opened 2021-07-13
implement generator-friendly cartesian product

hyper-neutrino:
As per #146 this implements cartesian product that works with generators. The order is not exactly consistent; however, unlike currently, it will actually return every element (it is not possible for this to be consistent), with generators on the left, right, or both.


ysthakur (approved):
LGTM

PR #202, opened 2021-07-28
Create the lexer

Lyxal:


PyGamer0 (commented):
Why dont you use an Enum for TokenTypes?

vyxal/lexer.py
```diff
@@ -0,0 +1,193 @@
...
+            tokens.append(Token(TokenType.NAME, contextual_token_value))
+        elif head in "k∆øÞ¨":
+            if source:
+                tokens.append(Token(TokenType.GENERAL, head + source.popleft()))
+        else:
+            tokens.append(Token(TokenType.GENERAL, head))
+    return tokens
+
+
+if __name__ == "__main__":
```
ysthakur, line 181: Might want to run these as proper tests (compare against an expected output and throw an error if it doesn't work).

ysthakur, line 181: I think these should be run as proper tests (compare against an expected output and throw an error if it doesn't work). It doesn't feel right to be running this module all by itself, and IMO the only module with an `if __name__ == "__main__"` should be the interpreter/compiler.

ysthakur (changes requested):
Might want to run these as proper tests (compare against an expected output and throw an error if it doesn't work). It doesn't feel right to be running this module all by itself. Other than that, looks pretty good to me.

I can always remove this request if y'all feel that this is okay, though. Can you indicate that using the thumbs up/down emojis?

ysthakur (approved):

vyxal/lexer.py
Tams-Tams: What do mean by proper tests?

ysthakur: Well, each group of tests would be in a function. I was thinking all the tests for the lexer would be in a `lexer_tests.py` file with all the other tests. That way, if something fails, we'll know when we push.

Tams-Tams: who are you, actually?

Tams-Tams: A username?

vyxal/lexer.py
```diff
@@ -0,0 +1,178 @@
...
+structures, they need to be turned into tokens representing the
+different components of a program. For the full specification on token
+types, go to documents/specs/Lexer.md
+"""
+
+import collections
+import string
+
+
+class TokenType:
```
PyGamer0, line 13: why not an Enum??

PyGamer0 (commented):
mmm

vyxal/lexer.py
Tams-Tams: Yes yes

vyxal/lexer.py
ysthakur: I don't understand. In real life? I'm not telling you that, for obvious reasons.

Tams-Tams (approved):

vyxal/lexer.py
Lyxal: no. that seems _too_ much

ysthakur (approved):

PR #203, opened 2021-07-28
implement contextual name grouping for for loop variable names

hyper-neutrino:


Lyxal (changes requested):
The comment stuff is good, but leave the for loop name grouping for the parser.

PR #206, opened 2021-08-01
Fb parser

Lyxal:


PyGamer0 (approved):
Looks good

vyxal/parser.py
```diff
@@ -0,0 +1,436 @@
...
+This is done by treating the tokens as a queue and dequeuing tokens
+until a predicate is matched for structures.
+"""
+
+from __future__ import annotations
+
+import string
+from collections import deque
+
+try:
```
ysthakur, line 14: This is very sus imo, but what happens if you just do `import lexer` or `import vyxal.lexer`?

ysthakur (changes requested):
Other than that one import, this looks pretty good to me. The comments and formatting are quite helpful.

vyxal/parser.py
```diff
@@ -0,0 +1,436 @@
...
+            elif structure_name == StructureType.LAMBDA_MAP:
+                branches.insert(0, "1")
+                branches[1] = parse(branches[1])
+                structure_name = StructureType.LAMBDA
+                after_token = Structure(
+                    StructureType.NONE,
+                    lexer.Token(lexer.TokenType.GENERAL, "M"),
+                )
+                # laziness ftw
+            elif structure_name == StructureType.LAMBDA_MAP:
```
pxeger, line 345: This condition seems to be duplicated?

vyxal/parser.py
```diff
@@ -0,0 +1,436 @@
...
+                branches.insert(0, "1")
+                branches[1] = parse(branches[1])
+                structure_name = StructureType.LAMBDA
+                after_token = Structure(
+                    StructureType.NONE,
+                    lexer.Token(lexer.TokenType.GENERAL, "F"),
+                )
+                # laziness ftw
+            elif structure_name == StructureType.LAMBDA_MAP:
```
pxeger, line 354: duplicated again?

tests/test_parser.py
```diff
@@ -0,0 +1,76 @@
+import os
+import sys
+
+THIS_FOLDER = os.path.dirname(os.path.abspath(__file__)) + "/.."
```
pxeger, line 4: this is very janky - what are you trying to achieve? It seems like you're not running your code correctly if you need this

tests/test_parser.py
```diff
@@ -0,0 +1,76 @@
...
+                    ["none", ["general", "∑"]],
+                    ["none", ["general", "∴"]],
+                ],
+            ],
+        ],
+        ["none", ["general", "M"]],
+    ]
+
+
+if __name__ == "__main__":  # For testing outside of the workflow
```
pxeger, line 74: what exactly is "the workflow"? Using `pytest`? Just use pytest; it's not very complicated

pxeger (changes requested):

tests/test_parser.py
ysthakur: Yeah, the tests should simply be functions whose names start with `test_`. We shouldn't need the `__name__ == "__main__"` thing *anywhere* except in the main interpreter file.

tests/test_parser.py
ysthakur: Doesn't seem that bad, but maybe we could use relative imports (`...`) instead?

tests/test_parser.py
Lyxal: Outside of "the workflow" is testing locally using the command line.

ysthakur: @Lyxal It's still fine - we can do that using `pytest tests` (or whatever directory/file you want).

Lyxal: I more mean actually running the file on my computer, because I don't have py_tests set up locally.

tests/test_parser.py
razetime: seems like something necessary to keep the imports clean. @Lyxal thoughts?

razetime (approved):
All things reviewed, should find an alternate method later without adding the vyxal foflder to path.

pxeger (approved):

tests/test_parser.py
```diff
@@ -74,4 +74,3 @@ def test_fizzbuzz():
 if __name__ == "__main__":  # For testing outside of the workflow
```
ysthakur, line 74: This needs to be removed too btw.

Lyxal: yesn't

ysthakur: wdym

razetime: this is a general thing in python programs

pxeger: not with pytest it ain't

PR #213, opened 2021-08-13
Heed (most of) mypy's and flake8's warnings/errors

ysthakur:
Ran mypy and flake8 on vyxal and found a few potential bugs. There's more weird stuff, such as `"1"` being passed when a `list[Structure]` is expected (@Lyxal is that intended or just something left over from old code?), but I didn't try fixing it.


vyxal/LazyList.py
```diff
@@ -153,7 +152,7 @@ def output(self, end="\n"):
             print("⟩", end=end)
 
 
-"""
+r"""
 
 ,--------. ,-----. ,------.   ,-----.         ,--.   ,--.        ,--.
 '--.  .--''  .-.  '|  .-.  \ '  .-.  '.--.    |   `.'   | ,--,--.|  |,-. ,---.
```
pxeger, line 158: dare I ask?

vyxal/LazyList.py
```diff
@@ -99,7 +98,7 @@ def infinite_index():
                 while len(self.generated) < position + 1:
                     try:
                         self.__next__()
```
pxeger, line 100: `next(self)`?

vyxal/helpers.py
```diff
@@ -67,6 +67,45 @@ def uncompress(token: lexer.Token) -> Union[int, str]:
...
+    string: str
+        The string to uncompress.
+
+    Returns
+    -------
+    str
+        The uncompressed string.
+    """
+    # TODO (lyxal) Implement string (un)compression
+    raise ValueError("Not implemented")
```
pxeger, line 93: raise `NotImplementedError`? (not to be confused with the special value `NotImplemented`)

vyxal/helpers.py
```diff
@@ -67,6 +67,45 @@ def uncompress(token: lexer.Token) -> Union[int, str]:
...
+    number: str
+        The number to uncompress.
+
+    Returns
+    -------
+    int
+        The uncompressed number.
+    """
+    # TODO (lyxal) Implement number (un)compression
+    raise ValueError("Not implemented")
```
pxeger, line 111: `NotImplementedError`?

vyxal/transpile.py
```diff
@@ -50,40 +50,41 @@ def transpile_ast(program: list[structure.Structure], indent=0) -> str:
...
+        f" was {type(token_or_struct).__name__}: {token_or_struct}"
     )
 
 
 def transpile_token(token: Token, indent: int) -> str:
-    from helpers import indent_str, uncompress
+    from vyxal.helpers import indent_str, uncompress
 
     if token.name == TokenType.STRING:
         # Make sure we avoid any ACE exploits
```
pxeger, line 68: Just use Python's builtin `repr` here (or the `!r` format string flag); it's guaranteed to be safer.

```python
return indent_str(f"stack.append({string!r})")
```

vyxal/transpile.py
```diff
@@ -50,40 +50,41 @@ def transpile_ast(program: list[structure.Structure], indent=0) -> str:
...
+            elements.elements.get(token.value, ("pass\n", -1))[0], indent
         )
     elif token.name == TokenType.COMPRESSED_NUMBER:
-        return indent_str(f"stack.append({uncompress(token)})")
+        return indent_str(f"stack.append({uncompress(token)})", indent)
     elif token.name == TokenType.COMPRESSED_STRING:
-        return indent_str(f"stack.append('{uncompress(token)}')")
+        return indent_str(f"stack.append('{uncompress(token)}')", indent)
         # No need to check for ACE exploits here because this string
         # type will only ever contain lower alpha + space.
```
pxeger, line 83: famous last words...

I'd recommend also using `repr` as above.

vyxal/transpile.py
```diff
@@ -272,31 +269,44 @@ def transpile_structure(struct: structure.Structure, indent: int) -> str:
...
             + element_B
             + "\n"
             + element_C
             + "\n"
-            + elements.modifiers.get(struct.branches[0])
+            + elements.modifiers[struct.branches[0]]
         )
 
     raise ValueError(struct)
+    raise ValueError(struct)
```
pxeger, line 312: be more specific with this error? (also why is it here twice?)

pxeger (changes requested):
mostly looking good

vyxal/transpile.py
Lyxal: Except this type of string can only contain lowercase letters and spaces once uncompressed.

vyxal/transpile.py
Lyxal: That seems like a code duplication error

vyxal/helpers.py
Lyxal: This seems trivial for something that will be removed soon

vyxal/helpers.py
Lyxal: This seems trivial for something that will be removed soon

vyxal/LazyList.py
Lyxal: No.

vyxal/transpile.py
pxeger: I still think it's important to fix this. What if we add decompression of words with apostrophes, and then they break out of the single-quoted string? Just use repr. https://en.wikipedia.org/wiki/Defense_in_depth_(computing)

vyxal/helpers.py
pxeger: Mehj, ok. It was just a suggestion

vyxal/transpile.py
Lyxal: No no I mean quite literally it's physically impossible to ACE exploit this string type.

It's `«...«`. You're confusing this type with dictionary strings.

Your point is valid for the other string type, but not this one.

vyxal/transpile.py
Lyxal: Fixed

Lyxal (approved):

vyxal/transpile.py
pxeger: I don't care; it is good practice to apply such considerations even in situations where you think it's physically impossible

vyxal/helpers.py
ysthakur: @Lyxal Maybe, but it's good to know for the future.

PR #214, opened 2021-08-14
Use separate parameters instead of list[list[Structure]] for structure constructors

ysthakur:
- Moved the branch-making stuff into a function `_get_branches` to reduce the length/complexity of `parse()`.
- Make `branches` in structures variadic.
- Make a `Branch` typealias (union of `str`, `list[Token]`, and `list[Structure]`) so that `*branches: Branch`.
- Most `Structure` subclasses have parameters such as `name: str, body: Branch` instead of `branches` or `*branches`.
- Gave `Lambda` an `arity: int` parameter. The default value is `1` instead of `"1"`.
- Made a `FunctionDef` class to differentiate between function definitions and function calls. Though they may be syntactically similar, they are entirely different concepts, and I thought they ought to be represented by different classes.
- Removed `eval(str())` from tests. It can be replaced with normal comparisons to structure objects.

These are somewhat big changes, so please review them and see if you want them merged at all.


ysthakur:
@Lyxal Why not make it a package? It'd make using the repl much easier too.

Lyxal:
> hey sorry but it literally is

then why do I get `vyxal isn't a package` as an error when running Vyxal.py?

pxeger:
Don't try to run files in a package directly; use `python -m`:

```bash
$ python -m vyxal.vyxal
  [...]
  File "/[...]/vyxal/vyxal/vyxal.py", line 8, in <module>
    from context import Context
ModuleNotFoundError: No module named 'context'
$ # now switch to before Lyxal removed all the `vyxal.`s in the imports
$ git checkout HEAD~1
[...]
HEAD is now at df42d84 Merge branch 'fb-transpiler' into copy-fbt
$ python -m vyxal.vyxal
[a (somewhat broken) vyxal REPL appears]
```

If you wanted more evidence: [the definition of a package](https://docs.python.org/3/glossary.html#term-package):
> A Python module which can contain submodules or recursively, subpackages. Technically, a package is a Python module with an `__path__` attribute.

```python
>>> import vyxal
>>> vyxal.__path__
['/[...]/vyxal/vyxal']
```

Typically a package is just a directory containing an `__init__.py` file.

pxeger (approved):
:thumbsup:

Lyxal (changes requested):
Try running `1 1+` in the REPL. You'll see that you get:

```
  File "...\Python\Python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "...\Python\Python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "...\vyxal\vyxal\vyxal.py", line 20, in <module>
    line = transpile(line)
  File "...\vyxal\vyxal\transpile.py", line 36, in transpile
    return transpile_ast(parse.parse(lexer.tokenise(program)))
  File "...\vyxal\vyxal\transpile.py", line 45, in transpile_ast
    return "\n".join(
  File "...\vyxal\vyxal\transpile.py", line 46, in <genexpr>
    transpile_single(struct, indent=indent) for struct in program
  File "...\vyxal\vyxal\transpile.py", line 57, in transpile_single
    return transpile_structure(token_or_struct, indent)
  File "...\vyxal\vyxal\transpile.py", line 96, in transpile_structure
    return transpile_single(struct.branches[0][0], indent)
  File "...\vyxal\transpile.py", line 58, in transpile_single
    raise ValueError(
ValueError: Input must be a Token or Structure, was list: [Token('number', '1')]
```

Very sad.

pxeger:
Oh, sorry, I assumed we knew it didn't work yet

Lyxal:
I would expect the basic transpilation/parsing/lexing to work before merging (like at least print generic things like `pass` instead of erroring)

Lyxal (approved):
I made the changes I requested

vyxal/transpile.py
```diff
@@ -130,7 +130,9 @@ def transpile_structure(struct: structure.Structure, indent: int) -> str:
         return res
     if isinstance(struct, structure.ForLoop):
         # TODO (user/ysthakur) make it work with multiple variables
-        var = struct.names[0] if struct.names else f"LOOP{secrets.token_hex(16)}"
+        var = (
+            struct.names[0] if struct.names else f"LOOP{secrets.token_hex(16)}"
```
pxeger, line 134: I'm sure some of us must have slightly different linting settings, because this kind of meaningless reformatting seems to happen a lot. Not a big deal, but discuss in chat?

pxeger (approved):

vyxal/transpile.py
Lyxal: sure

Issue #160, opened 2021-07-13
Much feature requests


chunkybanana:
### Constants
- [x] `kT` - BF command set
- [x] `kṗ` - List  of all bracket pairs - `[(),[],{},<>]`
- [x] `kṖ` - All brackets nested - `([{<>}])`
- [x] `kg` - Phi
- [x] `kS` - Amogus (`ඞ`)
- [x] `k₂` - 2^20
- [x] `k₃` - 2^30
- [x] `k∪` - lowercase vowels with `y`
- [x] `k⊍` - uppercase vowels with `y`
- [x] `k∩` - both cases of vowels with `y`
- [x] `Þo` -Infinite list of ordinals
- [x] `Þc` - Infinite list of cardinals
- [x] `Þp` - Infinite list of primes
### Flags
- [x] #407
- [x] #408
- [x] `t` - Lists are truthy if any item is truthy, else falsy
- [x] #456
- [x] #455
- [x] `2` - Default lambda arity = 2
- [x] `3` - Default lambda arity = 3
### String stuff
- [x] `øb` = Enclose in `()`
- [x] `øB` = Enclose in `[]`
- [x] `øḃ` = Enclose in `{}`
- [x] `øḂ` = enclose in `<>`
- [x] `øβ` - Are brackets (`(),{},[],<>`) balanced?
- [x] `ø↳` and `ø↲` = custom left / right padding
- [x] `øĿ` - Ring translate
- [x] `øv` -  Is_vowel - handles like `C` for multichar strings
- [x] `øċ` - Is_uppercase - handles like `C` for multichar strings, 1 if char is an uppercase ASCII letter, else 0
- [x] `øM` = Palindromise (don't duplicate middle) with ascii-art swap mappings
### Number stuff
- [x] `∆i` - nth-digit-of-pi
- [x] `∆ė` - nth-digit-of-e
- [x] `∆I` - First n digits of pi
- [x] `∆Ė` - First n digits of e
- [x] `∆f` - nth-fibonacci-number
- [x] `∆ε` - Evaluate polynomial equation with coefficents and variable
- [x] `ÞB` -  Array of random bits to length n 
- [x] `∆Z` - zfill - prepend 0s to a until it's length b
### Other stuff
- [x] #399
- [x] `Þu` - All_unique
- [x] `ÞẊ` - Cartesian power
- [x] `Þx` - `Þ×` but with replacements
- [x] `Þf` - Flattento  a certain depth
[...]

PR #218, opened 2021-08-16
Update fresh-beginnings with our current transpiler stuff before starting the command dictionary

Lyxal:
This is just to get the structures into the fresh beginnings branch. The transpiler branch won't be deleted, and will still be used.

_Are you really sure we need this?_

Yes. It creates a point in fresh-beginnings that comes before the actual elements.


Lyxal:
Also, I'm personally requesting no merging until 4 reviews instead of the usual 2

PyGamer0 (approved):
looks good. have a cookie for you hard work: 🍪 :P

documents/specs/Vecorisation.md
```diff
@@ -0,0 +1,21 @@
...
+Then, it's just:
+
+```python
+LazyList(map(function, left))
+```
+# Dyads
+
+```python
+LazyList(map(lambda x: apply(*x), vy_zip(left, right)))
+```
```
pxeger, line 21: what about triads?

tests/test_transpiler.py
```diff
@@ -0,0 +1,18 @@
...
+    py = transpile(vy)
+    expected = """condition = pop(stack, ctx=ctx)
+ctx.context_values.append(condition)
+if boolify(condition, ctx):
+    stack.append(1)
+else:
+    stack.append(2)
+ctx.context_values.pop()
+"""
+    print(py)
```
pxeger, line 17: is this `print` still needed?

vyxal/LazyList.py
```diff
@@ -0,0 +1,221 @@
+"""
+File: LazyList.py
+Description: A generic wrapper for all sorts of generators. This is
+because itertools doesn't return things that return true when yeeted
```
pxeger, line 4: yote

documents/specs/Elements.txt
```diff
@@ -584,6 +585,10 @@ k⋎                                   = 2 ** 30
 øṀ (a: string)                       = vertical_mirror(a, mapping=flip_slashes_and_brackets) # [ <-> ], ( <-> ), { <-> }, < <-> >, / <-> \
    (otherwise)                       = vectorises
 ø¦ (a: string, b: string)            = vertical_mirror(a, mapping=b) # mapping in the form of ["normal", "flipped"] - custom mapping
+øb (a: string)                       = bracketify(a)
+øe (a: string, b: string)            = enclose b in a
```
razetime, line 589: will this also mirror?

will enclosing "Hello" in `{" be "{Hello}" or "{Hello{"?

vyxal/LazyList.py
```diff
@@ -0,0 +1,221 @@
...
+            print("⟩", end=end)
+
+
+r"""
+
+,--------. ,-----. ,------.   ,-----.         ,--.   ,--.        ,--.
+'--.  .--''  .-.  '|  .-.  \ '  .-.  '.--.    |   `.'   | ,--,--.|  |,-. ,---.
+   |  |   |  | |  ||  |  \  :|  | |  |'--'    |  |'.'|  |' ,-.  ||     /| .-. :
+   |  |   '  '-'  '|  '--'  /'  '-'  '.--.    |  |   |  |\ '-'  ||  \  \\   --.
+   `--'    `-----' `-------'  `-----' '--'    `--'   `--' `--`--'`--'`--'`----'
```
pxeger, line 161: ...

vyxal/LazyList.py
```diff
@@ -0,0 +1,221 @@
...
+                print("|", end="")
+            while True:
+                print(lhs, end="")
+                lhs = next(self)
+                print("|", end="")
+        except:
+            print("⟩", end=end)
+
+
+r"""
```
razetime, line 155: please shorten this. This is a waste of time to read and make.

vyxal/elements.py
```diff
@@ -0,0 +1,45 @@
...
+
+    arity : int
+        The arity of the element - <= 3
+
+    Returns
+    -------
+    (str, int)
+        See documents/specs/Transpilation.md for information
+    """
+    arguments = ["third", "rhs", "lhs"][-arity:] if arity else "_"
```
pxeger, line 31: the behaviour of arity=0 functions isn't made very clear here. Could I suggest having a separate code path to handle them explicitly?

vyxal/helpers.py
```diff
@@ -0,0 +1,111 @@
...
+    ----------
+    string: str
+        The string to be indented.
+    indent: int
+        How many levels of indentation are required.
+
+    Returns
+    -------
+    str
+        `string`, indented by `(4 * indent)` spaces, after `\n`.
```
pxeger, line 28: what does "after `\n`" mean here?

vyxal/lexer.py
```diff
@@ -100,32 +99,33 @@ def __repr__(self) -> str:
         Returns
         -------
         str
-            [name, value]
+            Token(name, value)
         """
 
-        return str([self.name.value, self.value])
+        return f"Token({repr(self.name.value)}, {repr(self.value)})"
```
pxeger, line 105: Fun fact: in a format string, you can do `f"{foo!r}"` instead of `f"{repr(foo)}"`

Doesn't really need changing, but I thought you might like to know

vyxal/lexer.py
```diff
@@ -100,32 +99,33 @@ def __repr__(self) -> str:
...
 
         Returns
         -------
 
-        True iff the two token names and values are the same.
+        True iff `rhs` is a token and has the same name and value.
         """
-
+        if not isinstance(rhs, Token):
+            return False
```
pxeger, line 124: I believe standard Python practice would be to return `NotImplemented` for other types, so the other object can decide whether it's comparable

vyxal/lexer.py
```diff
@@ -158,7 +158,6 @@ def tokenise(source: str) -> list[Token]:
             if source:
                 # This has the consequence of making backslahses at the
                 # end of a program not error.
-
                 tokens.append(Token(TokenType.STRING, source.popleft()))
```
pxeger, line 161: why? I don't quite understand why this fixes backslashes

vyxal/vyxal.py
```diff
@@ -0,0 +1,21 @@
...
+from vyxal.context import Context
+from vyxal.transpile import transpile
+
+ctx = Context()
+
+if __name__ == "__main__":
+    # I'm allowed to have this here this time. Frick you if you say I
+    # can't.
+
+    while True:
```
razetime, line 17: is every line a separate vyxal program, or do we have plans of a repl where all commands effect the global stack?

vyxal/parse.py
```diff
@@ -11,107 +11,28 @@
...
-        value
-
-        Returns
-        -------
-        str
-            [name, value]
-        """
-
-        return str([self.name.value, self.branches])
+    import vyxal.structure as structure
```
pxeger, line 21: `from vyxal import structure` is better

razetime (changes requested):
I've given some feedback after reading through everything.

vyxal/parse.py
```diff
@@ -270,58 +159,87 @@ def parse(tokens: list[lexer.Token]) -> list[Structure]:
...
-                branches[0] = process_parameters(branches[0])
+                    var_names = [
+                        variable_name(branch) for branch in branches[:-1]
+                    ]
+                body = parse(branches[-1])
+                structures.append(structure.ForLoop(var_names, body))
+            elif structure_cls == structure.WhileLoop:
+                if len(branches) == 1:
+                    # If there's no condition, it's an infinite loop
+                    condition = [lexer.Token(lexer.TokenType.NUMBER, "1")]
```
pxeger, line 172: this seems like an overly complicated way of handling an infinite loop. Is there some more elegant way of setting the condition to `1` (or `True`)?

vyxal/parse.py
```diff
@@ -331,32 +249,23 @@ def parse(tokens: list[lexer.Token]) -> list[Structure]:
...
-                        StructureType.MONADIC_MODIFIER,
-                        [head.value, [remaining[0]]],
-                    )
+                    structure.MonadicModifier(head.value, remaining[0])
                 )
             structures += remaining[1:]
             break
         elif head.value in DYADIC_MODIFIERS:
             remaining = parse(tokens)
             if head.value == "‡":
```
pxeger, line 261: add a comment to describe what this is doing so I don't have to go and look up the meaning of `‡`?

vyxal/structure.py
```diff
@@ -21,76 +39,93 @@ def __init__(self, branches: list[list[Structure]]):
...
-        self.parameters = branches[0]
-        self.body = None
-        if len(branches) >= 2:
-            self.body = branches[-1]
 
-    def transpile(self) -> str:
-        return ""
+class LambdaMap(Lambda):
+    def __init__(self, body: list[Structure]):
+        print("Here")
```
pxeger, line 84: remove?

vyxal/transpile.py
```diff
@@ -0,0 +1,293 @@
...
+        return indent_str(f"stack.append(VAR_{token.value})", indent)
+    elif token.name == TokenType.VARIABLE_SET:
+        return indent_str(f"VAR_{token.value} = pop(stack, ctx=ctx)", indent)
+    raise ValueError(f"Bad token: {token}")
+
+
+def transpile_structure(struct: structure.Structure, indent: int) -> str:
+    """
+    Transpile a single structure.
+    # TODO (exedraj/lyxal, user/ysthakur) implement all structures here
```
pxeger, line 91: is there a reason that the transpilation isn't a method of each type of structure, called with dynamic dispatch, instead of explicitly checking with a whole load of `isinstance`s?

pxeger: I realise it might be a bit late to suggest this lol

vyxal/transpile.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    from vyxal.helpers import indent_str
```
pxeger, line 93: why isn't this at the top of the file?

vyxal/transpile.py
```diff
@@ -0,0 +1,293 @@
...
+                    "ctx.input_values[ctx.input_level] = [stack[::], 0]",
+                    indent + 1,
+                )
+                + transpile_ast(struct.body, indent + 1)
+                + indent_str("ctx.context_values.pop()", indent + 1)
+                + indent_str("ctx.input_level -= 1", indent + 1)
+                + indent_str("return stack", indent + 1)
+            )
+    if isinstance(struct, structure.Lambda):
+        signature = secrets.token_hex(16)
```
pxeger, line 191: I don't think `signature` is the right word. Signature typically refers to the arity and types of a function. Maybe just `id` would be better?

vyxal/transpile.py
```diff
@@ -0,0 +1,293 @@
...
+            element_A
+            + "\n"
+            + element_B
+            + "\n"
+            + element_C
+            + "\n"
+            + elements.modifiers.get(struct.modifier, "pass")
+        )
+
+    raise ValueError(struct)
```
pxeger, line 293: Is this error never supposed to be reached? If it is unreachable, use `assert False`; otherwise, add some more detail to the error message?

vyxal/vyxal.py
```diff
@@ -0,0 +1,21 @@
+"""
+File: vyxal.py
+Description: This is the main file for the project - the transpilation
+of Vyxal programs actually happens here, and this is what gets executed
+offline.
```
pxeger, line 5: this description seems outdated. The transpilation does not happen here.

vyxal/vyxal.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+
+from vyxal.context import Context
+from vyxal.transpile import transpile
+
+ctx = Context()
+
+if __name__ == "__main__":
+    # I'm allowed to have this here this time. Frick you if you say I
+    # can't.
```
pxeger, line 15: how are we going to handle flags, and command line parsing in general?

vyxal/vyxal.py
```diff
@@ -0,0 +1,21 @@
...
+File: vyxal.py
+Description: This is the main file for the project - the transpilation
+of Vyxal programs actually happens here, and this is what gets executed
+offline.
+"""
+
+from vyxal.context import Context
+from vyxal.transpile import transpile
+
+ctx = Context()
```
pxeger, line 11: does this need to be passed in somewhere?

pxeger (changes requested):
Looking pretty good, just some minor things

documents/specs/Elements.txt
```diff
@@ -584,6 +585,10 @@ k⋎                                   = 2 ** 30
 øṀ (a: string)                       = vertical_mirror(a, mapping=flip_slashes_and_brackets) # [ <-> ], ( <-> ), { <-> }, < <-> >, / <-> \
    (otherwise)                       = vectorises
 ø¦ (a: string, b: string)            = vertical_mirror(a, mapping=b) # mapping in the form of ["normal", "flipped"] - custom mapping
+øb (a: string)                       = bracketify(a)
```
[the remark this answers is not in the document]
razetime: It was suggested as the underload type, so `()`

documents/specs/Elements.txt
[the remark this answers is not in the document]
razetime: In stax it says this: Prepend and append string/array. Wraps the first array between copies of the second.

I assume it is the  same here.

documents/specs/Context.md
```diff
@@ -1,8 +1,100 @@
...
+class Context:
+    def __init__(self):
+        self.stack = []
+        self.register = 0
+        # and so on
+```
+
+At the start of the interpreter, an instance of `Context` called `ctx` (think `ctx = Context()`) will be created. This will need to be passed between element functions, meaning they all need a `ctx` parameter.
+
+However, we're yet to decide whether this should happen explicitly or implicitly.
```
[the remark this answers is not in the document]
Lyxal: Explicit is the way we're going. I'll update the code guide accordingly

documents/specs/Elements.txt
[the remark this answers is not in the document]
pxeger: so `øe("abc", "def")` = `abcdefabc`?

documents/specs/Context.md
```diff
@@ -1,8 +1,100 @@
...
+```
+
+## The Explicit Way
+
+(for example, this could be `vyxal.py`)
+
+```python
+# Simulate main.py
+
+from implicits import implicits
```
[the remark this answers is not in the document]
Lyxal: Yes, it should

documents/specs/Elements.txt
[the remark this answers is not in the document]
pxeger: so should that be documented?

documents/specs/Context.md
```diff
@@ -1,8 +1,100 @@
...
+lst = [1, 2, 3, 4]
+print(lst)
+x = elements.pop(lst, 2, ctx)
+print(x, lst)
+```
+
+(and this could be `elements.py`)
+
+```python
+from implicits import implicits
```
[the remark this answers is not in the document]
Lyxal: Correct

vyxal/transpile.py
Lyxal: `id` is a python built-in. But I agree that a name change would be suitable

pxeger: `id_` would be the standard workaround, but tbh, when have you ever used the `id` builtin? I normally just cope with it being shadowed

documents/specs/Elements.txt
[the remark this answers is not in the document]
razetime: I'm asking if the string gets "mirrored" as well. (see my review)

documents/specs/Elements.txt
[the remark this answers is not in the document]
razetime: I think it should, yes,

vyxal/transpile.py
Lyxal: Originally, there was a transpile method inside each structure subclass. It was removed by user though.

But I personally think it's better to have the if chain because it could easily be refactored to use pattern matching if we ever migrate to python 3.10

pxeger: hmm, yeah ok. I guess it's the old "separation of concerns vs single responsibility principle" question, and the answer probably doesn't matter too much

documents/specs/Elements.txt
```diff
@@ -536,6 +536,7 @@ k∩                                   = "aeiouyAEIOUY"
 kṗ                                   = phi
 k⋏                                   = 2 ** 20
 k⋎                                   = 2 ** 30
+k*                                   = "+-<>[].,"
```
chunkybanana, line 539: If we're going to have this, we should add *everything* from #160 .

documents/specs/Elements.txt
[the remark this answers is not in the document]
chunkybanana: What I meant wgen making this was `øe("{}", "def")` = `{def}`.

documents/specs/Elements.txt
pxeger: I think they should be added separately and one at a time. Arguably this shouldn't be added now at all

documents/specs/Vecorisation.md
```diff
@@ -0,0 +1,21 @@
...
+Numbers are converted to either `range(int(left))` or `digits_of(left)` based on the
+flags used.
+
+Then, it's just:
+
+```python
+LazyList(map(function, left))
+```
+# Dyads
+
```
chunkybanana, line 18: Can we specify that nilads are just filling the list?

chunkybanana (changes requested):

vyxal/vyxal.py
Lyxal: Every line is a different vyxal program (at least that's how I intend it)

vyxal/LazyList.py
Lyxal: no.

vyxal/helpers.py
Lyxal: I don't know. Ask @ysthakur

Lyxal: But it seems it means that it only inserts indents after newlines: https://docs.python.org/3/library/textwrap.html#textwrap.indent

vyxal/parse.py
Lyxal: At least with this way, `1` is always placed inside the context variable. I prefer this behaviour.

vyxal/vyxal.py
Lyxal: This will be passed between functions in the transpiled program.

documents/specs/Elements.txt
Lyxal: AUsername means adding them to the element list, rather than implementing them.

pxeger: IMO it doesn't make much sense to add things to the element list separately; they should be added in the same PR as the code that implements them

Lyxal: Sure, but I added some of the requests from #160 to the elements list before this, so they suggested to go the full way.

pxeger (dismissed):
yay

vyxal-cookie:
I agree with @pxeger

razetime (approved):

documents/specs/Vectorisation.md
```diff
@@ -0,0 +1,31 @@
...
+
+Then, it's just roughly equivalent to:
+
+```python
+LazyList(map(function, left))
+```
+# Dyads
+
+```python
+LazyList(map(lambda x: apply(*x), vy_zip(left, right)))
```
ysthakur, line 24: May I ask why this isn't `lambda x, right: apply(x, right)` like triads?

tests/test_parser.py
```diff
@@ -6,6 +6,7 @@
 
 from vyxal.lexer import *
 from vyxal.parse import *
+from vyxal.structure import *
```
ysthakur, line 9: I think I accidentally added this line because of the previous `eval` stuff, I'll remove it later.

vyxal/LazyList.py
ysthakur: cookie

ysthakur (approved):
Looks good to me.

documents/specs/Vectorisation.md
Lyxal: because vy_zip is smart and uses 0 if an iterator is exhausted until both are exhausted

ysthakur (approved):

tests/test_parser.py
```diff
@@ -6,7 +6,6 @@
 
 from vyxal.lexer import *
 from vyxal.parse import *
-from vyxal.structure import *
```
pxeger: I trust you @ysthakur that this is not going to horribly break everything and this PR is still approve-worthy

pxeger (approved):

PyGamer0 (approved):
🍪

documents/protocols/code guide.md
```diff
@@ -175,12 +175,4 @@ Instead, we create test files for each file so that they can be all run together
...
-- Functions that require access to flags or functions that directly or indirectly call functions requiring access to flags must be decorated using `@implicits("ctx")`, and their last parameters must be `*` and `ctx`. The `ctx` parameter refers to an object containing settings, the stack, and other values that need to be passed around to various functions. The decorator is from the [implicits package](https://pypi.org/project/implicits/). Here's an example:
-
-```python
-@implicits("ctx")
-def add(lhs, rhs, *, ctx):
-    # body of function
-```
-
-_Note that this revision is far from complete._
+- Whenever you write an element function, make sure you include a `ctx` parameter as the last parameter. This allows element functions to access values that would otherwise be global variables.
```
hyper-neutrino, line 178: is the `ctx` parameter supposed to be kwarg-only or just a positional, and if the latter, why the last and not the first parameter?

that is, is it `def add(lhs, rhs, ctx)` or `def add(lhs, rhs, *, ctx)`?

tests/test_parser.py
```diff
@@ -30,19 +30,19 @@ def fully_parse(program: str) -> list[Structure]:
 
 def test_basic():
     assert (
-        str(fully_parse("1 1+"))
+        fully_parse("1 1+")
         == "[['none', ['number', '1']], ['none', ['number', '1']], "
         + "['none', ['general', '+']]]"
```
hyper-neutrino, line 35: is it better practice to split lines across strings using `+` or just placing them side-by-side? also, since you are not calling `str` on the result of `fully_parse`, shouldn't the right side of `==` be a list and not its string representation?

tests/test_parser.py
```diff
@@ -30,19 +30,19 @@ def fully_parse(program: str) -> list[Structure]:
...
     )
 
     assert (
-        str(fully_parse("`Hello, World!`"))
+        fully_parse("`Hello, World!`")
         == "[['none', ['string', " + "'Hello, World!']]]"
```
hyper-neutrino, line 40: see above comment; this looks like it should be a list not a string representation, unless you overloaded `__eq__` for the result of `fully_parse`

hyper-neutrino (commented):
Please clarify the expected structure of `ctx` when declaring functions and see my comments regarding the testing. Unfortunately I don't really have time to look through every single line but the rest looks fine from a glance.

tests/test_parser.py
pxeger: Yes, implicit concatenation is preferred by most style guides.

```python
    assert (
        fully_parse("1 1+")
        == "[['none', ['number', '1']], ['none', ['number', '1']], " \
        "['none', ['general', '+']]]"
```

tests/test_parser.py
pxeger: yeah, `fully_parse` returns a plain `list`

ysthakur: Yeah, the tests need to be rewritten. They used to use `__repr__` to turn into stringy lists and then they'd be `eval`ed to get a proper list back or something like that. Shouldn't be terribly difficult though, I'll get on it once we merge this.

vyxal/parse.py
```diff
@@ -270,58 +153,87 @@ def parse(tokens: list[lexer.Token]) -> list[Structure]:
...
-            elif structure_name == StructureType.LAMBDA_FILTER:
-                branches.insert(0, "1")
-                branches[1] = parse(branches[1])
-                structure_name = StructureType.LAMBDA
-                after_token = Structure(
-                    StructureType.NONE,
-                    lexer.Token(lexer.TokenType.GENERAL, "F"),
+
+            elif structure_cls == structure.LambdaFilter:
+                structures.append(structure.Lambda(1, parse(branches[0])))
```
chunkybanana, line 216: What casing are we using?

vyxal/transpile.py
```diff
@@ -0,0 +1,291 @@
...
+                indent_str("def list_item(s, ctx):", indent)
+                + indent_str("stack = s[::]", indent + 1)
+                + transpile_ast(x, indent + 1)
+                + indent_str("return pop(stack, ctx=ctx)", indent + 1)
+                + indent_str(
+                    "temporary_list.append(list_item(stack, ctx)", indent
+                )
+            )
+
+        temp += indent_str("stack.append(temp_list[::]", indent)
```
chunkybanana, line 256: This is a *big* problem - which are we using, `temporary_list` or `temp_list`? Because this is going to be bad. Also how do we avoid name conflicts with that, and list_item?

vyxal/parse.py
ysthakur: wdym?

ysthakur: Heh, looks like I can reply from VSCode now.

vyxal/transpile.py
razetime: I say shorter names are better.

vyxal/parse.py
razetime: looks like CamelCase for classes and snake_case for variable, @chunkybanana.

documents/protocols/code guide.md
Lyxal: This is defined in context.md

vyxal/transpile.py
Lyxal: This isn't an issue because function scoping.

Lyxal: Also, `temp_list`.

vyxal/parse.py
Lyxal: This is actually covered in the code guide: https://github.com/Vyxal/Vyxal/blob/fb-transpiler/documents/protocols/code%20guide.md

razetime (approved):

vyxal/transpile.py
chunkybanana, line 256: Put in `temp_list` then!!!

chunkybanana (approved):

razetime (approved):

documents/protocols/code guide.md
```diff
@@ -1,10 +1,10 @@
 # The Definitive Vyxal Code Style Guide
-_Fourteenth revision_
+_Sixteenth revision_
```
pxeger, line 2: Do we really need to keep track of these revisions explicitly? You can just look back at the file's history in git

pxeger (approved):

documents/protocols/code guide.md
Lyxal: It's mostly for posterity.

pxeger: Git has posterity built-in? (PS: please don't "Resolve conversation" if discussion is clearly ongoing)

Lyxal: I like having an explicit revision count.

AMiller42 (approved):
Using that heckin' dank energy to do stuff I forgot to do yesterday.

PR #220, opened 2021-08-17
Relax code guide rules about docstrings

pxeger:
https://chat.stackexchange.com/transcript/message/58909173#58909173


Lyxal (dismissed):
This looks reasonable to me.

Lyxal (dismissed):

pxeger:
Sorry, I shouldn't really keep requesting your review; I should wait until I'm finished

Lyxal:
> Sorry @Lyxal, I shouldn't really keep requesting your review; I should wait until I'm finished

It's only fair after I requested review from y'all so many times for #218 ;P

Lyxal (approved):
I'm good with this.

vyxal/structure.py
```diff
@@ -9,9 +13,7 @@ def __init__(self, *branches: Branch):
         self.branches = branches
 
     def transpile(self) -> str:
```
ysthakur, line 15: I don't think we need this anymore, honestly. Transpilation should be separated from the definitions of the structures (and it already is anyway).

ysthakur (approved):
Looks good to me.

AMiller42 (approved):
ok

vyxal/structure.py
pxeger: Yeah, I thought that too, but that's a PR for a different day I think

PR #221, opened 2021-08-17
Made link open a new tab

Electogenius:
I wouldn’t want my code to disappear when I want to, say, look up a command or star the repo.

Note that I haven’t tested if this works yet but it probably works

### DIFF
--- a/templates/main.html
+++ b/templates/main.html
@@ -196,7 +196,7 @@
 <body onload="initCodeMirror(); decodeURL();  updateCount()">
     <textarea id=dummy style="visibility: hidden;position:absolute; height: 0px; font: size 20px;"></textarea>
     <session-code>{{session}}</session-code>
-    <h2 style="display: inline-block;"><a href="https://github.com/Vyxal/Vyxal">Vyxal</a></h2>
+    <h2 style="display: inline-block;"><a href="https://github.com/Vyxal/Vyxal" onclick="window.open('https://github.com/Vyxal/Vyxal')">Vyxal</a></h2>
     <button id="run_button" title="Run Program" type="button" style="color:white">
         <svg style="width:24px;height:24px" viewBox="0 0 24 24">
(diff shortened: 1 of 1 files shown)


ysthakur:
Sounds like a good idea to me.

vyxal-cookie (approved):
cookie

ysthakur (approved):

ysthakur:
@vyxal-cookie May I take your review as a serious one and merge this? I have no idea who you are but I'm assuming you're one of the actual contributors.

Electogenius:
Thanks!

Why’d the docs update fail? I don’t think adding an html attribute interferes with building docs, it says there was a 403 error fetching the git thingy.

(Also yes I don’t know who vyxal-cookie is 🤷‍♂️)

ysthakur:
@Electogenius Don't worry about that, it's just because the Update docs workflow pushes to master and it can't do that right now because it's a different branch. It always fails on PRs. (I shouldn't have run the workflows in the first place, there was nothing to test)

templates/main.html
```diff
@@ -196,7 +196,7 @@
 <body onload="initCodeMirror(); decodeURL();  updateCount()">
     <textarea id=dummy style="visibility: hidden;position:absolute; height: 0px; font: size 20px;"></textarea>
     <session-code>{{session}}</session-code>
-    <h2 style="display: inline-block;"><a href="https://github.com/Vyxal/Vyxal">Vyxal</a></h2>
+    <h2 style="display: inline-block;"><a href="https://github.com/Vyxal/Vyxal" onclick="window.open('https://github.com/Vyxal/Vyxal')">Vyxal</a></h2>
```
pxeger, line 199: What's wrong with `<a ... target="_blank">...</a>`?

ysthakur: That's what I found on SO too, but this works, so ¯\\_(ツ)_/¯.

Electogenius:
Thanks! I didn’t know `target="_blank"` was a thing (though i’d seen it before) so I guess I just learned something new hehe

PR #244, opened 2021-09-11
Docs

chunkybanana:


Lyxal (approved):

documents/knowledge/elements.yaml
```diff
@@ -2452,7 +2452,7 @@
 - element: "⇩"
   name: Grade Down
   arity: 1
-  description: Indices of elements to sort in descending order / lowercase / decrement number twice
+  description: 'Indices of elements to sort in descending order / lowercase / decrement number twice'
```
ysthakur, line 2455: Why the quotes? Was there an issue with the YAML?

documents/knowledge/elements.yaml
```diff
@@ -3179,10 +3179,411 @@
...
+
+- element: "øṁ"
+  name: Vertical Mirror
+  description: Vertical Mirror - Split by newlines, mirror each line, join by newlines
+  arity: 1
+  overloads:
+    str: vertical_mirror(a)
+  vectorise: true
+  tests:
+    # TODO: MULTILINE TESTCASSES
```
ysthakur, line 3567: YAML multiline strings can be done with `>-` if we want to turn the newlines into spaces or `|` if we want to keep the newlines.

ysthakur (approved):
Looks good, great work

documents/knowledge/elements.yaml
chunkybanana: Ok, I'll do this in a bit.

documents/knowledge/elements.yaml
chunkybanana: No clue.

PR #247, opened 2021-09-16
Allow Vyxal to be installed with pip install

ysthakur:
Made a pyproject.toml, a setup.cfg with requirements copied from requirements.txt, a boilerplate setup.py to just use setup.cfg, and a `vyxal/__main__.py` whose contents are the same as `vyxal/interpreter.py` and allow running Vyxal using `python -m vyxal`. This PR does not absolutely have to be merged because we may not want to mess with what is already broken, but this should certainly be added to the rewrite.


vyxal/__main__.py
```diff
@@ -0,0 +1,594 @@
+import os
+import secrets
+import sys
+
```
pxeger, line 4: why is this file here?

pxeger (changes requested):
a query

vyxal/__main__.py
ysthakur: So that it can be run with `python -m vyxal`. I guess calling it `vyxal.py` would have worked too, but I wasn't sure if that meant that any `import vyxal.asdf` import would require loading `vyxal.py` first.

pxeger: Why did you copy the contents of `interpreter.py` though? Just import it

Lyxal (changes requested):
basically what pxeger said about having __main__.py import interpreter.py

vyxal/__main__.py
ysthakur: Will the `if __name__ == "__main__":` stuff still be triggered, though?

pxeger: good point - how about you symlink `__main__.py` to `interpreter.py` - that'll work without duplicating the code.

ysthakur: Er, I'm not sure how to do that. How about a plain `main` method inside `interpreter.py` instead?

pxeger: on Linux or macOS: `ln -s interpreter.py __main__.py`; on Windows: [instructions](https://t.pxeger.com/xs.html?t=we+do+a+little+trolling&v=get+a+better+OS+lmao)

vyxal/__main__.py
```diff
@@ -1,594 +1,5 @@
...
-        elif vy_globals._join:
-            vy_print("\n".join([vy_str(n) for n in pop(vy_globals.stack)]))
-        elif flags and "J" in flags:
-            vy_print("\n".join([vy_str(n) for n in vy_globals.stack]))
-        else:
-            vy_print(pop(vy_globals.stack))
+from vyxal import interpreter
 
 
 if __name__ == "__main__":
```
pxeger, line 4: this is `__main__.py` - `__name__` will *always* be `__main__`.

pxeger (approved):

vyxal/__main__.py
ysthakur: bruh

Lyxal (approved):

PR #266, opened 2021-11-04
Make stack a field of context

cgccuser:
This is done so that `stack = asdf` works in transpiled code. Currently, it just seems to make a new local variable named `stack`.

### DIFF
--- a/documents/knowledge/processor.py
+++ b/documents/knowledge/processor.py
@@ -33,18 +33,18 @@ with open(TEST_ELEMENTS_PY, "w", encoding="utf-8") as tests:
                         unordered = (
                             len(expected) > 1
                             and expected[0] == '{' and expected[-1] == '}')
-                        tests.write(f"\tstack = {stack}\n")
-                        tests.write(f"\texpected = {expected}\n")
                         tests.write(f"\tctx = Context()\n")
+                        tests.write(f"\tctx.stack = {stack}\n")
+                        tests.write(f"\texpected = {expected}\n")
                         tests.write(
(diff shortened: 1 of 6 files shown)


Lyxal (changes requested):
Use `ctx.stacks[-1]` instead.

Lyxal:
I also think that having `stack` as an attribute of `Context` would screw up local stacks for functions - defining a new stack for a function would overwrite the existing stack.

Lyxal:
Because I actually tried doing this once, only to realise it was a bad idea

Lyxal:
I don't exactly remember why I thought it was a bad idea, but I do know that it lead me to adding `stacks` as an attribute of `Context`

cgccuser:
@Lyxal The context isn't supposed to be kept the same across functions, a copy of it should be made so that the stack (and, if we ever choose to interpret Vyxal, variables) isn't overwritten.

Lyxal:
I thought context was supposed to be the same between all functions as a way of removing the need for globals

cgccuser:
Oh, what I had in mind for it was keeping track of different scopes using different Context objects. Would a custom `ctx.stack` getter/setter work?

Lyxal:
Why not just let python handle things with its default scoping? That way, you don't need to assign stack to context, as local stacks are contained within their own scope.

cgccuser:
Because with stuff like `W`, a new local variable called `stack` is created. That messes up the tests, and I'm pretty sure it'll mess things up in lambdas too.

cgccuser:
You know what, never mind, let's just continue using `stack` and change the tests or modify `W` in some hacky way to make it work.

PR #268, opened 2021-11-10
Element implementation

Lyxal:
Lets get 2.6pre1 into base.


vyxal/LazyList.py
```diff
@@ -16,21 +16,19 @@
 def vyxalify(value: Any) -> Any:
     """Takes a value and returns it as one of the four types we use here."""
 
-    if isinstance(value, sympy.core.numbers.Integer):
+    if isinstance(value, (sympy.core.numbers.Integer)):
```
cgccuser, line 19: Why extra parentheses here btw?

cgccuser (approved):

vyxal/LazyList.py
Lyxal: Honestly I don't know.

razetime (approved):

PR #270, opened 2021-11-11
add "Number Stuff primitives from #160

razetime:
Requires the equation one, which will be added soon. I'd like some verification on the other primitives i've added.


vyxal/helpers.py
```diff
@@ -654,3 +654,16 @@ def wrapify(item: Any, count: int = None, ctx: Context = None) -> List[Any]:
             return [item]
         else:
             return item
+
+def pi_digits(x: int):
+    """Generate x digits of Pi. Spigot's formula."""
```
Lyxal, line 659: Don't forget to wrap this in a lazylist - pushing raw generators doesn't work. Also, this needs a context parameter if it's going to be used as an element function.

vyxal/elements.py
```diff
@@ -3013,6 +3013,46 @@ def zero_slice(lhs, rhs, ctx):
...
+    """Element ∆i
+    (int) -> nth_pi(a)
+    """
+    return pi_digits(lhs)[lhs]
+
+
+def nth_e(lhs, ctx):
+    """Element ∆ė
+    (int) -> nth_e(a)
+    """
```
Lyxal, line 3026: This should be made to vectorise over lists, which can be done by checking if `vy_type(lhs, simple=True) is list`

vyxal/elements.py
```diff
@@ -3013,6 +3013,46 @@ def zero_slice(lhs, rhs, ctx):
     }.get(ts, lambda: vectorise(zero_slice, lhs, rhs, ctx=ctx))()
 
 
+def nth_pi(lhs, ctx):
+    """Element ∆i
+    (int) -> nth_pi(a)
+    """
```
Lyxal, line 3019: Same thing about vectorising

Lyxal (changes requested):

vyxal/helpers.py
```diff
@@ -654,3 +654,21 @@ def wrapify(item: Any, count: int = None, ctx: Context = None) -> List[Any]:
             return [item]
         else:
             return item
+
+def pi_digits(x: int):
```
Lyxal, line 658: `def pi_digits(x: int, ctx: Context)`

vyxal/elements.py
```diff
@@ -3408,13 +3448,20 @@ def zero_slice(lhs, rhs, ctx):
...
+    "kṗ": process_element('LazyList(["()","[]","{}","<>"])', 0),
     "kS": process_element('"ඞ"', 0),
     "k₂": process_element("2 ** 20", 0),
     "k₃": process_element("2 ** 30", 0),
     "k∪": process_element('"aeiouy"', 0),
     "k⊍": process_element('"AEIOUY"', 0),
     "k∩": process_element('"aeiouyAEIOUY"', 0),
+    "∆i": process_element(nth_pi, 1),
+    "∆ė": process_element(nth_e, 1),
+    "∆I": process_element(lambda lhs, ctx: pi_digits(lhs), 1),
```
Lyxal, line 3460: `"∆I": process_element(pi_digits, 1),`

Lyxal (changes requested):

vyxal/helpers.py
razetime: fixed this thing

Lyxal (approved):

PR #285, opened 2021-11-28
run black on flask_app.py

pxeger:


pxeger:
Just tried re-running the check, but of course the workflow changes from #286 don't apply here...

pxeger:
probably should reorganise the imports in this as well, actually

cgccuser (approved):
I could've sworn I approved this already.

Lyxal (approved):

Lyxal:
Note that I moved the version stuff fully inside the version route because it was causing errors when running the flask app locally

PR #286, opened 2021-11-28
Exclude tests/test_elements.py from Black lint

pxeger:


pxeger:
I didn't bother doing this for flake8 as well, since noone seems to ever act on its linting suggestions anyway.

cgccuser (approved):
Looks good to me. Setting the line length in pyproject.toml will also save us from typing out `--line-length=80` each time. Thanks!

PR #313, opened 2021-12-02
Adding relative imports

chunkybanana:
Currently, Vyxal uses module imports - `import vyxal.elements` or whatever. This is, and will continue to be, an absolute nightmare for development as it uses the version of vyxal on pypi instead of what you're actually working on and want to test; or currently, nothing.

### DIFF
--- a/vyxal/__main__.py
+++ b/vyxal/__main__.py
@@ -1,4 +1,4 @@
-import vyxal.main
+import main
 import sys
 
 if len(sys.argv) > 1:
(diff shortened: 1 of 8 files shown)


pxeger:
Relative imports are strongly discouraged in Python. You must be doing something wrong if it's pulling a PyPI-installed version of `vyxal` rather than the one right there in your devtree. Either or both of these things may solve it:

- Make sure you use a [virtual environment](https://docs.python.org/3/tutorial/venv.html)
- Use `pip install -e .` to create and install a fake package that links to your development copy of the `vyxal` package (see [editable installs](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#development-mode))

PR #340, opened 2021-12-05
Add a few little things

Lyxal:


chunkybanana (approved):
Looks good to me, plus I really want sorting to be great again

vyxal/elements.py
```diff
@@ -2976,9 +2976,10 @@ def gen():
             found = 0
             item = 1
             while True:
+                res = safe_apply(function, item, ctx=ctx)
```
pxeger, line 2979: should this be called only after checking `found == count`, for maximum laziness?

vyxal/main.py
```diff
@@ -142,58 +142,59 @@ def execute_vyxal(file_name, flags, inputs, output_var=None, online_mode=False):
...
+                output = []
+            else:
+                stack.append(output)
+                output = join(stack, "\n", ctx)
+        elif flag == "…":
+            if vy_type(output, simple=True) is list:
+                output = output[:100]
+        else:
+            pass
+        if not (ctx.printed or "O" in flags) or "o" in flags:
```
pxeger, line 196: is this bit supposed to be inside the for loop?

pxeger (changes requested):

pxeger (approved):

PR #347, opened 2021-12-06
Remove unnecessary ctx argument to strip in P aka strip

cgccuser:
There were two extra `ctx` arguments given to `str.strip` inside `def strip()`/element `P`, this is just a trivial thing removing those. I haven't run any tests, but it probably works.

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -3130,7 +3130,7 @@ def strip(lhs, rhs, ctx):
         (NUMBER_TYPE, NUMBER_TYPE): lambda: vy_eval(
             vy_str(lhs).strip(vy_str(rhs))
         ),
-        (NUMBER_TYPE, str): lambda: vy_eval(vy_str(lhs).strip(rhs)),
+        (NUMBER_TYPE, str): lambda: vy_eval(vy_str(lhs).strip(rhs), ctx),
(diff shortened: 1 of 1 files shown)


Lyxal (approved):

vyxal/elements.py
```diff
@@ -3128,9 +3128,9 @@ def list_helper(left, right):
     ts = vy_type(lhs, rhs)
     return {
         (NUMBER_TYPE, NUMBER_TYPE): lambda: vy_eval(
-            vy_str(lhs).strip(vy_str(rhs), ctx)
+            vy_str(lhs).strip(vy_str(rhs))
         ),
-        (NUMBER_TYPE, str): lambda: vy_eval(vy_str(lhs).strip(rhs), ctx),
```
chunkybanana, line 3133: This one looks like it should stay as it's inside the call to `vy_eval`

chunkybanana (changes requested):
One of these looks like it should stay, as it's in a Vyxal function

Lyxal (changes requested):
I agree with EmanresuA

cgccuser:
Whoops

Lyxal (approved):
Lgtm

chunkybanana (approved):

PR #348, opened 2021-12-06
Use abs for factorial of negative numbers

cgccuser:
Use `factorial(abs(x))` so that negative numbers don't go boom.


Lyxal (approved):

pxeger:
Wouldn't it make more sense to use the pi function (i.e., the gamma function of `n+1`)? https://en.wikipedia.org/wiki/Gamma_function#Pi_function

PR #350, opened 2021-12-07
Fix #349

### LINKED ISSUE #349: Implicit range generation seems broken
idk why but it is

(created by lyxal [here](https://chat.stackexchange.com/transcript/message/59810710))

Lyxal:
Closes #349

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -4601,6 +4601,7 @@ elements: dict[str, tuple[str, int]] = {
     "k⟇": process_element("codepage", 0),
     "k½": process_element("LazyList([1,2])", 0),
     "kḭ": process_element("2 ** 32", 0),
+    "k₁": process_element("LazyList([1, 1])", 0),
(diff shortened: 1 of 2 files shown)


vyxal/elements.py
```diff
@@ -4600,6 +4601,7 @@ def zfiller(lhs, rhs, ctx):
     "k⟇": process_element("codepage", 0),
     "k½": process_element("LazyList([1,2])", 0),
     "kḭ": process_element("2 ** 32", 0),
+    "k₁": process_element("LazyList([1, 1])", 0),
```
cgccuser, line 4604: Should this element be included in this PR?

cgccuser (approved):
LGTM, looks like we need to be more careful about un-vyxalified and un-iterableified values.

vyxal/elements.py
Lyxal: Yes because I need it for an answer lol

chunkybanana (approved):

PR #353, opened 2021-12-08
Make the stack pad with 0s if no input offline, and fix an issue with set operators

### LINKED ISSUE #304: Stack isn't padded with zero
When no input is given to a program, the interpreter thinks that an empty string was given as input, and will return that when the stack is empty, instead of zero.

Example: [Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLCpD0iLCIiLCIiXQ==)

### LINKED ISSUE #352: `⊍` and co error when lists contain lists
This should be fixed by just making sure everything inside is a tuple or maybe a `LazyList`

(created by lyxal [here](https://chat.stackexchange.com/transcript/message/59821793))

Lyxal:
Closes #304 
Closes #352 

I also fixed a little division by 0 error I noticed - it's just implementing the behaviour that already exists in 2.4 and has proven to be useful: `0 / 0 = 0` for golfing purposes.

### DIFF
--- a/vyxal/helpers.py
+++ b/vyxal/helpers.py
@@ -206,7 +206,9 @@ def get_input(ctx: Context) -> Any:
         else:
             try:
                 temp = vy_eval(input("> " * ctx.repl_mode), ctx)
-            except EOFError:
+                if ctx.empty_input_is_zero and temp == "":
+                    return 0
+            except Exception:
                 temp = 0
             return temp
     else:
--- a/vyxal/context.py
+++ b/vyxal/context.py
@@ -6,6 +6,7 @@ class Context:
 
     def __init__(self):
         self.context_values = [0]
+        self.empty_input_is_zero = True
         self.default_arity = 1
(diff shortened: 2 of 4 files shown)


Lyxal:
https://chat.stackexchange.com/transcript/message/59823663#59823663 counts as 1 approval

vyxal/helpers.py
```diff
@@ -206,7 +206,9 @@ def get_input(ctx: Context) -> Any:
         else:
             try:
                 temp = vy_eval(input("> " * ctx.repl_mode), ctx)
-            except EOFError:
+                if ctx.empty_input_is_zero and temp == "":
+                    return 0
+            except Exception:
```
razetime, line 211: Is there good reason to use the more general Exception class here

razetime (commented):
just 1 question

PyGamer0 (approved):

chunkybanana (approved):

PR #360, opened 2021-12-10
fix 356, 357, 358 and 359

Lyxal:


chunkybanana (approved):
lgtm

pxeger:
In future can you try and fix one bug at a time and open separate PRs for each?

vyxal/transpile.py
```diff
@@ -90,9 +90,9 @@ def transpile_token(
         # screws up escape sequences.
 
         # So instead, we have to manually escape the string
-        string = string.replace("\\", "\\\\").replace('"', '\\"')
-        string = string.replace("\n", "\\n")
-        return indent_str(f'stack.append("{string}")', indent)
+        string = f"{string!r}"
+        string = re.sub(r"\\([ntbrf])", r"\1", string).replace("\\`", "`")
```
pxeger, line 94: I'm pretty sure there should be a better way of doing this, but meh. Also, I think you should write some unit tests for this particular bit of code, because......

pxeger (approved):

PR #361, opened 2021-12-10
Update production (10/12/2021)

Lyxal:
Am I even using production correctly by updating it immediately after merging into main?


pxeger:
> Am I even using production correctly by updating it immediately after merging into main?

Yes. (well, you might want to wait a while to ensure the `main` at that point is stable, but that's kinda overkill)

pxeger (approved):

PR #372, opened 2021-12-11
Add some little features and add some overloads

Lyxal:


razetime (approved):
it's the little things that count

documents/knowledge/elements.yaml
```diff
@@ -1027,18 +1027,18 @@
...
-    num-str: every ath character of b
+    str-num: repeat characters of a until a is length b
+    num-str: repeat characters of b until b is length a
     str-str: regex.search(pattern=a, string=b).span() (Length of regex match)
   vectorise: true
   tests:
     - "[5,3] : 125"
     - "[0,0] : 1"
-    - '["hello",2] : "hlo"'
+    - '["hello",7] : "hellohe"'
```
cgccuser, line 1041: What happens for negative numbers? Would `["hello", -7]` become `"eholleh"`?

vyxal/elements.py
```diff
@@ -936,16 +936,16 @@ def expe_minus_1(lhs, ctx):
...
+    (str, num) -> repeat characters of a until a is length b
     (str, str) -> regex.search(pattern=a, string=b).span() (Length of regex match)
     """
 
     ts = vy_type(lhs, rhs)
     return {
         (NUMBER_TYPE, NUMBER_TYPE): lambda: lhs ** rhs,
-        (NUMBER_TYPE, str): lambda: rhs[:: int(lhs)],
-        (str, NUMBER_TYPE): lambda: lhs[:: int(rhs)],
+        (NUMBER_TYPE, str): lambda: (rhs * lhs)[:lhs],
```
cgccuser, line 947: Multiplying by the number will get a pretty long string (and won't work for negative numbers). How about only getting as many whole strings as you need and then concatenating a slice?

vyxal/elements.py
```diff
@@ -1163,11 +1163,26 @@ def from_base(lhs, rhs, ctx):
 
 def gen_from_fn(lhs, rhs, ctx):
     """Element Ḟ
-    (fun, lst) -> Generator from a with initial vector b
-    (lst, fun) -> Generator from b with initial vector a
+    (num, num) -> sympy.N(a, b)
```
cgccuser, line 1166: Maybe add a description of what `N` itself does here? (perhaps `evaluate a to b decimal digits, then convert to string`)

vyxal/elements.py
```diff
@@ -3908,7 +3923,7 @@ def vy_print(lhs, end="\n", ctx=None):
         res = lhs(ctx.stacks[-1], lhs, ctx=ctx)[-1]
         vy_print(res, ctx=ctx)
     else:
-        if ts == NUMBER_TYPE:
+        if is_sympy(lhs):
             if ctx.print_decimals:
                 lhs = eval(sympy.pycode(sympy.nsimplify(lhs)))
```
cgccuser, line 3928: This looks unnecessarily complicated. Can't `sympy.N` be used here?

cgccuser (changes requested):
Almost ready to merge, I just have some minor questions

documents/knowledge/elements.yaml
razetime: considering that length can't be negative it should probably error

cgccuser: Well, this is a golflang, so why not stuff some useful behavior in there instead of erroring? :P (although the useful behavior should probably make sense, I guess (which I feel reversing does))

chunkybanana: As I said in chat, this overload should go to `Ẏ` as it's a logical extension of its current functionality.

documents/knowledge/elements.yaml
```diff
@@ -1979,8 +1979,22 @@
   arity: 2
   description: Make a generator from function a with initial vector b.
```
chunkybanana, line 1980: Change the description?

chunkybanana (dismissed):
Overloads are a bit of a mess, we have logical places to put all of these so let's use them.

vyxal/elements.py
Lyxal: The problem with that is that it's fixed length and can have too many leading 0s sometimes. This way allows python to determine precision.

documents/knowledge/elements.yaml
```diff
@@ -1031,14 +1031,14 @@
   arity: 2
   overloads:
     num-num: a ** b (exponentiation)
-    str-num: repeat characters of a until a is length b
-    num-str: repeat characters of b until b is length a
+    str-num: append a[0] until a is length b (spaces are used if a is empty)
```
cgccuser, line 1034: How useful is this overload?

cgccuser (approved):

documents/knowledge/elements.yaml
Lyxal: pyth has it, so it's useful.

PR #373, opened 2021-12-11
update them docs

Lyxal:
to be merged along side #372

### DIFF
--- a/documents/knowledge/Elements.txt
+++ b/documents/knowledge/Elements.txt
@@ -659,11 +659,11 @@ d (Double)
 -------------------------------
 e (Exponentiation)
 
-- Exponentiate two numbers / get every nth character of a string / get length of a regex match
+- Exponentiate two numbers / extend string / get length of a regex match
 
     num a, num b: a ** b (exponentiation)
-    str a, num b: every bth character of a
-    num a, str b: every ath character of b
+    str a, num b: append a[0] until a is length b (spaces are used if a is empty)
+    num a, str b: append b[0] until b is length a (spaces are used if b is empty)
     str a, str b: regex.search(pattern=a, string=b).span() (Length of regex match)
     otherwise: vectorise
 -------------------------------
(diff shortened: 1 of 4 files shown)


documents/knowledge/Elements.txt
```diff
@@ -2118,7 +2118,7 @@ k□ (Directions)
     num a, num b: roots(a, b) / x^2 + ax + b = 0
     num a, str b: evaluate single variable expression b with x=a
     str a, num b: evaluate single variable expression a with x=b
-    str a, str b: solve equation a < b for x
+    str a, str b: solve equations a and b simultaneously
```
cgccuser, line 2121: The "simultaneously" here sounds a teeny bit odd. Maybe clarify that they're a system of equations or something like that?

cgccuser (approved):

documents/knowledge/Elements.txt
Lyxal: Where I'm from, they're called simultaneous equations, so that's why I used that term,

documents/knowledge/Elements.txt
```diff
@@ -659,11 +659,11 @@ d (Double)
 -------------------------------
 e (Exponentiation)
 
-- Exponentiate two numbers / get every nth character of a string / get length of a regex match
+- Exponentiate two numbers / extend string / get length of a regex match
 
     num a, num b: a ** b (exponentiation)
-    str a, num b: every bth character of a
-    num a, str b: every ath character of b
+    str a, num b: append a[0] until a is length b (spaces are used if a is empty)
```
cgccuser, line 665: Wait, isn't this change already done in the other PR?

cgccuser (changes requested):
We might have merge conflicts if we update the definition of `e` here too

documents/knowledge/Elements.txt
Lyxal: yes, but it's the exact same.

Lyxal:
> We might have merge conflicts if we update the definition of `e` here too

well yes but actually no.

I made sure there won't be.

chunkybanana (approved):
All of this is implemented, right?

Lyxal:
> All of this is implemented, right?

of course they are. all implemented in #372

cgccuser (approved):

PR #374, opened 2021-12-11
Use 'vyxal.encoding.codepage' instead of 'codepage' for k⟇

cgccuser:
`k⟇` was implemented as `codepage` instead of `vyxal.encoding.codepage` (which is not imported). Trivial fix.


chunkybanana:
This is such a trivial change that I'll just forcemerge it.

PR #376, opened 2021-12-12
fix a resizing issue that's been annoying me for a while

Lyxal:


cgccuser:
Not working for me, the code box is always about 2 lines tall, and once I have enough letters to reach the second line, there's a scroll bar, but the box doesn't resize.

cgccuser (approved):
Feels weird to have this heuristic-y thing, but it does work better than what we have currently so 👍.

PyGamer0 (approved):
idk js but looks fine

razetime (approved):

static/main.js
```diff
@@ -262,8 +262,9 @@ function initCodeMirror() {
         dummy.value = elem.doc.getValue()
         elem.setSize(
             null,
-            Math.max(dummy.scrollHeight - 5, elem.getTextArea().dataset.baseHeight || 27)
```
pxeger, line 265: does this mean the `dummy` element is no longer needed at all?

pxeger (changes requested):

static/main.js
Lyxal: No. It's still needed to set the other styles of the codemirror boxes.

PR #387, opened 2021-12-13
Potentially fix execing things in main

### LINKED ISSUE #386: Fix the `exec`
TODO: more info

(created by lyxal [here](https://chat.stackexchange.com/transcript/message/59863568))

Lyxal:
Closes #386

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup
 
 setup(
     name="vyxal",
-    version="2.6.1",
+    version="2.7.0",
     description="The Vyxal Programming Language",
     url="https://github.com/Vyxal/Vyxal",
     author="Vyxal Organisation",
--- a/vyxal/main.py
+++ b/vyxal/main.py
@@ -143,7 +143,8 @@ def execute_vyxal(file_name, flags, inputs, output_var=None, online_mode=False):
 
     ctx.stacks.append(stack)
     try:
-        exec(code)
+        locals()["stack"] = stack
+        exec(code, locals() | globals())
     except Exception as e:
         if ctx.online:
             ctx.online_output[2] += "\n" + traceback.format_exc()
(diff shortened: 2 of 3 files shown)


PyGamer0 (approved):

Tams-Tams (approved):
I dunno, looks ok. This counts as 1/2 approved.

AMiller42 (approved):
imo, lgtm, but idk tbh.

setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name="vyxal",
-    version="2.6.1",
+    version="2.7.0",
```
pxeger, line 5: a version bump should be separate from any other changes

vyxal/main.py
```diff
@@ -143,7 +143,8 @@ def execute_vyxal(file_name, flags, inputs, output_var=None, online_mode=False):
 
     ctx.stacks.append(stack)
     try:
-        exec(code)
+        locals()["stack"] = stack
```
pxeger, line 146: assigning to `locals()[]` is technically undefined behaviour in Python

vyxal/main.py
```diff
@@ -143,7 +143,8 @@ def execute_vyxal(file_name, flags, inputs, output_var=None, online_mode=False):
...
+        exec(code, locals() | globals())
```
pxeger, line 147: Merging these creates a new object, so any changes to the new object don't affect the old `locals()` or `globals()`. (i.e., any assignments inside `code` won't affect any variables in `execute_vyxal()`). Is this what you want?

If not, use a `collections.chainmap`.

pxeger (changes requested):

setup.py
Lyxal: I figured that too, but I realised I forgot to do it before actually making the release notes so I quickly placed it here.

vyxal/main.py
Lyxal: Looks like it isn't needed anyway. Good catch.

vyxal/main.py
Lyxal: I'm fine with it not actually changing locals or global. In fact, that's probably better in the bigger picture.

cgccuser (approved):

pxeger (approved):

PR #391, opened 2021-12-14
Update docs for `İ` and make the implementations of ẋ and İ good.

Lyxal:


razetime (approved):

vyxal/elements.py
```diff
@@ -2835,8 +2842,10 @@ def repeat(lhs, rhs, ctx):
         @lazylist
         def gen():
             prev = value
+            val = value
```
pxeger, line 2845: Why are there variables called `val` and `value`? Could they be called something more descriptive please?

vyxal/elements.py
```diff
@@ -2835,8 +2842,10 @@ def repeat(lhs, rhs, ctx):
...
             while True:
-                val = safe_apply(function, value, ctx=ctx)
+                val = safe_apply(function, val, ctx=ctx)
+                print(val)
```
pxeger, line 2848: Please avoid leaving `print`s lying around

pxeger (changes requested):

vyxal/elements.py
Lyxal: Good catch - I missed that bit of debugging last night.

vyxal/elements.py
```diff
@@ -1478,18 +1478,25 @@ def index_indices_or_cycle(lhs, rhs, ctx):
     (any, fun) -> Repeatedly apply b to a until cycle is formed, then
```
cgccuser, line 1478: Llooks like `İ` has retained the same behavior despite the change in docs, but if you're updating elements.yaml, it'd be good to change this too, for consistency.

vyxal/elements.py
Lyxal: This has been clarified. Thank you for noticing this. :)

cgccuser (changes requested):

cgccuser (approved):

pxeger (approved):

PR #394, opened 2021-12-15
Hide generated elements files from GitHub diffs

pxeger:
Whenever `elements.yaml` changes, I always see `Elements.txt` first and comment on that before realising it's not the source file. By marking these files as `linguist-generated=true` in a `.gitattributes` file, they will be excluded from GitHub diffs, and this won't happen.

See https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+documents/knowledge/elements.md linguist-generated=true
+documents/knowledge/Elements.txt linguist-generated=true
+static/parsed_yaml.js linguist-generated=true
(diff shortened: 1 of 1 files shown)


pxeger:
Should I add `tests/test_elements.py` to this list as well?

Lyxal (approved):

chunkybanana:
> Should I add `tests/test_elements.py` to this list as well?

Yes.

cgccuser (approved):

PR #410, opened 2021-12-16
Add eval as JS option

### LINKED ISSUE #407: `E` -  Eval as JS

pxeger:
Closes #407.

Hopefully this series of popups is good enough to prevent self-XSS.


chunkybanana (approved):

Lyxal (approved):

Lyxal (commented):
Maybe have a line that says "don't execute JavaScript from untrusted sources" like Google does if you open developer tools

pxeger:
I think this is good enough to basically cover that.

PR #418, opened 2021-12-16
Update Production (17/12/2021)

Lyxal:


cgccuser:
We really need to get vyxaldev.pythonanywhere.com working so we can properly test stuff online before merging. Testing the flask site locally is mildly annoying.

pxeger:
Don't merge until #421 is merged

vyxal/transpile.py
```diff
@@ -91,21 +91,20 @@ def transpile_token(
 
         # So instead, we have to manually escape the string
         string = f"{string!r}"
-        string = re.sub(r"\\([ntbrf])", r"\1", string).replace("\\`", "`")
+        string = re.sub(r"\\?\\([ntbrf])", r"\\\1", string).replace("\\`", "`")
```
pxeger, line 94: how many more times do you think we'll have to change this line before we decide to implement a proper string escaping system?

pxeger (approved):

vyxal/transpile.py
Lyxal: At least 3

PR #421, opened 2021-12-17
implement a whole bunch of stuff

### LINKED ISSUE #399: `¨ẇ` - N-wrap - pop an item, and wrap that many items from the stack into a list.

### LINKED ISSUE #401: `ÞM` - Indices of maximal elements (stolen from Jelly)

### LINKED ISSUE #402: `∆ṁ` - Median

### LINKED ISSUE #403: `∆M` - Mode

### LINKED ISSUE #404: `∆o` - nth-ordinal (first, second, third)

### LINKED ISSUE #405: `∆ċ` - nth-cardinal (minus one, zero, one, two, three)
Given a number, return that number as a  cardinal - minus one, zero, one, two, three etc

### LINKED ISSUE #416: Wrapping the stack breaks with functions
If you try to wrap the stack when there is a function on the stack, it clears the stack before wrapping.

Example: [Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLOuzEyMzsgMSAyIDMgNCBXIiwiIiwiIl0=)

Lyxal:
Closes #399
Closes #401 
Closes #402 
Closes #403 
Closes #404 
Closes #405 
Closes #416


chunkybanana (approved):

documents/knowledge/elements.yaml
```diff
@@ -3752,6 +3752,62 @@
...
 
+- element: "∆ċ"
+  name: Nth Cardinal
+  arity: 1
+  description: Get the nth cardinal / convert number to words
+  overloads:
+    num: num_to_words(a)
+  vectorise: true
+  tests:
+    - "[0] : 'zero'"
```
cgccuser, line 3763: Rather than have all these test cases for simple numbers, I think we should a) smoosh them together into a single testcase on a list (because it vectorises) and b) remove a couple of these and use edgier cases instead. For example, what happens for 1103? -1103? 1.2?

documents/knowledge/elements.yaml
```diff
@@ -3752,6 +3752,62 @@
...
+- element: "∆ṁ"
+  name: Median
+  arity: 1
+  description: Get the median of a list
+  overloads:
+    lst: median(a)
+  vectorise: false
+  tests:
+    - "[[1, 2, 3, 4, 5]] : 3"
+    - "[[1, 2, 3, 4, 5, 6]] : [3, 4]"
```
cgccuser, line 3809: Hrm, perhaps we could note in the description that this doesn't return the average of the middle two values if there's an even number of values but rather returns both of them in a list?

documents/knowledge/elements.yaml
```diff
@@ -4345,6 +4401,16 @@
   tests:
     - "[[1, 2, 3, 4, 5, 6, 7, 8, 9], [[1], [1, 2], [1, 2, 3], [1], [1, 2], [1, 2, 3]]] : [[1], [2, 3], [4, 5, 6], [7], [8, 9]]"
 
+- element: "ÞM"
+  name: Maximial Indicies
```
cgccuser, line 4405: Minor typo: Maximial -> Maximal

vyxal/elements.py
```diff
@@ -2432,6 +2498,8 @@ def one_slice(lhs, rhs, ctx):
     (str, str) -> re.match(pattern=a,string=b)
     """
 
+    # no, not one_shot, one_slice.
```
cgccuser, line 2501: ???

vyxal/elements.py
```diff
@@ -4617,6 +4691,7 @@ def zfiller(lhs, rhs, ctx):
     "ÞC": process_element(foldl_columns, 2),
     "ÞR": process_element(foldl_rows, 2),
     "Þṁ": process_element(mold_special, 2),
+    "ÞM": process_element(maximal_indicies, 1),
```
cgccuser, line 4694: indicies should be indices

cgccuser (changes requested):
Couple minor changes (and a couple comments that are just comments and not requested changes)

vyxal/elements.py
pxeger: probably copilot being funky

Lyxal: that's a reference to the game OneShot

documents/knowledge/elements.yaml
pxeger: Or - get this - it could just correctly implement median by returning the average of those two?

Lyxal: @pxeger but then what about strings? how do you average two strings?

cgccuser: @pxeger I feel like this is more useful behavior in case you want to get the lower/higher element. It's trivial to average them after that.

Lyxal: correct - `ṁ` exists to average a list

documents/knowledge/elements.yaml
Lyxal: fixed

vyxal/elements.py
Lyxal: fixed

documents/knowledge/elements.yaml
Lyxal: fun fact: copilot made the cases lol

documents/knowledge/elements.yaml
pxeger: But if you don't know whether it's gonna be two or one, then `ṁ` won't help (because it returns 0 when given a list).

pxeger (approved):

documents/knowledge/elements.yaml
Lyxal: done.

documents/knowledge/elements.yaml
Lyxal: clarification made.

cgccuser (approved):

PR #433, opened 2021-12-18
fix some documentation

Lyxal:


documents/knowledge/elements.yaml
```diff
@@ -3695,6 +3697,17 @@
     - "[8] : 5"
     - "[9] : 3"
 
+- element: "∆Ė"
+  name: N Digits of E / Sympy Evaluate
```
pxeger, line 3701: Specify that `E` is Euler's constant? It took me a while to work out the it wasn't an argument. Will also help with searching

vyxal/elements.py
```diff
@@ -278,6 +278,7 @@ def apply_at(lhs, rhs, other, ctx):
 def arccos(lhs, ctx):
     """Element ∆C
     (num) -> arccos(lhs)
+    (str) -> arccos(expression)
```
pxeger, line 281: How come you're not adding this for all the other functions like this?

pxeger (changes requested):

razetime (approved):

documents/knowledge/elements.yaml
```diff
@@ -3474,9 +3474,9 @@
     - "[1.5] : 0"
 
 - element: "∆e"
-  name: e raised to power a
+  name: Euler's raised to power a
```
cgccuser, line 3477: Euler's should be Euler's number. Also, can't we just call it e instead?

documents/knowledge/elements.yaml
```diff
@@ -3474,9 +3474,9 @@
...
   arity: 1
-  description: Get the value of e raised to the power of a
+  description: Get the value of Euler's numberraised to the power of a
```
cgccuser, line 3479: Space needed between Euler's number and raised, but why not say e instead?

vyxal/elements.py
```diff
@@ -796,11 +800,14 @@ def dyadic_minimum(lhs, rhs, ctx):
 def e_digits(lhs, ctx):
     """Element ∆Ė
     (int) -> e_digits(a)
+    (str) -> evaluate as sympy
     """
     if vy_type(lhs) == NUMBER_TYPE:
-        estr = str(sympy.N(sympy.E, int(lhs) + 1))
+        estr = str(sympy.N(sympy.E, int(lhs) + 2))
```
cgccuser, line 806: What was this for btw? (this is a comment, not a request for a change, feel free to resolve it)

cgccuser (dismissed):
I don't understand why we're calling it Euler's number rather than e. It's certainly easier to search for, but then why not have both? e is more recognizable than Euler's number.

documents/knowledge/elements.yaml
cgccuser: [Euler's constant](https://en.wikipedia.org/wiki/Euler's_constant) is another number entirely. It's kinda confusing, and I really like `e` better than Euler's number because of that.

vyxal/elements.py
Lyxal: Because it was returning `3` for `0`

PyGamer0 (approved):

pxeger (approved):

PR #450, opened 2021-12-20
Handle reversed ranges for ṡ

cgccuser:
Just added an else branch that handles cases when rhs < lhs


chunkybanana (approved):

Lyxal (approved):
Funny thing is that I already had this on a local branch. Protip: cast the ranges to lazy list at some point.

PR #458, opened 2021-12-20
Add a whole bunch of new and exciting math overloads

### LINKED ISSUE #442: Totient function builtin

Lyxal:
Also update the docs for some

Closes #442


PyGamer0 (approved):
very cool math

vyxal/elements.py
```diff
@@ -2546,6 +2552,33 @@ def pluralise_count(lhs, rhs, ctx):
...
+    ts = vy_type(lhs)
+    x = sympy.symbols("x")
+    return {
+        NUMBER_TYPE: lambda: str(
+            sum(map(lambda arg: x ** arg, range(0, lhs + 1)))
+        ),
+        str: lambda: lhs,
+        list: lambda: str(
+            sum(
+                map(
```
pxeger, line 2573: this would look much cleaner as a for comprehension

vyxal/helpers.py
```diff
@@ -294,6 +294,28 @@ def levenshtein_distance(s1: str, s2: str) -> int:
...
 
+def local_minima(lhs: str) -> List[Union[int, float]]:
+    """Find the local minima of a mathematical function using Sympy"""
+
+    x = sympy.symbols("x")
+    d_dx = sympy.diff(sympy.sympify(lhs), x)
+    second_dx = sympy.diff(d_dx, x)
+    zeros = sympy.solve(d_dx, x)
+
+    return LazyList(z for z in zeros if second_dx.subs(x, z) > 0)
```
pxeger, line 305: There is also a local minimum if `f''(x) = 0` and `f'''(x) > 0`, right? And so on. [This](https://gist.github.com/tcibinan/8b7b58c6861a03752f8ec309e809d770) works.

vyxal/helpers.py
```diff
@@ -616,6 +638,16 @@ def simplify(value: Any) -> Union[int, float, str, list]:
...
 
 
+def stationary_points(lhs: str) -> List[Union[int, float]]:
+    """Returns a list of stationary points of a mathematical function"""
+
+    x = sympy.symbols("x")
+    d_dx = sympy.diff(sympy.sympify(lhs), x)
+    zeros = sympy.solve(d_dx, x)
+
+    return LazyList(z for z in zeros)
```
pxeger, line 648: just do `LazyList(zeros)`

pxeger (changes requested):

vyxal/helpers.py
Lyxal: The way I learned it, you only need the first and second derivatives.

documents/knowledge/elements.yaml
```diff
@@ -3597,11 +3601,12 @@
     - "[360] : 6.283185307179586"
 
 - element: "∆Ṗ"
-  name: Next Prime After a Number
+  name: Next Prime After a Number / Discrimant of Polynomial
```
cgccuser, line 3602: Discriminant

documents/knowledge/elements.yaml
```diff
@@ -3597,11 +3601,12 @@
...
   arity: 1
-  description: Get the next prime number after a given number
+  description: Get the next prime number after a given number / the discrimant of a polynomial
```
cgccuser, line 3604: Again, discriminant

documents/knowledge/elements.yaml
```diff
@@ -3610,13 +3615,15 @@
     - "[4] : 5"
     - "[5] : 7"
     - "[69] : 71"
+    - "['3 * x ** 2 + 493 * x - 2319'] : 270877"
 
 - element: "∆ṗ"
-  name: First Prime Before a Number
+  name: First Prime Before a Number / Factorise Expression
```
cgccuser, line 3619: Factorise Expression -> Factor Expression? "Factor" is enough afaik, and it's compatible with both British and American English :P

vyxal/elements.py
```diff
@@ -3081,7 +3115,12 @@ def strict_greater_than(lhs, rhs, ctx):
         (NUMBER_TYPE, str): lambda: int(str(lhs) > rhs),
         (str, NUMBER_TYPE): lambda: int(lhs > str(rhs)),
         (str, str): lambda: int(lhs > rhs),
-    }.get(ts, lambda: int(list(lhs) > list(rhs)))()
+    }.get(
+        ts,
+        lambda: int(
+            bool(list(iterable(lhs, ctx=ctx)) > list(iterable(rhs, ctx=ctx)))
```
cgccuser, line 3131: Can we just use a normal function with if-else branches here? The `lambda :`'s are okay in other elements but this is getting out of hand.

vyxal/elements.py
```diff
@@ -3910,14 +3966,14 @@ def vy_repr(lhs, ctx):
         # actually make the repr kinda make sense
     }.get(
         ts,
-        lambda: "⟨ "
-        + " | ".join(
+        lambda: ("⟨ " if ctx.vyxal_lists else "[")
```
cgccuser, line 3979: vy_str and vy_repr seem to repeat code for lists. Maybe extract this bit into a helper or something?

cgccuser (changes requested):
Great changes, just a few minor things

vyxal/elements.py
Lyxal: It only looks complicated because of the type changing - an if-else would be overkill for this.

vyxal/elements.py
Lyxal: It'd be too specialised for a helper - also it's purely for flag purposes.

vyxal/elements.py
```diff
@@ -2564,17 +2581,10 @@ def polynomial_expr_from_coeffs(lhs, ctx):
...
+        NUMBER_TYPE: lambda: str(sum(x ** arg for arg in range(0, lhs + 1))),
         str: lambda: lhs,
         list: lambda: str(
-            sum(
-                map(
-                    lambda arg: arg[1] * x ** arg[0],
-                    enumerate(reverse(lhs, ctx)),
-                )
-            )
+            sum(arg[1] * x ** arg[0] for arg in enumerate(reverse(lhs, ctx)))
```
pxeger, line 2587: now you're not using a lambda, you can eliminate the tuple indexing:
```suggestion
            sum(c * x ** i for i, c in enumerate(reverse(lhs, ctx)))
```

documents/knowledge/elements.yaml
```diff
@@ -4509,4 +4520,5 @@
   vectorise: false
   tests:
      - "[[1, 1, 1], [9, 9, 9]] : 1"
-     - "[[1, 2, '3'], [1, 2, '2']] : 0"
\ No newline at end of file
+     - "[[1, 2, '3'], [1, 2, '2']] : 0"
+
```
pxeger, line 4524: Now there are two blank lines here! Most editors put one in by default, so adding one manually ends up with 2. But some editors don't, so sometimes you have to add one. Anyway, don't worry about it here, I'll put in a check and an auto-fixer

pxeger (changes requested):

vyxal/helpers.py
pxeger: @Lyxal I bet you this'll come back to bite as a bug at some point...

documents/knowledge/elements.yaml
Lyxal: That probably will work better than me trying to fix the newlines manually :p

pxeger (approved):

documents/knowledge/elements.yaml
pxeger: Factorise is what we say in BrE. "Factor" sounds very American and very wrong to me. And since `factorise` is a superset of the letters of `factor`, this is better for searchability

documents/knowledge/elements.yaml
```diff
@@ -3529,22 +3529,25 @@
     - "[17] : 2.833213344056216"
 
 - element: "∆l"
-  name: Logarithm (log_2)
+  name: Logarithm (log_2) / Staionary Points
```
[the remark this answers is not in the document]
pxeger: You didn't fix this
```suggestion
  name: Logarithm (log_2) / Stationary Points
```

documents/knowledge/elements.yaml
pxeger:
```suggestion
  name: Next Prime After a Number / Discriminant of Polynomial
```

pxeger (changes requested):
sorry, un-approving, because there are still typos

documents/knowledge/elements.yaml
pxeger:
```suggestion
  description: Get the next prime number after a given number / the discriminant of a polynomial
```

pxeger (approved):

Lyxal:
> sorry, un-approving, because there are still typos

I had them fixed locally a few hours ago, but I forgot to push lol

pxeger:
Ah I see lol

pxeger (approved):

PR #466, opened 2021-12-21
Move protocols files into root folder to make visible to Github

cgccuser:
I also moved the code guide even if Github doesn't acknowledge it.


Lyxal (approved):

pxeger:
btw github would still recognise these if they were in a `docs/` directory, not only in the root

pxeger (approved):

Issue #379, opened 2021-12-12
Tabs and newlines don’t work in strings


AMiller42:
Tabs and newlines in strings are replaced with `t` and `n`, respectively.

Example: [Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCJgXHRcdFx0XHRcblxuXG5cbmAiLCIiLCIiXQ==)

PR #481, opened 2021-12-23
Another day another bug-fixing feature-adding PR

### LINKED ISSUE #478: Function arguments loop incorrectly
When implicitly getting the arguments of a function, the arguments are retrieved in the wrong order once it starts to loop.<br><br>Example: <a href="https://vyxal.pythonanywhere.com/#WyIiLCIiLCIxIDIgMyDOuzN8LCwsLCwsLCwsO+KAoCIsIiIsIiJd" rel="nofollow noopener noreferrer">Try it Online!</a>

(created by Aaroneous Miller [here](https://chat.stackexchange.com/transcript/message/59957998))

### LINKED ISSUE #479: Escaping bad
Escaping bad.

`` `\\` `` should return `\`, [but it don't.](https://vyxal.pythonanywhere.com/#WyIiLCIiLCJgXFxcXGAiLCIiLCIiXQ==)

`` `\\n` `` should return `\n`, [but it don't.](https://vyxal.pythonanywhere.com/#WyIiLCIiLCJgXFxcXG5gIiwiIiwiIl0=)

Each backslash escapes the next character, even if it is being escaped itself. Fix it.

Lyxal:
Fun for the whole family.

Closes #478
Closes #479


vyxal/dictionary.py
```diff
@@ -23102,7 +23102,17 @@
 mne
 mni
 mnu
-mny""".split(
+mny
+Rick Astley
+rick astley
+rickroll
+Rickroll
+pxeger
```
pxeger, line 23110: highly pog

vyxal/transpile.py
```diff
@@ -89,9 +89,17 @@ def transpile_token(
...
+        temp = ""
+        string = collections.deque(string)
+        while string:
+            char = string.popleft()
+            if char == "\\" and string:
+                temp += "\\" + string.popleft()
+            elif char == '"':
+                temp += '\\"'
+            else:
+                temp += char
```
pxeger, line 101: Use an iterator instead:
```suggestion
        iterator = iter(string)
        for char in iterator:
            if char == "\\":
                temp += "\\" + next(iterator, "")
            elif char == '"':
                temp += '\\"'
            else:
                temp += char
```

pxeger: I still don't know what escape sequences you [think](https://github.com/Vyxal/Vyxal/commit/e6a4030cc649a0ba750b7904d0dc4539aacbe60e) `{string!r}` (which is the same as `{repr(string)}`) breaks.

Also, won't this reintroduce #253 and/or #359 and/or #379?

pxeger (changes requested):

vyxal/transpile.py
Lyxal: If you have something like `\n` in a string, `{string!r}` changes it to `\\n`

Lyxal: And it shouldn't reintroduce any of those issues - #253 was to do with the fact that the D flag used to be handled differently, #359 is because of the way raw strings work and #379 was because of the regex/substitution I tried to use.

Lyxal: After some testing #253 was reintroduced. But that's fixed.

pxeger:
Still doesn't work.
```
`this is a string with a null byte in it --> \0 <--
```
```
  ...
  File ".../vyxal/main.py", line 147, in execute_vyxal
    exec(code, locals() | globals())
ValueError: source code string cannot contain null bytes
```


I still feel like escape sequence handling should be done entirely in the parser. If you did that, you could just use `repr`/`!r` in the transpilation and guarantee no ACE exploits (they would manifest as parser bugs instead, which are more tame).

pxeger:
By the way, can you try to give your branches more descriptive names (e.g. `fix-478-and-479`) rather than `thursday-fixes`, because if you reuse a branch name (say, next thursday), it can confuse git, and it confuses me, because I don't know what the branch actually contains, without  `git checkout`ing it.

Lyxal:
> ```
> `this is a string with a null byte escape in it --> \0 <--
> ```

Well of course it won't work - the version online doesn't have the changes from this commit.

Lyxal:
It works with the changes on this branch

Lyxal:
image

pxeger:
Oops I was using it from the command line and forgetting that `echo` evaluates escape sequences in zsh

Lyxal:
> By the way, can you try to give your branches more descriptive names (e.g. `fix-478-and-479`) rather than `thursday-fixes`, because if you reuse a branch name (say, next thursday), it can confuse git, and it confuses me, because I don't know what the branch actually contains, without `git checkout`ing it.

Sure, I'll do that from now on. I thought it'd be a nice thematic thing for just this week, but you're right that the meaning can be unclear

vyxal/dictionary.py
```diff
@@ -23167,7 +23177,44 @@
...
+<h1>
+</h1>
+<h2>
+</h2>
+<div>
+</div>
+<body>
+<svg>
+</body>
+</svg>
```
vyxal-cookie, line 23207: All of the tags have the closing tag immediately after, except for `<body>` and `<svg>`

vyxal-cookie (changes requested):

vyxal/dictionary.py
Lyxal: ...

That's not an issue at all. All it does it change the index in the short dictionary.

vyxal/dictionary.py
```diff
@@ -23167,7 +23177,44 @@
...
+</s>
+<strong>
+</strong>
+<em>
+</em>
+<a href="
+</a>
+<img src="
+</img>
+</br>
```
pxeger, line 23193:
```suggestion
<br/>
```

vyxal/dictionary.py
```diff
@@ -23167,7 +23177,44 @@
...
+<s>
+</s>
+<strong>
+</strong>
+<em>
+</em>
+<a href="
+</a>
+<img src="
+</img>
```
pxeger, line 23192: `` autocloses, so `</img>` isn't necessary

pxeger: In fact, it's technically invalid HTML. https://stackoverflow.com/a/1946452

pxeger (approved):

PyGamer0 (approved):
never gonna give you up

PR #487, opened 2021-12-23
Rename step in testing workflow

pxeger:
It was called `build` which was kinda confusing; now it's called `test`


Lyxal (approved):
Lgtm

chunkybanana (approved):

PR #488, opened 2021-12-23
Make the find element and all slices element better

### LINKED ISSUE #453: Every slice of a string/list digraph
`(a: num, b: str | lst) => [b[0::a],b[1::a]...b[a-1::a]]` or something like that - get all slices of a list, skipping a certain number of items

### LINKED ISSUE #485: Make find rearrange its arguments if one argument is a list and the other is a scalar
because finding a list in a scalar will always return 1

(created by lyxal [here](https://chat.stackexchange.com/transcript/message/59961986))

### LINKED ISSUE #486: Find doesn't seem to work with infinite lazy lists
e.g. Finding a factorial in the infinite factorial list

(created by lyxal [here](https://chat.stackexchange.com/transcript/message/59961997))

Lyxal:
Closes #453 
Closes #485
Closes #486


vyxal/elements.py
```diff
@@ -1021,14 +1018,13 @@ def fibonaacis(_, ctx):
...
 
-    @lazylist
     def gen():
         i = 0
         while True:
             yield sympy.fibonacci(i + 1)
             i += 1
 
-    return gen()
+    return LazyList(gen(), isinf=False)
```
cgccuser, line 1027: Why specifically say `isinf=False` when that's the default?

cgccuser (dismissed):

vyxal/elements.py
Lyxal: Typo

Lyxal: Should be true

pxeger (approved):

cgccuser (approved):

PR #489, opened 2021-12-23
Make empty modifiers not error

Lyxal:


cgccuser (changes requested):
Looks good to me, but we should also add testcases for empty modifiers.

pxeger (approved):

PyGamer0 (approved):

PR #491, opened 2021-12-23
Add direction parsing elements `¨^` and `¨□`

### LINKED ISSUE #452: direction parsing
A digraph to convert strings of `>v<^` to [0,1,2,3]

pxeger:
Closes #452

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -2585,6 +2585,38 @@ def parity(lhs, ctx):
     }.get(ts, lambda: vectorise(parity, lhs, ctx=ctx))()
 
 
+def parse_direction_arrow_to_integer(lhs, ctx):
+    """Element ¨^
+    (str) -> map characters in `>^<v` to integers
+    """
+    ts = vy_type(lhs)
+    if ts is str and len(lhs) == 1:
+        return {
+            ">": 0,
+            "^": 1,
+            "<": 2,
+            "v": 3,
+        }.get(lhs, -1)
+    else:
+        return vectorise(parse_direction_arrow_to_integer, lhs, ctx=ctx)()
+
+
+def parse_direction_arrow_to_vector(lhs, ctx):
+    """Element ¨^
+    (str) -> map characters in `>^<v` to direction vectors
+    """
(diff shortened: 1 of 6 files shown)


Lyxal (approved):
Looks good to me, but why the explicit `+1`s in the vectors?

pxeger:
Just to make it line up nicely with the `-1`s so it's easier to read, and make it, as you say, more explicit.

vyxal/elements.py
```diff
@@ -2585,6 +2585,38 @@ def parity(lhs, ctx):
     }.get(ts, lambda: vectorise(parity, lhs, ctx=ctx))()
 
 
+def parse_direction_arrow_to_integer(lhs, ctx):
+    """Element ¨^
+    (str) -> map characters in `>^<v` to integers
+    """
+    ts = vy_type(lhs)
+    if ts is str and len(lhs) == 1:
+        return {
```
cgccuser, line 2594: Why not manually vectorise for strings here instead? I feel like the intent may be clearer that way.

vyxal/elements.py
```diff
@@ -2585,6 +2585,38 @@ def parity(lhs, ctx):
...
+    else:
+        return vectorise(parse_direction_arrow_to_integer, lhs, ctx=ctx)()
+
+
+def parse_direction_arrow_to_vector(lhs, ctx):
+    """Element ¨^
+    (str) -> map characters in `>^<v` to direction vectors
+    """
+    ts = vy_type(lhs)
+    if ts is str and len(lhs) == 1:
```
cgccuser, line 2609: Same here

cgccuser (dismissed):

vyxal/elements.py
pxeger: Why not manually vectorise for lists here as well? Because we already have a function for it.

cgccuser: Strings aren't lists, though, and I don't feel comfortable relying on vectorise treating strings as iterables.

cgccuser (approved):
\*grumble grumble\*

PR #492, opened 2021-12-24
Address #475 and #490

### LINKED ISSUE #475: Use `‟` for short dictionary access
We should be able to start an SCC with `‟` to indicate that the SCC is pointing to the short dictionary. Currently, the only way to access the short dictionary is by putting an ASCII character after an SCC character. However, that may not always be feasible. This would allow us to access the short dictionary alongside the main dictionary if need be, and it would still leave over 2200 empty slots in the main dictionary for any further expansion.

Example:
If you wanted to do `https://www.google.com/query?q=rick astley`, and you tried to use `∷` to give you the `https://www.google.com/query?q=`, you would have to put an ASCII character after it, such as a space, or do `rick astley` uncompressed. Instead, you could do `‟∷ṗṫ …↓⇧ṙ`, which may not save any bytes over putting an ASCII character after the `∷`, but it would give you the exact string you wanted.

### LINKED ISSUE #490: Range generation flags don't affect `℅`
Regardless of which flags you use, the `℅` command will cast numbers to the range `[0..n]`.

Example: [Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCIy4oSFICMgVHJ5IHRoZXNlIGZsYWdzOiBNIG0g4bmAIiwiIiwiIl0=)

Lyxal:
Closes #475
Closes #490

### DIFF
--- a/vyxal/helpers.py
+++ b/vyxal/helpers.py
@@ -759,6 +759,8 @@ def uncompress_dict(source: str) -> str:
                 if pos < len(vyxal.dictionary.small_dictionary):
                     ret += vyxal.dictionary.small_dictionary[pos]
                 temp_scc = ""
+                if char == " ":
+                    continue
             ret += char
 
     if temp_scc:
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -2782,9 +2782,8 @@ def random_choice(lhs, ctx):
     (lst) -> random element of a
     (num) -> Random integer from 0 to a
     """
-    if vy_type(lhs) == NUMBER_TYPE:
-        return random.randint(0, lhs)
-    return random.choice(iterable(lhs, ctx=ctx))
+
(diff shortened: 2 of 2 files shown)


vyxal/helpers.py
```diff
@@ -759,6 +759,8 @@ def uncompress_dict(source: str) -> str:
                 if pos < len(vyxal.dictionary.small_dictionary):
                     ret += vyxal.dictionary.small_dictionary[pos]
                 temp_scc = ""
+                if char == " ":
+                    continue
```
pxeger, line 763: unrelated but why does this function use a deque? Why not just iterate over it?

pxeger (approved):

PyGamer0 (approved):

vyxal/helpers.py
Lyxal: It's purely because deques are used in other places like the lexer and the parser.

PR #493, opened 2021-12-24
Get rid of numpy

### LINKED ISSUE #483: Eliminate dependency on numpy
numpy is large and we only use it a small number of places, which could fairly simply be replaced

cgccuser:
Closes #483

### DIFF
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -4349,7 +4349,7 @@
     lst: Diagonals of a matrix, starting with the main diagonal.
   vectorise: false
   tests:
-    - "[[[1,2,3],[4,5,6],[7,8,9]]] : [[1,5,9],[2,6],[3],[4,8],[7]]"
+    - "[[[1,2,3],[4,5,6],[7,8,9]]] : [[1,5,9],[2,6],[3],[7],[4,8]]"
 
 - element: "ÞS"
   name: Sublists
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -3862,7 +3862,7 @@ def vy_hex(lhs, ctx):
 def vy_int(item: Any, base: int = 10, ctx: Context = DEFAULT_CTX):
     """Converts the item to the given base. Lists are treated as if
     each item was a digit.
-    
+
     Used for multiple elements, and has to be here because it uses
     functions defined only here."""
     t_item = type(item)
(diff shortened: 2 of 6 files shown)


Lyxal (approved):

documents/knowledge/elements.yaml
```diff
@@ -4349,7 +4349,7 @@
     lst: Diagonals of a matrix, starting with the main diagonal.
   vectorise: false
   tests:
-    - "[[[1,2,3],[4,5,6],[7,8,9]]] : [[1,5,9],[2,6],[3],[4,8],[7]]"
```
chunkybanana, line 4352: Is this just a minor bugfix?

vyxal/elements.py
```diff
@@ -135,26 +134,15 @@ def all_diagonals(lhs, ctx):
...
-            yield vyxalify(current_diagonal)
-            diag_num -= 1
-            current_diagonal = numpy.diag(vector, k=diag_num)
-
-    return gen()
+    vector = [iterable(x, ctx=ctx) for x in lhs]
+    all_diags = [[] for _ in range(len(vector) * 2 - 1)]
+    start = 0
+    print(vector)
+    for row in vector:
```
chunkybanana, line 141: This code looks like it relies on the matrix to be square...

vyxal/elements.py
```diff
@@ -4058,11 +4049,7 @@ def vy_type(item, rhs=None, other=None, simple=False):
         )
     elif rhs is not None:
         return (vy_type(item, simple=simple), vy_type(rhs, simple=simple))
-    elif (
-        (x := type(item)) in (int, complex, float)
-        or is_sympy(item)
-        or isinstance(item, numpy.number)
-    ):
+    elif (x := type(item)) in (int, complex, float) or is_sympy(item):
```
chunkybanana, line 4052: `type` not `vy_type`?

vyxal/helpers.py
```diff
@@ -487,11 +486,14 @@ def pad_to_square(array: VyList) -> VyList:
     Returns an array padded to the square of the largest dimension.
     https://stackoverflow.com/a/11763827/9363594
```
chunkybanana, line 487: Probably get rid of this since it's not using that answer anymore

chunkybanana (commented):
Just in general make sure there's nowhere that still uses numpy

documents/knowledge/elements.yaml
cgccuser: Nah, it's just that my implementation returns them in a different order. We currently have no way for our tests to check things without the order, I should make an issue for that.

vyxal/elements.py
cgccuser: I think because `vy_type` returns Vyxal-specific types? @Lyxal wrote this, not me

vyxal/elements.py
cgccuser: It does, why?

chunkybanana: If that's the spec, never mind.

vyxal/elements.py
Lyxal: Y'all realise that this is _inside_ the `vy_type` function right?

chunkybanana (approved):

PR #522, opened 2021-12-28
Fix everything lgtm was complaining about

chunkybanana:


cgccuser:
Looks like you forgot to import math somewhere?

vyxal/helpers.py
```diff
@@ -8,7 +8,6 @@
 import collections
 import inspect
 import itertools
-import math
```
cgccuser, line 11: why?

cgccuser (changes requested):

Lyxal (changes requested):
The test cases need that math import. Leave it in helpers because I think there's a few other things that need it (mostly Sympy to python stuff)

pxeger:
I can't see anywhere in the tests themselves that use `math`; it seems to be just the sympy conversion things. Besides, it's bad practice to change your main code to make tests easier

Lyxal:
> I can't see anywhere in the tests themselves that use `math`; it seems to be just the sympy conversion things. Besides, it's bad practice to change your main code to make tests easier

That math import is also needed for the main files too.

vyxal/helpers.py
chunkybanana: Lgtm said it was an unnecessary import, and a quick ctrl-f in there found nothing but description comments.

chunkybanana:
> > I can't see anywhere in the tests themselves that use `math`; it seems to be just the sympy conversion things. Besides, it's bad practice to change your main code to make tests easier
> 
> That math import is also needed for the main files too - it's used for converting Sympy stuff to decimals using python precision

My point is, nothing in `helpers.py` directly uses `math`, so it doesn't need to be there.

Lyxal:
Except it can't.

```
exec(sympy.pycode(expr))
```

Will sometimes contain references to the math library. It does not contain an import statement however.

This is so that Floats can be displayed with things like that special flag without having an unruly amount of 0s at the end, and also means that complex numbers are converted to their python equivalent for display and for testing purposes.

Lyxal:
This is one of those cases where it's best to just slip one of those `ignore issue` comments into the source.

cgccuser:
Can we not `import math` right above `exec(sympy.pycode(expr))`?

cgccuser:
Or can we use `value.evalf()` instead of `eval(sympy.py_code(value))`?

chunkybanana:
> Except it can't.
> 
> ```
> exec(sympy.pycode(expr))
> ```
> 
> Will sometimes contain references to the math library. It does not contain an import statement however.
> 
> This is so that Floats can be displayed with things like that special flag without having an unruly amount of 0s at the end, and also means that complex numbers are converted to their python equivalent for display and for testing purposes.

Oh ok.

Lyxal:
> # lgtm [py/unused-import]

not without potentially having a lot of trailing 0s for numbers like `0.5`

Lyxal (approved):
very happy now

cgccuser (approved):
Oh well, I'm not happy about evaling strings by sympy but if it's the best option we have, let's go ahead with this.

PR #523, opened 2021-12-28
Add GitHub actions workflow to build and publish package to PyPI

pxeger:
You will soon be able to do `pip install vyxal` with no further details. Closes https://github.com/attempt-this-online/attempt-this-online/issues/6 (CC @Lyxal)

This needs to be pulled into `main`, tested with a placeholder tag (I suggest `v2.7.2post1`), and if/when it works:
1. remove the `repository_url` so it uses the real PyPI
2. replace the PyPI token with one that has permission to upload to the `vyxal` package namespace
  - I suggest I own that package initially and then once I've got it all working I'll grant permissions to people (you'll need PyPI accounts!)

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -1,0 +1,26 @@
+name: Publish to PyPI
+
+on:
+  push:
+    tags: ['*']
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


Lyxal (approved):
The proposed package ownership plan seems reasonable. Perhaps the test release tag could be 2.7.2.1

Lyxal:
Also it looks like I already have a pypi account if that helps you at all.

cgccuser (approved):

PR #526, opened 2021-12-28
Heed an LGTM alert

cgccuser:
- `vy_eval` wasn't being called with a context argument in one place
- lgtm was complaining about a non-callable being called so I shut it up


pxeger (approved):

pxeger:
Needs reformatting with Black

Lyxal (approved):
Format with black and this will go through

pxeger:
uh

Lyxal:
Wait why'd it let it go through without passing all checks?

pxeger:
I think the branch protection rules don't require all checks to pass in order to merge; they only require 2 reviews. I'll change that.

pxeger:
would...

image

did :)

image

PR #528, opened 2021-12-28
Make deploy yml file name match workflow name

cgccuser:
I'm trying to get [workflow status badges](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge) to work


cgccuser:
Wait no I just need to say `yaml` instead of `yml` I'm dumb

PR #533, opened 2021-12-28
Generate tests in the test workflow

cgccuser:
I just realized I made a mistake in elements.yml but it wasn't noticed because the tests weren't regenerated, which is not good.


pxeger:
I think we should add a separate CI check to ensure this file (and the others generated from `elements.yaml`) get regenerated? I think that could be done with https://pre-commit.ci pretty easily.

But this is a good step for now

pxeger (approved):

pxeger:
No squash please

cgccuser:
Why no squash?

pxeger:
~~Badminton's just better~~

Because it obscures commit history which carries useful information.

pxeger (changes requested):
This needs to install `yaml` first

Lyxal (approved):
Maybe also add something later for documentation rerolling

cgccuser:
Sounds good, although generating documentation should go in its own workflow.

.github/workflows/run-tests.yaml
```diff
@@ -28,6 +28,12 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install poetry
```
pxeger, line 29:
```suggestion
        python -m pip install poetry yaml
```

pxeger (changes requested):

cgccuser:
There are so many commits that at this point we might as well squash them or make a new branch and add those in manually.

pxeger (approved):

pxeger:
I still don't like squashing, but whatever

cgccuser:
I mean, we could make a new branch and add those commits in again.

cgccuser:
You know what, I'm too lazy to do that, so here we go. And fwiw, tennis > badminton.

PR #534, opened 2021-12-28
Remove a few badges

cgccuser:
This will keep only the lgtm code quality badge (because it looks nice if we have A quality code) and the tests badge (because it's important for our code to run). PA deployment can be seen on the right in the Environments section, I think. We're not gluten free anymore because of sympy, and we don't want people seeing how many LGTM alerts we have.


pxeger (approved):
This is much more sensible tbh

Lyxal (approved):
Remind me to add something from deepsource too

PR #546, opened 2021-12-30
Vectorised printing should no longer be an issue

### LINKED ISSUE #336: Vectorised printing in general
Some commands like `,` need to not lazily evaluate their arguments/discard the value when vectorised

(created by lyxal [here](https://chat.stackexchange.com/transcript/message/59787158))

Lyxal:
Closes #336 

Consider something like:

```
₁ƛ₍₃₅kF½*∑∴,
```

Usually, that prints an ugly mess of list syntax:

```
⟨ 1
12
 | 2Fizz
 | 34
 | 4Buzz
 | 5Fizz
 | 67
 | 78
 | 8Fizz...
```

This PR fixes that by making it so that functions can tell if there's a possibility they might print, and if so, force eager evaluation in situations where lazy evaluation wouldn't work.

### DIFF
--- a/vyxal/structure.py
+++ b/vyxal/structure.py
@@ -13,7 +13,7 @@ Branch = Union[str, list["Structure"], list["Token"]]
 class Structure:
     def __init__(self, *branches: Branch):
         # Don't do anything with the arguments
-        self.branches = branches
+        self.branches = list(branches)
 
     def transpile(self) -> str:
         """Return the transpiled version of the structure"""
@@ -91,6 +91,7 @@ class Lambda(Structure):
         super().__init__(str(arity), body)
         self.arity = arity
         self.body = body
+        self.force_eval = can_print(body)
 
 
 class LambdaMap(Lambda):
@@ -151,3 +152,30 @@ class TriadicModifier(Structure):
         return (
             f"{type(self).__name__}({repr(self.branches)}, {self.modifier!r})"
         )
+
+
+def can_print(structures: list[Structure]) -> bool:
+    """Checks if the structures are printing instructions"""
+    for struct in structures:
+        if isinstance(
+            struct,
+            (
+                BreakStatement,
+                RecurseStatement,
+                FunctionCall,
+            ),
+        ):
+            continue  # These statements can't possibly print, so it's
+            # okay to just continue.
+        elif isinstance(struct, GenericStatement):
+            if struct.branches[0][0].value in list(",₴…") + ["¨,", "¨…"]:
+                return True
+        else:
+            for branch in struct.branches:
+                if isinstance(branch, (Structure)):
+                    branch = [branch]
+                if isinstance(branch, (list, tuple)):
+                    if can_print(branch):
+                        return True
+
+    return False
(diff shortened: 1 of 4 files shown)


nailuj29gaming (approved):

vyxal/structure.py
```diff
@@ -151,3 +152,34 @@ def __repr__(self):
...
+            (
+                BreakStatement,
+                RecurseStatement,
+                FunctionCall,
+            ),
+        ):
+            continue  # These statements can't possibly print, so it's
+            # okay to just continue.
+        elif isinstance(struct, GenericStatement):
+            if struct.branches[0][0].value in list(",₴…") + ["¨,", "¨…"]:
```
chunkybanana, line 171: Hardcoding this... could cause problems.

PyGamer0: i agree

Lyxal: well where else is it supposed to store the elements that print?

pxeger:
Honestly I don't see this as a good idea. I think it will just create more bugs and inconsistencies, e.g. if I add a print instruction when debugging, then my loop's behavior completely changes making it harder to debug. There definitely at least needs to be a way to override it and disable forced evaluation.

Lyxal:
@pxeger the idea of this PR is to make it so that having `,` or other printing commands in things like map and filter don't produce things like the example in the PR description.

Having a `,` at the end of a map is a feasible way of joining a map on newlines if you don't want to use flags and you don't want to have to suffer the pain of `;⁋`

pxeger:
I know what it's supposed to do, but I maintain my opinion that the current behaviour is not a bug nor does anything about it really need changing. (Maybe an element to strictly evaluate a list and then do nothing with the result, but no drastic behaviour modifications).

Lyxal:
Compared to 2.4, the current feature could be considered a bug. 

Compare:

[Try it Online!](https://lyxal.pythonanywhere.com?flags=&code=10%20%C6%9B3%2B%2C&inputs=&header=&footer=) (2.4)

to

[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCIxMCDGmzMrLCIsIiIsIiJd) (2.7.2)

vyxal/structure.py
cgccuser: You could extract it into a toplevel variable called `ELEMENTS_THAT_PRINT` or something so it's easier to hunt for.

TheSecondComing123 (dismissed):
I mostly approve of this, but I personally think that "force_eval" should be in a constant (for example `FORCE_EVAL = "force_eval"`) in case the name is changed.

TheSecondComing123:
I also think that this:

```python
if FORCE_EVAL in dir(function) and function.force_eval:
            return list(ret)
        else:
            return LazyList(ret)
```

should be in a function to avoid repeated code.

Lyxal:
Also, this can be generalised to cover things like global variable usage and global array access.

Lyxal:
After consideration and asking in the nineteenth byte, I've decided that this isn't actually a bug, and hence the PR will be closed until further notice.

cgccuser:
Sorry this PR didn't work out :(. Let's keep the branch around anyway since a lot of work seems to have gone into it.

PR #552, opened 2022-01-01
It's the new year, so the license needs updating.

Lyxal:


Tams-Tams (approved):

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Vyxal Organisation
+Copyright (c) 2022 Vyxal Organisation
```
cgccuser, line 3: Wait I thought it was `(c) 2021-2022`?

Lyxal: I'm pretty sure it should just be the current year

PR #553, opened 2022-01-01
Add installation instructions

cgccuser:
Add installation (and running) instructions to README.

Credit to mathcat for [suggesting this in chat](https://chat.stackexchange.com/transcript/message/60030145#60030145).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,28 @@ Vyxal is also deliberately designed to be an easy language to learn. While it ma
 
 Ultimately, Vyxal is a language for golfers, by golfers.
 
+## Installation
+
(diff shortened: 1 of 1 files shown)


Tams-Tams (changes requested):
Looks good, idk, maybe it's worth adding https://pypi.org/project/vyxal/ below too.

cgccuser:
@Tams-Tams I don't think people really need to see the package page itself, it's not necessary for installation. All the information present on the PyPI page is in our repository. I don't think I've seen other projects do that? Correct me if I'm wrong though

Tams-Tams (approved):
Oh, then ok.

Lyxal (approved):

PR #558, opened 2022-01-03
Make `o` do the same as `ṅ` but starting at `1` instead of `0`

### LINKED ISSUE #511: Way to get first n *positive* integers

Lyxal:
Closes #511

### DIFF
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -1148,9 +1148,9 @@
   description: Remove instances of b in a
   arity: 2
   overloads:
-    any-any: a.replace(b,"")
     num-fun: first a positive integers where b is truthy
     fun-num: first b positive integers where a is truthy
+    any-any: a.replace(b,"")
   vectorise: false
   tests:
     - '["hello","l"] : "heo"'
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -1149,6 +1149,8 @@
   arity: 2
   overloads:
     any-any: a.replace(b,"")
+    num-fun: first a positive integers where b is truthy
+    fun-num: first b positive integers where a is truthy
   vectorise: false
   tests:
     - '["hello","l"] : "heo"'
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -2855,8 +2855,9 @@ def regex_sub(lhs, rhs, other, ctx):
 
 def remove(lhs, rhs, ctx):
     """Element o
-    (any, any) -> a.remove(b)
+    (num, fun) -> first a positive integers where b is truthy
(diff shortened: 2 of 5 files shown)


PyGamer0 (approved):

Tams-Tams (approved):

documents/knowledge/elements.yaml
```diff
@@ -1149,6 +1149,8 @@
   arity: 2
   overloads:
     any-any: a.replace(b,"")
+    num-fun: first a positive integers where b is truthy
+    fun-num: first b positive integers where a is truthy
```
cgccuser, line 1153:
```suggestion
    num-fun: first a positive integers where b is truthy
    fun-num: first b positive integers where a is truthy
    any-any: a.replace(b,"")
```
It just seems like the any-any overloads is matched before the num-fun/fun-num overloads right now.

vyxal/elements.py
```diff
@@ -2854,9 +2856,24 @@ def regex_sub(lhs, rhs, other, ctx):
 def remove(lhs, rhs, ctx):
     """Element o
     (any, any) -> a.remove(b)
+    (fun, num) -> first b positive integers where a is truthy
```
cgccuser, line 2859: This doesn't have the (num, fun) overload. Also, could the (any, any) overload be moved to the bottom here too?

vyxal/elements.py
```diff
@@ -2854,9 +2856,24 @@ def regex_sub(lhs, rhs, other, ctx):
...
     """
     lhs = iterable(lhs)
     ts = vy_type(lhs)
+    if set(vy_type(lhs, rhs)) == {types.FunctionType, NUMBER_TYPE}:
+        lhs, rhs = (rhs, lhs) if ts is types.FunctionType else (lhs, rhs)
```
cgccuser, line 2864: We really need a better way to check types and swap than this :\. Not a big deal though, you can resolve this comment and we can work on it later

cgccuser (changes requested):
The code itself is good, just a few minor suggestions

documents/knowledge/elements.yaml
Tams-Tams: What do you think of a file containing the order of overloads?

cgccuser: `elements.yaml` has that already

vyxal/elements.py
cgccuser, line 2859:
```suggestion
    (num, fun) -> first a positive integers where b is truthy
    (fun, num) -> first b positive integers where a is truthy
    (any, any) -> a.remove(b)
```

cgccuser (changes requested):

cgccuser (approved):

PR #559, opened 2022-01-03
Some fixes from deepsource

cgccuser:
Fixed a few things to get deepsource to shut up, added some docstrings, made flake8 ignore test_elements.py

### DIFF
--- a/flask_app.py
+++ b/flask_app.py
@@ -60,53 +60,50 @@ def execute():
     with open(f"sessions/{session}/.stdin", "w", encoding="utf-8") as f:
         f.write(input_list)
 
-    with open(f"sessions/{session}/.stdin", "r", encoding="utf-8") as x:
-        with open(f"sessions/{session}/.stdout", "w", encoding="utf-8") as y:
-            with open(
-                f"sessions/{session}/.stderr", "w", encoding="utf-8"
-            ) as z:
-                manager = multiprocessing.Manager()
-                ret = manager.dict()
+    with open(f"sessions/{session}/.stdin", "r", encoding="utf-8") as x, open(
+        f"sessions/{session}/.stdout", "w", encoding="utf-8"
+    ) as y, open(f"sessions/{session}/.stderr", "w", encoding="utf-8") as z:
+        manager = multiprocessing.Manager()
+        ret = manager.dict()
 
-                if "5" in flags:
-                    time = 5
-                elif "T" in flags:
-                    time = 60
-                elif "b" in flags:
-                    time = 15
-                elif "B" in flags:
-                    time = 30
-                else:
-                    time = 10
-                ret[1] = ""
-                ret[2] = ""
-                fcode = (
-                    (header and (header + "\n"))
-                    + code
-                    + (footer and ("\n" + footer))
-                )
+        if "5" in flags:
+            time = 5
+        elif "T" in flags:
+            time = 60
+        elif "b" in flags:
+            time = 15
+        elif "B" in flags:
+            time = 30
+        else:
+            time = 10
+        ret[1] = ""
+        ret[2] = ""
+        fcode = (
+            (header and (header + "\n")) + code + (footer and ("\n" + footer))
+        )
 
-                sessions[session] = multiprocessing.Process(
-                    target=execute_vyxal,
-                    args=(fcode, flags, input_list, ret, True),
-                )
-                sessions[session].start()
-                sessions[session].join(time)
+        sessions[session] = multiprocessing.Process(
+            target=execute_vyxal,
+            args=(fcode, flags, input_list, ret, True),
+        )
+        sessions[session].start()
+        sessions[session].join(time)
 
-                if session in terminated:
-                    terminated.remove(session)
-                    ret[2] += "\nSession terminated upon user request"
+        if session in terminated:
+            terminated.remove(session)
+            ret[2] += "\nSession terminated upon user request"
 
-                if sessions[session].is_alive():
+        if sessions[session].is_alive():
 
-                    sessions[session].kill()
-                    if 2 in ret:
-                        ret[2] += "\n" + f"Code timed out after {time} seconds"
-                y.write(ret[1])
-                z.write(ret[2])
-    with open(f"sessions/{session}/.stdout", "r", encoding="utf-8") as x:
-        with open(f"sessions/{session}/.stderr", "r", encoding="utf-8") as y:
-            val = {"stdout": x.read(), "stderr": y.read()}
+            sessions[session].kill()
+            if 2 in ret:
+                ret[2] += "\n" + f"Code timed out after {time} seconds"
+        y.write(ret[1])
+        z.write(ret[2])
+    with open(f"sessions/{session}/.stdout", "r", encoding="utf-8") as x, open(
+        f"sessions/{session}/.stderr", "r", encoding="utf-8"
+    ) as y:
+        val = {"stdout": x.read(), "stderr": y.read()}
     shutil.rmtree(f"sessions/{session}", ignore_errors=True)
     return val
 
--- a/flask_app.py
+++ b/flask_app.py
@@ -60,9 +60,11 @@ def execute():
     with open(f"sessions/{session}/.stdin", "w", encoding="utf-8") as f:
         f.write(input_list)
 
-    with open(f"sessions/{session}/.stdin", "r", encoding="utf-8") as x, open(
-        f"sessions/{session}/.stdout", "w", encoding="utf-8"
-    ) as y, open(f"sessions/{session}/.stderr", "w", encoding="utf-8") as z:
+    with (
+        open(f"sessions/{session}/.stdin", "r", encoding="utf-8") as x,
+        open(f"sessions/{session}/.stdout", "w", encoding="utf-8") as y,
+        open(f"sessions/{session}/.stderr", "w", encoding="utf-8") as z,
+    ):
         manager = multiprocessing.Manager()
         ret = manager.dict()
 
--- a/vyxal/LazyList.py
+++ b/vyxal/LazyList.py
@@ -35,9 +35,12 @@ class LazyList:
                 last = 0
 
             while last <= lhs:
-                last = next(self)
-                if last == lhs:
-                    return 1
+                try:
+                    last = next(self)
+                    if last == lhs:
+                        return 1
+                except StopIteration:
+                    break
             return 0
         else:
             for temp in self:
--- a/vyxal/LazyList.py
+++ b/vyxal/LazyList.py
@@ -131,19 +131,19 @@ class LazyList:
         self.generated[position] = value
 
     def count(self, other):
+        """Number of occurrences of `other` in this `LazyList`"""
         temp = self.listify()
         return temp.count(other)
 
+    @lazylist
     def filter(self, fn):
-        @lazylist
-        def gen():
-            for item in self:
-                if fn(item):
-                    yield item
-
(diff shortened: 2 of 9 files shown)


vyxal/elements.py
```diff
@@ -4268,9 +4281,9 @@ def zero_matrix(lhs, ctx):
     """
     mat = []
     temp = 0
-    for index in iterable(lhs, ctx=ctx):
+    for ind in iterable(lhs, ctx=ctx):
```
cgccuser, line 4284: Deepsource was complaining about redefining `index` here so I renamed it to `ind`

vyxal/helpers.py
```diff
@@ -96,19 +96,24 @@ def digits(num: NUMBER_TYPE) -> List[int]:
 
 
 @lazylist
-def enumerate_md(haystack: VyList, index_stack: List[int] = []) -> VyList:
```
cgccuser, line 99: Extracted this into a helper because Deepsource was complaining that `[]` as a default argument was dangerous because it's mutable. We're not mutating it so we could've just left that in, but it's one less thing Deepsource will complain about and I don't want to allow ourselves to accidentally set `index_stack` to something other than `[]` when calling `enumerate_md`.

flask_app.py
```diff
@@ -60,53 +60,50 @@ def execute():
     with open(f"sessions/{session}/.stdin", "w", encoding="utf-8") as f:
         f.write(input_list)
 
-    with open(f"sessions/{session}/.stdin", "r", encoding="utf-8") as x:
```
cgccuser, line 63: Sorry for the ginormous diff here, but Deepsource said to merge the `with`s into a single one. Let me know if it looks prettier the old way

Lyxal (approved):
Lgtm

vyxal/elements.py
Lyxal: that's understandable - name masking is a bad thing

flask_app.py
```diff
@@ -60,53 +60,50 @@ def execute():
...
-                    if 2 in ret:
-                        ret[2] += "\n" + f"Code timed out after {time} seconds"
-                y.write(ret[1])
-                z.write(ret[2])
-    with open(f"sessions/{session}/.stdout", "r", encoding="utf-8") as x:
-        with open(f"sessions/{session}/.stderr", "r", encoding="utf-8") as y:
-            val = {"stdout": x.read(), "stderr": y.read()}
+    with open(f"sessions/{session}/.stdin", "r", encoding="utf-8") as x, open(
+        f"sessions/{session}/.stdout", "w", encoding="utf-8"
+    ) as y, open(f"sessions/{session}/.stderr", "w", encoding="utf-8") as z:
```
pxeger, line 65:
```suggestion
    with (
        open(f"sessions/{session}/.stdin", "r", encoding="utf-8") as x,
        open(f"sessions/{session}/.stdout", "w", encoding="utf-8") as y,
        open(f"sessions/{session}/.stderr", "w", encoding="utf-8") as z,
    ):
```

pxeger: Can we also take this opportunity to change the terrible `x` `y` `z` variable names?

vyxal/LazyList.py
```diff
@@ -149,6 +152,7 @@ def listify(self):
         return temp
 
     def output(self, end="\n", ctx=None):
+        """Print this LazyList"""
```
pxeger, line 155:
```suggestion
```

vyxal/LazyList.py
```diff
@@ -172,11 +176,9 @@ def output(self, end="\n", ctx=None):
...
+    @lazylist
     def reversed(self):
-        @lazylist
-        def temp():
-            self.generated += list(itertools.tee(self.raw_object)[-1])
-            for item in self.generated[::-1]:
-                yield item
-
-        return temp()
+        """Reversed version of this LazyList"""
```
pxeger, line 181:
```suggestion
```

vyxal/LazyList.py
```diff
@@ -128,19 +131,19 @@ def __setitem__(self, position, value):
...
-                    yield item
-
-        return gen()
+        """A `LazyList` containing only elements for whom `fn` is true"""
+        for item in self:
+            if fn(item):
+                yield item
 
     def listify(self):
+        """Turn into a list"""
```
pxeger, line 146: Use docstrings only where necessary to explain what something does. In this case it's pretty obvious.
```suggestion
```

vyxal/elements.py
```diff
@@ -587,18 +585,19 @@ def coords_deepmap(lhs, rhs, ctx):
...
-    ]
+    def f(a, g, pos):
+        return [
+            f(b, g, (*pos, i))
+            if isinstance(b, list)
+            else safe_apply(g, [*pos, i], ctx=ctx)
+            for i, b in enumerate(a)
+        ]
 
     # the above curtosey of pxeger
```
pxeger, line 596:
```suggestion
    # the above courtesy of pxeger
```

vyxal/elements.py
```diff
@@ -587,18 +585,19 @@ def coords_deepmap(lhs, rhs, ctx):
 
     lhs = iterable(lhs, ctx=ctx)  # Make sure lhs is actually iterable
 
-    f = lambda a, g, pos=(): [
-        f(b, g, (*pos, i))
-        if isinstance(b, list)
-        else safe_apply(g, [*pos, i], ctx=ctx)
-        for i, b in enumerate(a)
-    ]
+    def f(a, g, pos):
```
pxeger, line 588:
```suggestion
    def f(a, g, pos=()):
```

vyxal/elements.py
```diff
@@ -587,18 +585,19 @@ def coords_deepmap(lhs, rhs, ctx):
...
+            else safe_apply(g, [*pos, i], ctx=ctx)
+            for i, b in enumerate(a)
+        ]
 
     # the above curtosey of pxeger
     # https://chat.stackexchange.com/transcript/message/59662694#59662694
     # thank you very cool
 
-    return f(lhs, rhs)
+    return f(lhs, rhs, ())
```
pxeger, line 600:
```suggestion
    return f(lhs, rhs)
```

vyxal/elements.py
```diff
@@ -2979,10 +2979,12 @@ def replace_until_no_change(lhs, rhs, other, ctx):
 
 
 def request(lhs, ctx):
+    """Element ¨U
+    (str) -> Send a GET request to a URL if online"""
     x = urllib.request.urlopen(urlify(lhs)).read()
     try:
         return x.decode("utf-8")
-    except Exception as e:
+    except Exception:
```
pxeger, line 2987:
```suggestion
    except UnicodeDecodeError:
```

vyxal/helpers.py
pxeger: Why not just use a tuple?

pxeger (dismissed):

flask_app.py
cgccuser: Yes please

vyxal/LazyList.py
cgccuser: Deepsource was complaining about missing docstrings, we could add a comment to ignore that in the future

vyxal/elements.py
cgccuser: It's funiner this way

vyxal/helpers.py
cgccuser: I didn't like the idea of exposing implementation details by allowing callers to set `index_stack` themselves, but if an optional argument is more idiomatic, we can go with that, sure

pxeger: I think the idiomatic way to specify that an argument is an implementation detail and shouldn't be used by API consumers, is just to name the argument with a leading underscore:

` def enumerate_md(haystack: VyList, _index_stack: List[int] = ()) -> VyList:`

flask_app.py
chunkybanana: This way's nicer :P

vyxal/elements.py
```diff
@@ -587,12 +585,13 @@ def coords_deepmap(lhs, rhs, ctx):
...
-    ]
+    def f(a, g, pos=()):
+        return [
+            f(b, g, (*pos, i))
+            if isinstance(b, list)
+            else safe_apply(g, [*pos, i], ctx=ctx)
+            for i, b in enumerate(a)
+        ]
 
     # the above curtosey of pxeger
```
chunkybanana, line 596: curtosey

vyxal/elements.py
```diff
@@ -3052,17 +3054,17 @@ def roman_numeral(lhs, ctx):
             raise ValueError("Number must be between 1 and 3999")
```
chunkybanana, line 3054: This was made by copilot under my supervision and I've realised it should actually say what the problem is

vyxal/structure.py
```diff
@@ -15,20 +15,13 @@ def __init__(self, *branches: Branch):
...
-
     def __repr__(self):
         return f"{type(self).__name__}({repr(self.branches)})"
 
 
 class GenericStatement(Structure):
     """Generic statements are elements and so on"""
 
-    def __init__(self, *branches: Branch):
-        super().__init__(*branches)
```
chunkybanana, line 30: Did we need these?

vyxal/transpile.py
```diff
@@ -36,6 +36,12 @@ def lambda_wrap(
 
 
 def transpile(program: str, dict_compress: bool = True) -> str:
+    """Transpile an entire Vyxal program
+    Parameters:
+    program: str
```
chunkybanana, line 41: Uh, you can kinda tell what parameters it takes by *reading the function header*

chunkybanana: This much is kinda unnecessary

chunkybanana (approved):
lgtm

vyxal/structure.py
cgccuser: No, `super().__init__` does the same thing, no need to overload. The other structures, however, do need overloads because they do special things with `branches`.

vyxal/transpile.py
cgccuser: Sure, I'll get rid of this too

vyxal/transpile.py
```diff
@@ -36,6 +36,12 @@ def lambda_wrap(
...
+      The Vyxal program to transpile
+    dict_compress: bool = True
+      Whether or not to use dictionary compression"""
```
cgccuser, line 44:
```suggestion
```

vyxal/elements.py
```diff
@@ -4263,8 +4277,7 @@ def f():
                     exhausted += 1
                 if exhausted == 2:
                     break
-                else:
-                    yield [left_item, right_item]
+                yield [left_item, right_item]
```
pxeger, line 4280: This is definitely clearer as is.

```suggestion
                else:
                    yield [left_item, right_item]
```

BTW, I'm pretty sure I've previously rejected a change like this; possibly even this exact bit of code. Deepsource is just too aggressive with these complaints for my liking.

pxeger (changes requested):

pxeger:
Just one thing, then I'll approve

vyxal/elements.py
Tams-Tams: I always feel return and yield without else make python code pythonicaler.

cgccuser: Alright, I'll revert it. We should also add a comment to make Deepsource ignore it in the future.

Tams-Tams (approved):

PR #563, opened 2022-01-04
implement both suggestions from issue #548

### LINKED ISSUE #548: Digraph for `Ṫ0p`
Remove the last, prepend a zero.

This is very useful in anything involving prefixes. It'd also be nice to have one for `¦Ṫ0p` - ^ but with cumulative sums.

Lyxal:
closes #548

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -3113,7 +3113,7 @@ def sans_last_prepend_zero(lhs, ctx):
     return {
         NUMBER_TYPE: lambda: tail_remove(lhs, ctx),  # prepending a 0 to
         # a number just makes it the same number
-        str: lambda: "0" + lhs[:-1],
+        str: lambda: "0" + lhs[:-1],  # leave as string
     }.get(ts, lambda: prepend(tail_remove(lhs, ctx), 0, ctx=ctx))()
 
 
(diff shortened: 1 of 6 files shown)


vyxal/elements.py
```diff
@@ -3097,6 +3105,18 @@ def run_length_decoding(lhs, ctx):
         return LazyList(temp)
 
 
+def sans_last_prepend_zero(lhs, ctx):
```
cgccuser, line 3108: Could we rename this `drop_last_prepend_zero` instead? `drop` feels more intuitive than `sans`, even if `sans` is fancier.

vyxal/elements.py
```diff
@@ -3097,6 +3105,18 @@ def run_length_decoding(lhs, ctx):
...
+    """Element Þr
+    Remove the last item of a list and prepend 0
+    """
+    ts = vy_type(lhs)
+    return {
+        NUMBER_TYPE: lambda: tail_remove(lhs, ctx),  # prepending a 0 to
```
cgccuser, line 3114: Hmm, maybe have a more useful overload for numbers then?

cgccuser (approved):
LGTM overall

pxeger (approved):

pxeger:
I feel like this would be better to implement with a single-byte overload which prepends zero

PyGamer0 (approved):

Lyxal:
> I feel like this would be better to implement with a single-byte overload which prepends zero

Then you have to have an element that has number and string overloads that can get away with not being vectorised by default.

vyxal/elements.py
chunkybanana: ^

PR #570, opened 2022-01-07
A bunch of ctx calls were missing to vy_str, which needs them

chunkybanana:
`vy_str` uses the `ctx.vyxal_lists` setting to determine how to stringify lists. When a ctx parameter isn't given, this throws an error.


Lyxal (approved):

pxeger (dismissed):
Please delete the `.DS_Store` files, and then ideally delete your computer because it uses macOS and macOS is bad

pxeger:
Why hasn't this been a problem before? Why doesn't `vy_str` require the `ctx` argument? Can you add some tests?

chunkybanana:
Turns out some of these currently don't cause problems because they're only stringifying scalars

chunkybanana:
# [Vyxal](https://github.com/Vyxal/Vyxal), 15 bytes
```
`abc` ⟨1|2|3⟩ Ẇ
```
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCJgYWJjYCDin6gxfDJ8M+KfqSDhuoYiLCIiLCIiXQ==)
# [Vyxal](https://github.com/Vyxal/Vyxal), 43 bytes
```
`hello` ⟨`a`|`l`|`[^h+]`⟩ ⟨\e|\t|`blah`⟩ øṙ
```
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCJgaGVsbG9gIOKfqGBhYHxgbGB8YFteaCtdYOKfqSDin6hcXGV8XFx0fGBibGFoYOKfqSDDuOG5mSIsIiIsIiJd)

# [Vyxal](https://github.com/Vyxal/Vyxal), 17 bytes
```
⟨4|5|6⟩ ⟨1|2|3⟩ j
```
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLin6g0fDV8NuKfqSDin6gxfDJ8M+KfqSBqIiwiIiwiIl0=)

# [Vyxal](https://github.com/Vyxal/Vyxal), 29 bytes
```
⟨⟨1|2|3⟩|⟨4|5|6⟩|⟨7|8|9⟩⟩ ¶ j
```
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLin6jin6gxfDJ8M+KfqXzin6g0fDV8NuKfqXzin6g3fDh8OeKfqeKfqSDCtiBqIiwiIiwiIl0=)

Yes, these don't make much sense as they are, but they should do something useful rather than randomly erroring.

The third one could probably be `⟨4|1|2|3|5|1|2|3|6⟩` or something. The second should probably behave like transliterate, but with regex. The first and fourth, I have no idea.

I'll make those into separate enhancement issues, this is a sort of patch.

cgccuser (approved):

PR #574, opened 2022-01-09
Version 2.7.4

Lyxal:


Not-Lyxal (approved):

pxeger (commented):
It doesn't really matter here, because this is a pretty uncontroversial pull request, but in general, please don't just force-accept PRs using alts. I'm not saying you might do this, I just don't want it to become a pattern.

Lyxal:
Of course.

I would have just disabled branch protection and force-pushed anyway - approving with alts was just quicker because I already had the accounts signed in.

PR #576, opened 2022-01-13
Handle desugaring of lambdas in transpiler instead of parser

cgccuser:
Instead of immediately splitting map, filter, and sort lambdas into two elements (the lambda and the operation after it), this adds a field `after` to the `Lambda` class itself that will contain an element to be run after the lambda (or be `None` if it's a normal lambda). Then, in the transpilation phase, a call to the `after` element is added for those three special lambdas. This allows stuff like `⁽ƛ....;` to work, because previously, the lambda and the call to `M` were held as separate elements.


Lyxal (approved):
Lgtm

razetime (approved):

pxeger (approved):

PR #584, opened 2022-01-17
Separate map/filter/sort lambdas from normal lambdas

cgccuser:
My "fix" for map lambdas not being taken as single elements earlier actually broke this. Now I've made a `LambdaOp` class that's separate from the `Lambda` class. Let's hope it works properly now.

Also, mypy found over a hundred errors :(.

### DIFF
--- a/tests/test_complex.py
+++ b/tests/test_complex.py
@@ -75,7 +75,8 @@ def test_map_lambda_as_element():
     stack = run_vyxal("⁽ƛ1+;M", inputs=[[[1, 2], [3, 4]]])
     assert stack[-1] == [[2, 3], [4, 5]]
 
+
 def test_vectorise_map_lambda():
     """Test that a map lambda can be vectorised"""
-    stack = run_vyxal("vƛ30∴;", inputs=[[[34,1324,23],[45,3]]])
-    assert simplify(stack[-1]) == [[34,1324,30],[45,30]]
+    stack = run_vyxal("vƛ30∴;", inputs=[[[34, 1324, 23], [45, 3]]])
+    assert simplify(stack[-1]) == [[34, 1324, 30], [45, 30]]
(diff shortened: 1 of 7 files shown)


tests/test_complex.py
```diff
@@ -72,3 +74,9 @@ def test_map_lambda_as_element():
     """Test that a map lambda is held as a single element"""
     stack = run_vyxal("⁽ƛ1+;M", inputs=[[[1, 2], [3, 4]]])
     assert stack[-1] == [[2, 3], [4, 5]]
+
+
+def test_vectorise_map_lambda():
```
chunkybanana, line 79: Maybe add tests for filter/sort as well?

vyxal/elements.py
```diff
@@ -47,7 +47,7 @@ def process_element(
     if arity:
         arguments = ["third", "rhs", "lhs"][-arity:]
     else:
-        arguments = "_"
+        arguments = ["_"]
```
chunkybanana, line 50: What does this do?

vyxal/helpers.py
```diff
@@ -56,7 +56,7 @@ def collect_until_false(
         val = safe_apply(function, val, ctx=ctx)
 
 
-def concat(vec1: VyList, vec2: VyList, ctx: Context = None) -> VyList:
```
chunkybanana, line 59: Any particular reason for this?

chunkybanana (approved):

vyxal/elements.py
cgccuser: Makes mypy happy

vyxal/helpers.py
cgccuser: Making mypy happy

tests/test_complex.py
cgccuser: Good idea but i is 2 lzy

Lyxal (approved):
lgtm

PR #628, opened 2022-01-19
Don't clone the raw object in LazyList

cgccuser:
There's probably still some wonkiness left with `LazyList`, but not making copies of the underlying generator and stuff should help. This should fix overdot-H (tail/remove first). It should also allow you to do `bool(somelazyilst)` without worrying about whether or not it's infinite, because there's now a separate `__bool__` method.

### DIFF
--- a/vyxal/LazyList.py
+++ b/vyxal/LazyList.py
@@ -20,9 +20,12 @@ def lazylist(fn):
 
 class LazyList:
     def __add__(self, rhs):
-        from vyxal.helpers import join_with
+        @lazylist
+        def gen():
+            yield from self
+            yield from rhs
 
-        return LazyList(join_with(self.raw_object, rhs))
+        return gen()
 
     def __call__(self, *args, **kwargs):
         return self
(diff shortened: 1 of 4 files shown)


vyxal/LazyList.py
```diff
@@ -142,11 +177,12 @@ def filter(self, fn):
...
     def listify(self):
-        from vyxal.helpers import vyxalify
-
-        temp = self.generated + vyxalify(list(self.raw_object))
-        self.raw_object = iter(temp[::])
-        self.generated = []
+        temp = self.generated[::]
+        while True:
+            try:
+                temp.append(self.__next__())
```
Lyxal, line 183: don't forget to call `vyxalify` here - some generators might return tuples.

Lyxal (changes requested):

vyxal/LazyList.py
cgccuser:
```suggestion
                temp.append(vyxalify(self.__next__()))
```

cgccuser: Wait __next__ itself should call vyxalify

cgccuser: Yeah next does call vyxalify, it's all god

Lyxal:
wait it does?

wow

I really am out of touch with my own codebase.

cgccuser:
tmw your past slef is smorter than yuo

Lyxal (approved):
lgtm

vyxal/LazyList.py
```diff
@@ -64,8 +79,22 @@ def __getitem__(self, position):
...
+                        except StopIteration:
+                            return
+                    i = 0
+                    while True:
+                        try:
+                            item = next(copy)
+                        except StopIteration:
+                            break
+                        if i % step == 0:
+                            print("yielding ", item, ", position was", position)
```
cgccuser, line 95:
```suggestion

```

chunkybanana (approved):

PR #640, opened 2022-01-20
Switch from checking length to yielding each collective prefix

### LINKED ISSUE #598: `K` (prefixes) doesn't work on infinite lazylists
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLigb3igLox4bieIEsiLCIiLCIiXQ==)

Pretty obvious what it should do

Lyxal:
Closes #598

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -785,7 +785,15 @@ def divisors(lhs, ctx):
             ),
             ctx,
         )
-    return LazyList((lhs[: x + 1] for x in range(len(lhs))))
+
+    @lazylist
+    def gen():
+        temp = []
+        for item in iterable(lhs, ctx=ctx):
+            temp.append(deep_copy(item))
+            yield temp
+
+    return gen()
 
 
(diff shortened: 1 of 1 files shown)


PyGamer0 (approved):

chunkybanana (approved):

vyxal/elements.py
```diff
@@ -785,7 +785,15 @@ def divisors(lhs, ctx):
             ),
             ctx,
         )
-    return LazyList((lhs[: x + 1] for x in range(len(lhs))))
+
+    @lazylist
+    def gen():
+        temp = []
+        for item in iterable(lhs, ctx=ctx):
+            temp.append(deep_copy(item))
```
cgccuser, line 793: This is modifying `temp` each time. The same list is being yielded every time instead of a new copy. It should be `yield temp[:]` instead, I think.

cgccuser (commented):
Has this been tested for lazylists?

PR #660, opened 2022-01-21
fix the base_conversion helper thingy

### LINKED ISSUE #659: Compressed strings sometimes break
# [Vyxal](https://github.com/Vyxal/Vyxal), 5 bytes
```
«×Fṫ«
```
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLCq8OXRuG5q8KrIiwiIiwiIl0=)

Don't Ask Me Why ™

AMiller42:
fixes #659

### DIFF
--- a/tests/test_complex.py
+++ b/tests/test_complex.py
@@ -181,6 +181,7 @@ def test_compressed_strings():
     stack = run_vyxal("«×Fṫ«")
     assert stack[-1] == "a hyb"
 
+
 def test_to_base_digits():
     stack = to_base_digits(64, 2)
     assert stack == [1, 0, 0, 0, 0, 0, 0]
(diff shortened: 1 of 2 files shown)


pxeger (changes requested):
add tests?

tests/test_complex.py
```diff
@@ -175,3 +175,8 @@ def test_slice_to_end_infinite_lists():
 def test_interleave():
     stack = run_vyxal("⁽›1Ḟ ⁽⇧1Ḟ Y")
     assert stack[-1][:6] == [1, 1, 2, 3, 3, 5]
+
+
+def test_compressed_strings():
```
pxeger, line 180: This isn't really a unit test. It's fine to keep, but there should probably also be a direct test for the functionality of `to_base_digits`

pxeger (changes requested):

tests/test_complex.py
AMiller42: wdym? There is and definitely always has been a test for that function

pxeger (approved):

chunkybanana (approved):

PR #666, opened 2022-01-22
Make powerset work for infinite lists

cgccuser:
It's fitting that this is #666, because infinite lists sometimes feel like they came straight from hell.


Lyxal (approved):
Lgtm

chunkybanana (approved):

PR #671, opened 2022-01-23
Let `j` flag work with infinite output (WIP)

cgccuser:


cgccuser:
imma just make a new PR

PR #683, opened 2022-01-25
Implement gooder negative binary

### LINKED ISSUE #682: `b` fails for negative numbers
When using `b` on negative numbers, it errors. The reason for this is that internally, it uses the `bin()` function to convert the number to binary, removes the first two characters, then converts the resulting string to a list of digits. This works because for e.g. 10, `bin()` returns `0b1010`, and removing the first two characters gives `1010`. However, for -10, it returns `-0b1010`, so removing the first two characters gives `b1010`, and when `b` is converted to a number, it errors.

I see a couple potential solutions. The first is to simply remove 2 *or* 3 characters depending on if the number is negative or not. However, I think that the better solution would be to negate each digit of the resulting list of digits, e.g.  `10 b` would return `⟨ 1 | 0 | 1 | 0 ⟩`, but `10N b` would return `⟨ -1 | 0 | -1 | 0 ⟩`.

The reason that this would be better is that the way that `B` currently works, we are already able to convert a list of zeros and negative ones to a negative number, so this wouldn't require making any changes to `B`. It will also allow us to convert to and from binary while maintaining the sign of the number, and maintaining the truthiness of each digit in the list.

AMiller42:
Before we merge this PR, I wanna make sure that everyone agrees with this way of doing negative binary numbers, so I guess give it a thumbs up or down or smth.

Fixes #682


vyxal/elements.py
```diff
@@ -4009,12 +4009,18 @@ def vy_bin(lhs, ctx):
     (str) -> binary of each codepoint
     """
     ts = vy_type(lhs)
-    return {
-        (NUMBER_TYPE): lambda: [int(x) for x in bin(int(lhs))[2:]],
-        (str): lambda: vectorise(
-            vy_bin, wrapify(chr_ord(lhs, ctx=ctx), None, ctx), ctx=ctx
-        ),
-    }.get(ts, lambda: vectorise(vy_bin, lhs, ctx=ctx))()
+    if ts == (NUMBER_TYPE):
```
pxeger, line 4012: why the parens here? Is it supposed to be a tuple?

pxeger:
You didn't add the tests to `elements.yml`, so the bot overwrote them and they're no longer here.

AMiller42:
> You didn't add the tests to `elements.yml`, so the bot overwrote them and they're no longer here.

Alright, so to set up the element tests, you just change the `elements.yml` file, and the bot does the other stuff? I'm still figuring out how some of the tests and bots and things work.

AMiller42:
Is this how the element tests work?

AMiller42:
Apparently it is. Wow that is so much easier lol

vyxal/elements.py
Tams-Tams: Looks like in [vy-type](https://github.com/Vyxal/Vyxal/blob/d82539759c2379042bf987c5fa0e2d975499d239/vyxal/elements.py#L4367) only `NUMBER_TYPE` without parentheses are returned.

AMiller42: It works and passes the tests and everything, even with the parentheses, but I think I'll remove them anyway because I can and because why not

Tams-Tams: `2 == (2)` returns `True` in python.

Tams-Tams: That's probably typical python then

Tams-Tams (approved):

pxeger (approved):

cgccuser:
Yeah, just run `python documents/knowledge/processor.py` to test locally (or don't and the bot does it for you). We really need to document this properly in the contributing guide.

PR #684, opened 2022-01-25
Fix bool input

### LINKED ISSUE #676: Inputting `True` or `False` doesn't work properly
If you input `True` or `False` in the online interpreter, it is evaluated as a `bool`. This doesn't error in and of itself, but it does error if you try to do anything with it, because the `vyxalify` function doesn't cast it properly. This shouldn't be too much of an issue, since we can just add a special case for `bool` values in the function.

Example: [Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCIxKyIsIiIsIlRydWUiXQ==)

AMiller42:
Fixes #676

### DIFF
--- a/tests/test_complex.py
+++ b/tests/test_complex.py
@@ -257,7 +257,7 @@ def test_map_to_every_second_item():
         8,
         "          ",
     ]
-    
+
 
 def test_bool_input():
     stack = run_vyxal("1+", inputs=[True])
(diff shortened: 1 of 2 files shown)


cgccuser:
Before I approve, I just want to make sure everyone can live with this decision. If anyone has any reason for these changes not to be merged, speak now or forever hold your peace! (but seriously, if there's anything wrong with stringifying, could y'all mention that? I noticed @pxeger and @Lyxal wanted numbers)

Lyxal (approved):
Lgtm

chunkybanana (approved):
lgtm

PR #688, opened 2022-01-27
wrap online res.stdout setter in try/catch block

### LINKED ISSUE #449: Catch errors with `E` flag
at the moment the page just hangs

sporeball:
fixes #449.

line 213 is moved into a `try`/`catch` block, and `res.stdout` gets set to the result of the `catch` if necessary.


chunkybanana (approved):

Lyxal (approved):
lgtm

PR #690, opened 2022-01-27
Implement monadic filter

### LINKED ISSUE #680: `~` doesn't filter on monads
# [Vyxal](https://github.com/Vyxal/Vyxal), 8 bytes
```
12345f~∷
```
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCIxMjM0NWZ+4oi3IiwiIiwiIl0=)

AMiller42:
Fixes #680

### DIFF
--- a/tests/test_complex.py
+++ b/tests/test_complex.py
@@ -271,4 +271,4 @@ def test_tilde_monad():
 
 def test_tilde_dyad():
     stack = run_vyxal("1 2 ~+")
-    assert stack = [1, 2, 3]
+    assert stack == [1, 2, 3]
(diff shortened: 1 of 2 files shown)


AMiller42:
btw, this can be easily expanded to nilads once we decide what we want that to do by just doing `elif function_A.arity == 0`. However, I did notice while sifting through the code that when doing `~SomeNumber` e.g. `~69`, it considers the number to have an arity of 1, as opposed to niladic elements, e.g. `¤`, where it considers it to have an arity of 0. So when we figure out what to do on nilads, that will have to be tweaked.

pxeger (approved):

Lyxal (approved):
Looks good to me

PR #695, opened 2022-01-28
Add discord link

Tams-Tams:
Just added the server link to the readme. Is a pr okay for something small like this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Put simply, Vyxal exists because golfers need a golfing language - and because g
 - [Tutorial](https://github.com/Vyxal/Vyxal/blob/master/docs/Tutorial.md)
 - [Codepage](https://github.com/Vyxal/Vyxal/blob/master/docs/codepage.txt)
 -->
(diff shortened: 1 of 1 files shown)


pxeger (changes requested):
Unless this is an official discord server, I don't think this should be merged (and FWIW, I don't think we should have an official discord server. The SE chat room is good enough, and having two chat rooms will fragment discussion for no reason).

cgccuser:
@pxeger I'm pretty sure it's the "official" server, and hyper may make a bridge, reducing the fragmentation. Discord is more convenient for people outside SE.

cgccuser (approved):

Lyxal:
@pxeger fragmentation and a lack of need of a discord server has been why I've held off on making one for so long.

But @cgccuser is right in saying discord is more convenient for those not on SE - it could help bring people from numerous esolang focused servers into vyxal

chunkybanana (approved):

Tams-Tams:
@Lyxal Are there actually people who came across Vyxal, who're not on SE?

Lyxal:
@Tams-Tams there is - people from other discord servers I'm on know about Vyxal

PR #697, opened 2022-01-29
Update elements.yaml

Lyxal:


Tams-Tams:
Does x print the stack or does it push the listified stack?

Lyxal:
Prints the stack

Tams-Tams (approved):

pxeger:
I think `x` should recurse on the whole program even when not in a function

chunkybanana (approved):

PR #698, opened 2022-01-29
3 words

Fmbalbuena:
3 words: Fmbalbuena, CGCC, TNB

### DIFF
--- a/vyxal/dictionary.py
+++ b/vyxal/dictionary.py
@@ -23213,7 +23213,10 @@ x-1
 <li>
 </li>
 <kbd>
-</kbd>""".split(
+</kbd>
+TNB
+CGCC
+Fmbalbuena""".split(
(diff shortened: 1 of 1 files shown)


pxeger (changes requested):
You added these to the wrong dictionary. They should go into `contents`, not `small_dictionary`.

(going based on where similar ones like `pxeger` are)

Fmbalbuena (commented):
Yes

cgccuser:
I don't see much point to this, honestly. `TNB` and `CGCC` might be useful, but not `Fmbalbuena`. Also, the PR author hasn't made the changes yet and it doesn't seem like they will anytime soon.

Fmbalbuena:
@cgccuser  Because I can't see the edit, can you reopen?

PR #707, opened 2022-02-02
Make combiantions without replacement work for infinite lists (yes, t…

cgccuser:
…his one too)

My commit is bigger than your commit, Lyxal.

Is this more efficient than filtering powerset? Maybe, maybe not. Did I write it because I was bored? Yes. Feel free to reject it.


chunkybanana (approved):
lbtltm

Lyxal (approved):
Lgtm

PR #709, opened 2022-02-02
Don't vectorise boolify

### LINKED ISSUE #708: Boolify `ḃ` is vectorising
Given a list like `⟨1|2|3⟩` or `⟨0|``|⟨⟩⟩`, Vyxal is returning `⟨ 1 | 1 | 1 ⟩` and `⟨ 0 | 0 | ⟨  ⟩ ⟩` rather than `1` and `1` (because both lists are not empty).
[Try Bug Online](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLin6gxfDJ8M+KfqeG4gyxcbuKfqDB8YGB84p+o4p+p4p+p4biDLFxuXG4iLCIiLCIiXQ==)

Tams-Tams:
Fixes #708. Only if ya'll think it's better to use python boolify than to vectorise.


cgccuser:
I think not vectorizing is a better default, but I'll let others vote on this.

pxeger (approved):

vyxal/elements.py
```diff
@@ -415,13 +415,7 @@ def boolify(lhs, ctx):
...
     (any) -> is truthy?
     """
-    if vy_type(lhs, simple=True) is list:
-        if ctx.truthy_lists:
-            return any_true(lhs, ctx)
-        else:
-            return vectorise(boolify, lhs, ctx=ctx)
-    else:
-        return int(bool(lhs))
+    return int(bool(lhs))
```
cgccuser, line 418: If we are getting rid of the vectorised-ness, let's at least allow that old behavior under the `t` flag.

cgccuser (changes requested):

vyxal/elements.py
Tams-Tams: Ok, working on it

Tams-Tams: @cgccuser Done

Tams-Tams:
Currently, when boolifying a list like `[0|""|[]]` it return `any(list)` so it will return `0`. Unlike python, which checks if `length(list)>0`, so it will return `1`. What is better, in terms of golfing?

vyxal/elements.py
```diff
@@ -416,12 +416,12 @@ def boolify(lhs, ctx):
...
     if vy_type(lhs, simple=True) is list:
-        if ctx.truthy_lists:
-            return any_true(lhs, ctx)
-        else:
+        if ctx.vectorise_boolify:
             return vectorise(boolify, lhs, ctx=ctx)
-    else:
-        return int(bool(lhs))
+        else:
+            return any_true(lhs, ctx)
```
cgccuser, line 422: Wait, are we not doing `int(bool(lhs))` here too?

cgccuser (changes requested):

vyxal/elements.py
Tams-Tams: Forgot to change it, but what is better? Before the changes, the `t` flag checked if `any_true(list)`. Is the python version better?

Tams-Tams: @cgccuser changed it

documents/knowledge/elements.yaml
```diff
@@ -1444,12 +1444,14 @@
   arity: 1
   overloads:
     any: bool(a) (Booliify)
-  vectorise: true
+  vectorise: false
```
pxeger, line 1447: add a note here that its behaviour is changed by the `t` flag?

vyxal/main.py
```diff
@@ -120,7 +120,7 @@ def execute_vyxal(file_name, flags, inputs, output_var=None, online_mode=False):
     ctx.number_as_range = "R" in flags
     ctx.dictionary_compression = "D" not in flags
     ctx.variable_length_1 = "V" in flags
-    ctx.truthy_lists = "t" in flags  # L431 in elements.py
+    ctx.vectorise_boolify = "t" in flags  # L431 in elements.py
```
pxeger, line 123:
```suggestion
    ctx.vectorise_boolify = "t" in flags  # see boolify in elements.py
```

vyxal/main.py
```diff
@@ -49,7 +49,7 @@
     Ṫ    Print the sum of the entire stack
     ṡ    Print the entire stack, joined on spaces
     J    Print the entire stack, separated by newlines.
-    t    Lists are considered truthy if they are not empty
+    t    Vectorise boolify on Lists
```
pxeger, line 52: Hang on, isn't it the case that now all that this PR does is invert the meaning of `t`?

pxeger (dismissed):

vyxal/main.py
Tams-Tams: :face_palm: er yes

documents/knowledge/elements.yaml
Tams-Tams: as a comment?

vyxal/main.py
Tams-Tams: How should I change it?

documents/knowledge/elements.yaml
Tams-Tams: like that?

Lyxal (changes requested):
My only concern is that this was specifically requested as a feature: #227

Do you know what @AMiller42 thinks about this?

documents/knowledge/elements.yaml
```diff
@@ -1440,16 +1440,18 @@
 
 - element: "ḃ"
   name: Boolify
-  description: Convert an arbitrary value into a truthy or falsy value
+  description: Convert an arbitrary value into a truthy or falsy value, vetorises with flag t
```
cgccuser, line 1443:
```suggestion
  description: Convert an arbitrary value into a truthy or falsy value, vectorises with flag t
```

vyxal/elements.py
```diff
@@ -416,12 +416,12 @@ def boolify(lhs, ctx):
     (any) -> is truthy?
     """
     if vy_type(lhs, simple=True) is list:
```
cgccuser, line 418: Could you make this `if ctx.vectorise_boolify and vy_type(lhs, simple=True) is list:` instead? That way, the if would be simply `vectorise(boolify, lhs, ctx=ctx)` instead of a nested `if`.

cgccuser (changes requested):

AMiller42:
> My only concern is that this was specifically requested as a feature: #227
> 
> Do you know what @AMiller42 thinks about this?

I don’t think he’d care

cgccuser (approved):

AMiller42 (approved):
Aaron’s okay with it

AMiller42 (approved):
lgtm

PR #712, opened 2022-02-03
Update the call to vy-print

Lyxal:


Not-Lyxal (approved):
Is urgent that this goes through

chunkybanana (approved):

PR #733, opened 2022-02-20
Make τ vectorise

Lyxal:


SE-Nitro (approved):
Shall Vyxal be tied with Jelly!!!

razetime (approved):

PR #742, opened 2022-02-21
Implement cartesian power for infinite lists

cgccuser:
This is kinda like cartesian product, it goes diagonally (or whatever the word for multidimensional diagonals is). This implementation is only used for `LazyList`s because it's probably inefficient; lists use the original one.


cgccuser:
Oh come on, I even updated black on my end, why do you keep complaining about formatting?

Seggan (approved):

PR #749, opened 2022-02-28
As requested, `*` now vectorises if there's a function

### LINKED ISSUE #748: `*` doesn't vectorise on functions.
Don't ask why I needed this.

Seriously. Don't.

Also we should *probably* put that behaviour of `*` (and `Td`) in the docs.

Lyxal:
Closes #748 

Honestly idk why on earth you'd need this, but as you wish.

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -2418,10 +2418,14 @@ def multiply(lhs, rhs, ctx):
     ts = vy_type(lhs, rhs)
 
     if ts[0] is types.FunctionType:
+        if ts[1] in (list, LazyList):
+            return vectorise(multiply, lhs, rhs, ctx=ctx)
         lhs.stored_arity = rhs
         return lhs
 
     elif ts[1] is types.FunctionType:
+        if ts[0] in (list, LazyList):
+            return vectorise(multiply, lhs, rhs, ctx=ctx)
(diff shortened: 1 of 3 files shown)


vyxal/helpers.py
```diff
@@ -589,7 +589,9 @@ def gen():
 def primitive_type(item: Any) -> Union[str, type]:
     """Turns int/Rational/str into 'Scalar' and everything else
     into list"""
-    if type(item) in [int, sympy.Rational, str] or is_sympy(item):
+    if type(item) in [int, sympy.Rational, str, types.FunctionType] or is_sympy(
```
cgccuser, line 592: You can also use `isinstance(item, (int, sympy.Rational, str, types.FunctionType))` for that `type ... in` check

vyxal/elements.py
```diff
@@ -2418,10 +2418,14 @@ def multiply(lhs, rhs, ctx):
     ts = vy_type(lhs, rhs)
```
cgccuser, line 2418: You could set `simple=True` here and then just check for `list` below

cgccuser (approved):
Minor suggestions

Seggan (approved):
I understand none of this tbh, blind approval

PR #752, opened 2022-02-28
Add documentation for the probably unknown feature of function multiplication

Lyxal:


SE-Nitro (approved):

AMiller42:
> Who knew this stuff existed 🤔

Me, only because I know *way* too much about Vyxal and its quirks.

documents/knowledge/elements.yaml
```diff
@@ -857,12 +859,13 @@
     - '["X"] : "X"'
 
 - element: "T"
-  name: Truthy Indices
-  description: Get indices of truthy elements or triple
+  name: Truthy Indices / Triadify
+  description: Get indices of truthy elements or make the arity of a function 3
```
pxeger, line 863: This now loses the description for numbers, which is normal tripling

pxeger, line 863: This now loses the description for numbers, which is normal tripling

pxeger (changes requested):

documents/knowledge/elements.yaml
cgccuser, line 863:
```suggestion
  name: Truthy Indices / Triple / Triadify
  description: Get indices of truthy elements, triple, or make the arity of a function 3
```

cgccuser (changes requested):
Based on pxeger's comment

pxeger:
> Screw that guy

hey >:(

PR #759, opened 2022-03-05
Make Π do the same as Þ* when not given a list of numbers

### LINKED ISSUE #757: Merge `Þ*` into `Π`
`Π` is the product builtin, reducing an integer list by multiplication and doing not much else for string lists / list lists. `Þ*` reduces a list by cartesian product.

These seem perfect to merge - both have dubious, mostly useless behaviour on unintended inputs, and they're quite similar - reduce by cartesian product vs take product.

Lyxal:
Closes #757

### DIFF
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -4681,6 +4681,17 @@
     - "[[1, 2, 3, 4, 5, 6]] : [[1, 2, 3, 4, 5, 6], [2, 3, 4, 5, 6], [3, 4, 5, 6], [4, 5, 6], [5, 6], [6]]"
     - "[[]] : []"
 
+- element: "Þİ"
+  name: First n Items and Rest
+  description: a[:b] and a[b:]
+  arity: 2
+  overloads:
+    lst-int: a[:b] and a[b:]
+  vectorise: false
+  tests:
+    - "[[1, 2, 3, 4, 5, 6, 7, 8, 9], 4] : [5, 6, 7, 8, 9]"
+    - "['abcde', 2] : 'cde'"
+
 - element: "¨□"
   name: Parse direction arrow to integer
   description: Map characters in `>^<v` to integers (0, 1, 2, 3 respectively)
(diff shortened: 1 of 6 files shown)


vyxal/elements.py
```diff
@@ -2964,8 +2964,14 @@ def prepend(lhs, rhs, ctx):
 
 def product(lhs, ctx):
     """Element Π
-    (lst) -> product(list)"""
-    return vy_reduce(multiply, lhs, ctx=ctx)
+    (lst[num]) -> product(list)
+    (lst[str|lst]) -> Cartesian product over a list of lists
+    """
+
+    if all(vy_type(x) == NUMBER_TYPE for x in lhs):
```
cgccuser, line 2971: This will hang for infinite lists

cgccuser (changes requested):

vyxal/elements.py
chunkybanana: Neither product nor reduce by cartesian product make any sense for an infinite list. Doesn't matter.

chunkybanana (commented):
String testcase please?

chunkybanana (commented):
Actually, can we make it behave like regular cartesian product on strings, and turn a string list into a string list?

vyxal/elements.py
cgccuser: You right

documents/knowledge/elements.yaml
```diff
@@ -4676,17 +4681,6 @@
     - "[[1, 2, 3, 4, 5, 6]] : [[1, 2, 3, 4, 5, 6], [2, 3, 4, 5, 6], [3, 4, 5, 6], [4, 5, 6], [5, 6], [6]]"
     - "[[]] : []"
 
-
```
cgccuser, line 4679: Wait why was this removed ?

cgccuser (changes requested):
Looks good but why was PI removed and not P*?

documents/knowledge/elements.yaml
Lyxal: It wasn't. That's the result of github branch frickery

chunkybanana (approved):

cgccuser (approved):

Issue #741, opened 2022-02-21
Diagraphs for checking sortedness


AMiller42:
We should have some diagraphs for checking various sortednesses. At the moment, just checking if a list/string is sorted requires 3 bytes: `:s=`. Having this as a diagraph would be nice, and while we're at it, I think we should add a few more. I think we should add diagraphs to check if a list/string:

- [ ] Is sorted, e.g. `[1,2,3] → 1` `[3,2,1] → 0` `[1,3,2] → 0`
- [ ] Is reverse sorted, e.g. `[1,2,3] → 0` `[3,2,1] → 1` `[1,3,2] → 0`
- [ ] Is either sorted/reverse sorted, e.g. `[1,2,3] → 1` `[3,2,1] → 1` `[1,3,2] → 0`
- [ ] Is unsorted, e.g. `[1,2,3] → 0` `[3,2,1] → 0` `[1,3,2] → 1`

Let me know if you need any elaboration. Feel free to propose similar diagraphs and I'll add them to the list.

Lyxal:
Just on the "is unsorted" digraph - how would you define sorted? Because `[1, 3, 2]` could be considered sorted by some sort of ordering.

AMiller42:
> Just on the "is unsorted" digraph - how would you define sorted? Because `[1, 3, 2]` could be considered sorted by some sort of ordering.

This is pretty much what I had in mind:

Sorted: `:s⁼`
Reverse Sorted: `:sṘ⁼`
Sorted/Reverse Sorted: `₌‡s⁼≬sṘ⁼∨`
Unsorted: `₌‡s⁼≬sṘ⁼∨¬`

pxeger:
If there were a modifier for "is equal after applying", then an equivalent effect could be achieved, but with much a more general builtin.

AMiller42:
> If there were a modifier for "is equal after applying", then an equivalent effect could be achieved, but with much a more general builtin.

I agree, this is probably a much better idea

PR #774, opened 2022-03-09
Add ¨= modifier (is equal after applying)

Lyxal:
Based on @pxeger's suggestion in #741

And yes, you can have modifiers that are also digraphs. Believe it or not, I wrote the parser with digraphic modifiers in mind.


cgccuser (approved):
Cool modifier, fancy name

TheSecondComing123 (approved):
Lgtm

PR #775, opened 2022-03-09
Add Sorting Digraphs (ÞṠ, ÞṘ, ÞȮ, and ÞĊ)

### LINKED ISSUE #741: Diagraphs for checking sortedness

Lyxal:
Closes #741

### DIFF
--- a/documents/knowledge/elements.md
+++ b/documents/knowledge/elements.md
@@ -3307,6 +3307,38 @@ Sort a list by length.
 
 - lst a: `Sort a list by length.`
 -------------------------------
+## `` ÞṠ `` (Is Sorted?)
+
+Returns true if an item is sorted in ascending order using default sorting rules.
+
+### Overloads
+
+- lst a: `Returns true if an item is sorted in ascending order using default sorting rules.`
+-------------------------------
+## `` ÞṘ `` (Is Sorted in Reverse?)
+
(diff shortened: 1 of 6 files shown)


Seggan (approved):

chunkybanana (approved):

chunkybanana:
Note that this'll probably be removed once we get a "is equal under <element>" modifiier.

PR #777, opened 2022-03-09
Add while loop counter

### LINKED ISSUE #776: Counter for while loop
It would be nice to have a counter `n` for a while loop too.
[Try it online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCIxe24sIiwiIiwiIl0=)
So it would print 1, 2, 3, 4, 5, ...
Currently, it requires [7 bytes (I think)](https://vyxal.pythonanywhere.com/#WyIiLCIiLCIxwqN7wqXigKbigLrCoyIsIiIsIiJd)

SE-Nitro:
closes #776 
Now eg. `{n,` outputs:
```
1
2
3
...
```
Should the counter start at 1?

### DIFF
--- a/vyxal/main.py
+++ b/vyxal/main.py
@@ -26,7 +26,7 @@ FLAG_STRING = """ALL flags should be used as is (no '-' prefix)
     j    Print top of stack joined by newlines on end of execution
     L    Print top of stack joined by newlines (Vertically) on end of execution
     s    Sum/concatenate top of stack on end of execution
-    M    Make implicit range generation start at 0 instead of 1 (and while loop counter too)
+    M    Make implicit range generation and while loop counter start at 0 instead of 1
     m    Make implicit range generation end at n-1 instead of n
     Ṁ    Equivalent to having both m and M flags
     v    Use Vyxal encoding for input file
--- a/vyxal/main.py
+++ b/vyxal/main.py
@@ -26,7 +26,7 @@ FLAG_STRING = """ALL flags should be used as is (no '-' prefix)
     j    Print top of stack joined by newlines on end of execution
     L    Print top of stack joined by newlines (Vertically) on end of execution
     s    Sum/concatenate top of stack on end of execution
-    M    Make implicit range generation start at 0 instead of 1
+    M    Make implicit range generation start at 0 instead of 1 (and while loop counter too)
     m    Make implicit range generation end at n-1 instead of n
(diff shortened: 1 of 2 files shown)


Seggan:
> Should the counter start at 1?

 I vote 0

TheSecondComing123:
I vote 0.

TheSecondComing123 (changes requested):
I vote it start at 0.

SE-Nitro:
[Poll](https://strawpoll.com/polls/bVg86b84zyY)

SE-Nitro:
It would make sense to start with 1, because in vyxal [numbers also start with 1 when used as ranges (in a loop)](https://vyxal.pythonanywhere.com/#WyIiLCIiLCI1KG4sIiwiIiwiIl0=)

SE-Nitro:
> 

So, you mean, start at 0 by default and start at 1 with flag `M`?

chunkybanana:
0 because ( behaves like that.

AMiller42:
> 0 because ( behaves like that.

[But it don't tho.](https://vyxal.pythonanywhere.com/#WyIiLCIiLCI2OShuLCIsIiIsIiJd)

SE-Nitro:
Now, the default is `1` and `M` makes it `0`

SE-Nitro:
> > 0 because ( behaves like that.
> 
> [But it don't tho.](https://vyxal.pythonanywhere.com/#WyIiLCIiLCI2OShuLCIsIiIsIiJd)

<strike>funny number</strike>

vyxal/main.py
```diff
@@ -26,7 +26,7 @@
     j    Print top of stack joined by newlines on end of execution
     L    Print top of stack joined by newlines (Vertically) on end of execution
     s    Sum/concatenate top of stack on end of execution
-    M    Make implicit range generation start at 0 instead of 1
+    M    Make implicit range generation start at 0 instead of 1 (and while loop counter too)
```
Seggan, line 29: Maybe change to `M Make implicit range generation and the while loop counter start at 0 instead of 1

Seggan (changes requested):

vyxal/main.py
edison1119: `Make implicit range generation and loop counter start at 0 instead of 1`

Seggan (approved):

TheSecondComing123 (approved):

PR #785, opened 2022-03-11
Make Ṁ insert in multiple places if given a list

### LINKED ISSUE #783: `Ȧ` and `Ṁ` should try to find an integer argument
Currently `Ȧ` and `Ṁ` (assign + insert) require an integer in the second position. It'd be more versatile if they tried to find an integer argument instead of erroring.

Lyxal:
Closes #783


Seggan (approved):

chunkybanana (approved):
Not what I suggested, but better than that.

Issue #790, opened 2022-03-14
Chained stack modifiers don't actually operate on the stack


chunkybanana:
# [Vyxal](https://github.com/Vyxal/Vyxal), 12 bytes
```
3 4₌≬+:_≬+:_
```
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCIzIDTigoziiawrOl/iiawrOl8iLCIiLCIiXQ==)

for example

This could be quite hard to fix, I don't really understand the internal structure of Vyxal's modifiers.

chunkybanana:
Speaking of which, pretty sure modifier grouping isn't documented.

Lyxal:
here's the thing:

```
3 4₌λ2|+:_;λ2|+:_;
```

does what I think you intend.

but so should 

```
3 4₌‡≬+:_d‡≬+:_d
```

but it doesn't

Lyxal:
@chunkybanana about my above remark - what do you think?

chunkybanana:
I agree that that should just push the function, not try and call it. 

Also, would setting the arity to len(stack) dynamically work?

Lyxal:
You don't need to set it to `len(stack))` - setting it to `-1` should automatically use stack if that was actually implemented.

Lyxal:
> Also, would setting the arity to len(stack) dynamically work?

On second thought, no, it wouldn't work. Because what happens if you try something like `₌+-` on an empty stack with two inputs? `len(stack)` will set the arity to `0` and not pop any inputs despite the fact that there needs to be stuff popped.

PR #793, opened 2022-03-14
Make modifiers use the default arity variable (which can be changed b…

Lyxal:
…ased on flags) instead of assuming arity of 1

Goes towards fixing #790


chunkybanana (approved):
Looks fine :)

PyGamer0 (approved):

PR #799, opened 2022-03-15
Make the argument order of `∧` and `∧` swapped

Lyxal:

PR #800, opened 2022-03-15
Update Production (15/03/2022 and 16/03/2022)

Lyxal:
Merge after:

- [x] #793 
- [x] #796 
- [x] #797 
- [x] #798 
- [x] #799
- [x] #801 
- [x] #805


chunkybanana (approved):
Here's an approval so this can be merged when necessary.

PR #801, opened 2022-03-16
Improve the functionality of the rotate elements

Lyxal:
Also, add an `is_empty` method to `LazyList`

### DIFF
--- a/vyxal/LazyList.py
+++ b/vyxal/LazyList.py
@@ -269,3 +269,10 @@ class LazyList:
         self.generated += list(itertools.tee(self.raw_object)[-1])
         for item in self.generated[::-1]:
             yield item
+
+    def is_empty(self):
+        try:
+            next(self)
+            return False
+        except StopIteration:
+            return True
--- a/vyxal/LazyList.py
+++ b/vyxal/LazyList.py
@@ -269,10 +269,3 @@ class LazyList:
         self.generated += list(itertools.tee(self.raw_object)[-1])
         for item in self.generated[::-1]:
             yield item
-
-    def is_empty(self):
-        try:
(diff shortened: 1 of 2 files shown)


chunkybanana:
Isn't there just `lazylist.__bool__`?

vyxal/LazyList.py
```diff
@@ -269,3 +269,10 @@ def reversed(self):
         self.generated += list(itertools.tee(self.raw_object)[-1])
         for item in self.generated[::-1]:
             yield item
+
+    def is_empty(self):
+        try:
+            next(self)
```
cgccuser, line 275: This throws away the result of next

cgccuser (changes requested):

vyxal/LazyList.py
cgccuser: Try has_ind(0) maybe

Lyxal:
```py
 def has_ind(self, ind: int):
    """Whether or not this list is long enough for index `ind`"""
    if ind < len(self.generated):
        return 0 <= ind
    else:
        for _ in range(ind - len(self.generated) + 1):
            try:
                next(self)
            except StopIteration:
                return False
        return True
```

by that logic, `has_ind` should also throw away the result of `next`.

But it doesn't, because `next` is smart enough to store the item in `self.generated`

chunkybanana (changes requested):
This needs to be `not bool(lhs)`

chunkybanana (approved):

vyxal/LazyList.py
cgccuser: Oh right

PR #811, opened 2022-03-20
Add a starmap modifier

Lyxal:
It's equivalent to `Zvƒ`.

Idea borrowed from Add++ of all places.

No one tell caird we added something from Add++ lol.


PyGamer0 (approved):

cgccuser (approved):
Tests would be nice but LGTM, could be quite useful.

PR #817, opened 2022-03-21
Make lambdas with arity -1 use the stack as their input

### LINKED ISSUE #415: Commands using the stack break with modifiers
Commands that operate on some portion of the stack instead of popping from the stack (e.g. `W`, `!`, `^`, etc.) will operate on an empty stack when given modifiers such as `~₌₍`. The exceptions are the `„‟` commands, which just error instead, and the `v` modifier, which errors for any of the tested commands.

Example: [Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCIxIDIgMyA0IDUgNiB+XiBXLFxuMSAyIDMgNCA1IDYgflcgVyxcbjEgMiAzIDQgNSA2IH4hIFcsXG4xIDIgMyA0IDUgNiDigo3IriEgVyxcbjEgMiAzIDQgNSA2IOKCjF5XIFcsXG4xIDIgMyA0IDUgNiB+4oCeIFcsIiwiIiwiIl0=)

Lyxal:
And make modifiers account for that.

Closes #415


Lyxal:
So here's a fun thing - `function_call` and `safe_apply` and what not need to account for the stored arity of a lambda when doing their stuff.

12:07am probably isn't the best time for working on this.

chunkybanana (approved):
lgtm although I don't really understand modifiers

PR #818, opened 2022-03-22
Make sure to cast the result of integration to string

Lyxal:


cgccuser (commented):
Wait, what if you need it to be a number? Could we instead check if it can be converted to a number and only cast to string if not?

Lyxal:
> Wait, what if you need it to be a number? Could we instead check if it can be converted to a number and only cast to string if not?

Integration is a string overload.

cgccuser:
Right, but the result can also be a number instead of a symbolic thingy, right?

Lyxal:
Very rarely will the result of integration be a number. And even still, it's best to keep it as a string for consistency.

chunkybanana (approved):

chunkybanana (approved):

PyGamer0 (approved):

PR #823, opened 2022-03-24
Improve the documentation of `¨M`

### LINKED ISSUE #822: Add Apply at Index(cies) Element
Something like:
`lst a, lst b, fun c: [c(n) if i in b else n for n, i in enumerate(a)]`
`lst a, num b, fun c: a[b] = c(a[b])`

Lyxal:
Closes #822

### DIFF
--- a/documents/knowledge/elements.md
+++ b/documents/knowledge/elements.md
@@ -3460,13 +3460,14 @@ Usage:
 ```
 
 -------------------------------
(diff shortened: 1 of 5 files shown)


Seggan (approved):
Are we not adding the num overload?

chunkybanana (approved):
At some point, could we make this a single byte?

Issue #807, opened 2022-03-20
Better matrix / graph builtins


chunkybanana:
Specifically:

- [x] Matrix exponentiation
Repeated multiplication of the matrix with itself
- [x] Identity matrix of size n
A matrix with 1s on the main diagonal and zeroes elsewhere
E.g. for 3
```
[ [1, 0, 0],
  [0, 1, 0],
  [0, 0, 1] ]
```
- [x] Graph adjacency matrix
Take a graph of the form `[[1,3],[2,4],[3,4]]` and create an adjacency matrix e.g. for this,
```
[ [0, 0, 1, 0],
  [0, 0, 0, 1],
  [1, 0, 0, 1],
  [0, 1, 1, 0] ]
```
- [x] Graph distance matrix
Take a graph of the form `[[1,3],[2,4],[3,4]]` and create a distance matrix, e.g. for the above,
```
[ [0, 3, 1, 2],
  [3, 0, 2, 1],
  [1, 2, 0, 1],
  [2, 1, 1, 0] ]
```

Lyxal:
I'mma gon need some examples and descriptions of what these do.

chunkybanana:
Information added.

PyGamer0:
> Graph adjacency matrix

Did you just yoink the test case from my challenge lol?

Lyxal:
For matrix exponentiation, wouldn't `e` just work? Or am I missing something?

Lyxal:
Also, someone else might have to do the graph stuff - I don't know much about graph theory at all (s/much/anything)

chunkybanana:
Pretty sure it doesn't because it's repeated matrix multiplication.

cgccuser:
I guess we'll need one set of builtins for undirected graphs and one for directed graphs.

PR #824, opened 2022-03-24
Add `Þe` element

Lyxal:
This is what you meant right?

Goes towards closing #807


Lyxal:
Dang. Could someone fix the test case that's erroring for me?

Lyxal:
It's got too many brackets which is probably a typo I missed because the checks didn't run after the continuous integration commit. And now it's too late time wise for me to fix rn

chunkybanana (approved):

vyxal/elements.py
```diff
@@ -2379,6 +2379,24 @@ def matrix_determinant(lhs, ctx):
...
 
+def matrix_exponentiation(lhs, rhs, ctx):
+    """Element Þe
+    (lst, num) -> (a * a) b times
+    (num, lst) -> (b * b) a times
+    """
+
+    ts = vy_type(lhs, rhs, simple=True)
+    if set(ts) != {list, NUMBER_TYPE}:
+        raise TypeError("matrix exponentiation only works on lists and numbers")
```
cgccuser, line 2390: Maybe say instead "Matrix exponentiation requires a matrix and a number"? This implies that two numbers or two matrices would work

vyxal/elements.py
```diff
@@ -2379,6 +2379,24 @@ def matrix_determinant(lhs, ctx):
...
+
+    matrix, times = lhs, rhs if ts[0] == NUMBER_TYPE else rhs
+    original_matrix = deep_copy(matrix)
```
cgccuser, line 2393: There's no need to copy, is there? `multiply` doesn't mutate its arguments

cgccuser (changes requested):
LGTM, just a couple minor things

vyxal/elements.py
Lyxal: Sure, it doesn't. But assignment does.

PyGamer0 (approved):

vyxal/elements.py
cgccuser: @Lyxal It doesn't, `original_matrix = matrix` is enough

vyxal/elements.py
```diff
@@ -2379,6 +2379,24 @@ def matrix_determinant(lhs, ctx):
...
+    (lst, num) -> (a * a) b times
+    (num, lst) -> (b * b) a times
+    """
+
+    ts = vy_type(lhs, rhs, simple=True)
+    if set(ts) != {list, NUMBER_TYPE}:
+        raise TypeError("Matrix exponentiation requires a matrix and a number")
+
+    matrix, times = lhs, rhs if ts[0] == NUMBER_TYPE else rhs
+    original_matrix = deep_copy(matrix)
```
cgccuser, line 2393:
```suggestion
    original_matrix = matrix
```

cgccuser (changes requested):

vyxal/elements.py
cgccuser: If this doesn't work, something is wrong with the rest of our code

cgccuser (approved):
LGTM

PR #828, opened 2022-03-25
Move the philosophy stuff into a different repo

Lyxal:


Seggan:
Why?

chunkybanana:
^

PyGamer0 (approved):

Lyxal:
They feel unnecessary now and also I feel like they aren't exactly genuine because I got AI to write them. Also, it feels as if though these documents would be better suited as a sort of universal organisation thing rather than language specific. 


I plan to rewrite them one day in the future. But for now, they're better not in the main repo.

Seggan:
If it is destined to get a rewrite soon™, then why move it instead of rewrite?

cgccuser (approved):
Not having these in the main repo is a good call

PR #832, opened 2022-03-27
Add CodeQL scanning

cgccuser:
Why? I don't know. But why not, amirite?


cgccuser:
Wait nvm this would have this scanning on every PR and stuff

PR #834, opened 2022-03-30
Make lazylists immutable

### LINKED ISSUE #833: Make lists immutable
Jyxal already does this and it'd make implementing copying and stuff way easier. This *would* break answers that duplicate a list, modify it in one place, and expect the other copy to have those modifications, but there are <=11 such answers, so it should be fine. It's well worth the reduction in `LazyList` bugs and simplicity.

Lyxal:
Closes #833

### DIFF
--- a/vyxal/LazyList.py
+++ b/vyxal/LazyList.py
@@ -165,9 +165,7 @@ class LazyList:
         return item
 
     def __setitem__(self, position, value):
-        if position >= len(self.generated):
-            self.__getitem__(position)
-        self.generated[position] = value
+        raise NotImplementedError("LazyList does not support assignment")
 
     def __delitem__(self, index):
         if index >= len(self.generated):
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -376,8 +376,9 @@ def assign_iterable(lhs, rhs, other, ctx):
         lhs[rhs] = other
         return vy_sum(lhs, ctx=ctx)
     else:
-        lhs[rhs] = other
-        return lhs
+        yield from lhs[:rhs]
+        yield other
+        yield from lhs[rhs + 1 :]
 
 
 def base_255_string_compress(lhs, ctx):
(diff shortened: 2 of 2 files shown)


vyxal/LazyList.py
```diff
@@ -165,9 +165,7 @@ def __next__(self):
         return item
 
     def __setitem__(self, position, value):
```
cgccuser, line 167: We can just get rid of this method tbh, no need to implement it and raise a `NotImplementedError`

vyxal/LazyList.py
```diff
@@ -165,9 +165,7 @@ def __next__(self):
...
-        if position >= len(self.generated):
-            self.__getitem__(position)
-        self.generated[position] = value
+        raise NotImplementedError("LazyList does not support assignment")
 
     def __delitem__(self, index):
```
cgccuser, line 170: This method can also be removed.

cgccuser (changes requested):
This is great, just a couple methods to delete.

cgccuser:
Should this be marked as a draft while that anything else that relies on mutability stuff gets modified?

Seggan:
+1

vyxal/LazyList.py
Lyxal: My idea was that having an error would help make it clear if there were any errors trying to assign to a lazy lists.

vyxal/LazyList.py
Lyxal: To be removed later when we're sure we got every instance of mutability.

For now, it's a failsafe.

cgccuser:
@Lyxal If the method doesn't exist, it's going to error anyway.

Lyxal:
But at least with a custom error it's clearer what's going wrong

cgccuser (approved):

Seggan (approved):

PR #849, opened 2022-04-13
Convert to int before applying bitwise ops

cgccuser:
The tests for bitwise operators all fail because sympy `Integer`s are passed into them (even if they work online). It's better to turn all numbers received by those methods into `int`s first to avoid this sort of thing.


Seggan (approved):

AMiller42 (approved):
lgtm

PR #852, opened 2022-04-14
Make `vy_exec` vectorise properly on strings.

### LINKED ISSUE #851: Executing a list of strings recursively vectorises
The `Ė` command has a helper function that will either return the reciprocal of a value if it is number, or vectorise itself over the value otherwise. The command only checks once if the value is a string, so if the string is wrapped in a list, it recursively vectorises the helper function over the value until it hits the recursion limit.

Example: [Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLin6hgYWDin6kgxJYiLCIiLCIiXQ==)

Lyxal:
Fixes #851

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -4536,15 +4536,28 @@ def vy_exec(lhs, ctx):
                 lhs, ctx.dictionary_compression, ctx.variable_length_1
             )
         )
+
         return []
 
     def helper(lhs):
         if vy_type(lhs) == NUMBER_TYPE:
             return divide(1, lhs, ctx)
+        elif vy_type(lhs) is str:
+            import vyxal.transpile
+
+            stack = ctx.stacks[-1]
+            exec(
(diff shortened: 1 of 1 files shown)


AMiller42:
The strings inside of the lists don't have access to the stack, and will instead take input. Not sure if that should be changed or not, but it is something to be aware of. It could potentially be useful, since it kind of acts like a lambda with no arguments, except that it can take implicit input.

Example:
```
Input: 69

1 2 3 `d`wĖ

Output: ⟨ 138 ⟩
```

chunkybanana:
Could it act like `₍`?

cgccuser (approved):

Steffan153 (approved):

cgccuser:
I think the current behavior is fine. It'd be a pain to change it anyway.

Steffan153:
Should we merge?

PR #857, opened 2022-04-15
Make it so that `x` or `X` after a modifier functions correctly.

### LINKED ISSUE #846: Modifiers make `x` try to recurse
If you have a vectorised command somewhere in your program, and then later you use `x` to print, it will instead recurse, causing it to error.

Example: [Try it Online!](https://vyxal.pythonanywhere.com/#WyJjIiwiIiwiYG9vZmB2cVxueCIsIiIsIiJd)

Lyxal:
Fixes #846

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1,3 +1,4 @@
+from ast import Break
 import os
 import sys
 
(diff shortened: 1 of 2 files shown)


Steffan153 (approved):

cgccuser (changes requested):
Looks good, but could you write a test to make sure this works?

cgccuser (approved):
Tests look good

PR #858, opened 2022-04-15
Add replace nth occurrence digraph

Steffan153:


cgccuser (approved):
Great work, and thanks for adding tests! Btw, you can remove some of them and only keep the ones that are edge cases

Lyxal (approved):
Wowie a quadgraph. Never seen one of those before. Very cool.

cgccuser:
Wait a second, have we implemented vectorization for tetrads? It only goes up to triads right? @Lyxal

Lyxal:
That's a good point. I don't know how you'd even do 4 argument vectorisation.

Steffan153:
this element shouldn't need to vectorize though.

Lyxal:
User means how is it handled if `v` is used with it.

Steffan153:
oh, yeah. it shouldn't be hard to implement it. just need a bit of ~~elbow~~ keyboard grease.

PR #862, opened 2022-04-17
Make YAML conversion work with tetrads

Steffan153:


Lyxal (approved):

documents/knowledge/yaml_to_js.py
```diff
@@ -48,7 +48,7 @@
                 for overload in element["overloads"]:
                     data_types = map(
                         lambda x: " ".join(x),
-                        zip(overload.split("-"), "abc"),
+                        zip(overload.split("-"), "abcde"),
```
cgccuser, line 51: A tetrad can have far more than 5 overloads. We should handle that properly instead of just manually adding a letter to this string literal every time an element has too many overloads

cgccuser (changes requested):

documents/knowledge/yaml_to_js.py
Steffan153: wdym? this doesn't have to do with how many overloads there is. that doesn't matter. it's just how many arguments there are. so a tetrad has four, so we don't even use the `e`. that would only be used for pentads.

cgccuser: Oh, my bad, I should've read it better

cgccuser (approved):

PR #864, opened 2022-04-17
Fix an inaccurate piece of documentation (reverse monad)

Steffan153:


AMiller42 (approved):
nice catch, it was probably copied from `Bifurcate` and didn't get changed

Seggan (approved):

PR #884, opened 2022-04-21
Dyadic & triadic map

Steffan153:


Seggan (approved):
This is gonna be quite useful

Steffan153:
sorry some of these tests are failing... ill fix that later

Steffan153:
This doesn't quite yet close issue 494 as I need to do dyadic/triadic filter too

cgccuser (approved):
I agree with Seggan, very pogchamply of you or whatever it is gamers say

PR #886, opened 2022-04-22
dyadic/triadic filter

### LINKED ISSUE #494: Dyadic / triadic map
Maybe `¨2...;` or `¨3...;`? The code inside is a dyadic/triadic function, called like Javascript's `map`.

For `¨2`, the function should be passed the value and the index. (What order should these be in?)

For `¨3`, the function should be passed the value, index and array.

Also, map and filter (`M` / `F`) should accept dyadic / triadic functions and call them like this.

Steffan153:
closes #494


cgccuser (approved):
lgtm

Lyxal (approved):
lgtm

PR #890, opened 2022-04-23
the worstest possible way to fix the keyboard

Steffan153:
well it works for now


Lyxal (approved):
It's a good quick fix that will have to be properly addresses later.

lgtm

PR #912, opened 2022-05-01
Make ṡ return a Lazylist object instead of a range

### LINKED ISSUE #910: `ṡ` (range) returns Python `range` object instead of `list`/`LazyList`
Makes other elements bug up.

Lyxal:
Closes #910

I honestly don't know how this happened.


Steffan153 (approved):
> I honestly don't know how this happened.

I looked in the git blame and you're the one that added the command :>)

cgccuser (approved):

PR #947, opened 2022-05-08
Increment decrement until false monographs

### LINKED ISSUE #935: `∆›` and `∆‹` can (should) be single characters
We have plenty of space for function overloads, so we may as well give these single-character versions.

Steffan153:
Closes #935


Steffan153:
whoops how did those commits from 932 get in there

Steffan153:
What in the world

Lyxal (approved):

cgccuser (approved):
Might want to remove the old elements entirely but it doesn't hurt to keep them around, I guess

Steffan153:
It would break old code if we removed them. We can remove in v3

PR #949, opened 2022-05-08
Fix ACE

### LINKED ISSUE #948: Fix the ACE that has been here since v2.7.2
The cop post is now safe, so we need to patch this now. Look in the security team for the explanation.

Steffan153:
Closes #948


Lyxal (approved):
You should probably get your git base sorted out lol

Steffan153:
Lmao what in the world I dind't even notice that

I definitely checked out main this time

Steffan153:
Should add auto-label for that

Lyxal:
in the process of doing so now

cgccuser (approved):

PR #959, opened 2022-05-10
Add star map to modifier list

Steffan153:
it made it be the same as if you didn't use it at all


cgccuser (approved):
This is hilarious, good catch

Lyxal (approved):

PR #962, opened 2022-05-11
Add semi-optimal number compress

chunkybanana:


chunkybanana:
This is `øċ` and it's slightly brute-forcy and intensive, so it's a separate element from `øC`

Lyxal (approved):

Steffan153 (approved):

PR #975, opened 2022-05-13
Add `∆ǐ` to get prime factorisation + exponents

### LINKED ISSUE #969: Add element to get prime factorization with exponents
It would be Vyxal `ǐĠv₍hL`: [Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLHkMSgduKCjWhMIiwiIiwiNzIwIl0=)

We could also add one to only return the exponents.

Lyxal:
Closes #969


cgccuser (approved):

Steffan153 (changes requested):
[As emanresu pointed out](https://github.com/Vyxal/Vyxal/issues/969#issuecomment-1124505524), we can already do this in two bytes: `ǐĊ`. So this is redundant, I think. I left the issue open, though, so we could add one to only return the exponents, like 05AB1E `Ó`. You could also keep the current element, but make it also return zero exponents, e.g. `[2, 0], [3, 2], [5, 3], [7, 0]`.

chunkybanana:
Per Steffan's comment, I'm closing this for now.

PR #977, opened 2022-05-13
Add group by function overload to ġ

### LINKED ISSUE #964: Group by builtin
(can be a monograph)

A builtin that takes a function and a list, and groups the values of a list by the result when the function is applied to each of them.

Lyxal:
Closes #964

### DIFF
--- a/vyxal/helpers.py
+++ b/vyxal/helpers.py
@@ -360,7 +360,6 @@ def group_by_function(
 ) -> LazyList:
     """Group a list of elements by a function"""
 
-    @lazylist
     def gen():
         ret = {}
         for el in lst:
--- a/vyxal/helpers.py
+++ b/vyxal/helpers.py
@@ -355,6 +355,25 @@ def graph_distance(
         )
 
 
+def group_by_function(
+    lst: VyList, function: types.FunctionType, ctx: Context
+) -> LazyList:
+    """Group a list of elements by a function"""
+
+    @lazylist
+    def gen():
+        ret = {}
+        for el in lst:
+            key = safe_apply(function, el, ctx=ctx)
+            if key in ret:
+                ret[key].append(el)
+            else:
+                ret[key] = [el]
+        yield from ret.values()
+
+    return gen()
+
+
 def has_ind(lst: VyList, ind: int) -> bool:
     """Whether or not the list is long enough for that index"""
     if isinstance(lst, LazyList):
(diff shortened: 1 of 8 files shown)


vyxal/helpers.py
```diff
@@ -355,6 +355,25 @@ def graph_distance(
         )
 
 
+def group_by_function(
+    lst: VyList, function: types.FunctionType, ctx: Context
+) -> LazyList:
+    """Group a list of elements by a function"""
+
+    @lazylist
+    def gen():
```
cgccuser, line 364: There’s no need to make this a LazyList. It goes over the entire list before yielding all the values at once. If you like, I could try an infinite-list-friendly implementation later but it’s probably not going to work

cgccuser (changes requested):

vyxal/helpers.py
Steffan153, line 364:
```suggestion
    """Group a list of elements by a function"""

    def gen():
```

Steffan153 (approved):

vyxal/helpers.py
```diff
@@ -355,6 +355,24 @@ def graph_distance(
...
+        ret = {}
+        for el in lst:
+            key = safe_apply(function, el, ctx=ctx)
+            if key in ret:
+                ret[key].append(el)
+            else:
+                ret[key] = [el]
+        yield from ret.values()
+
+    return gen()
```
cgccuser, line 373: I meant to take out the `gen` function entirely, you sussy bakas
```suggestion
    ret = {}
    for el in lst:
        key = safe_apply(function, el, ctx=ctx)
        if key in ret:
            ret[key].append(el)
        else:
            ret[key] = [el]
    return ret.values()
```

cgccuser (changes requested):

vyxal/helpers.py
```diff
@@ -355,6 +355,21 @@ def graph_distance(
...
+    """Group a list of elements by a function"""
+
+    ret = {}
+    for el in lst:
+        key = safe_apply(function, el, ctx=ctx)
+        if key in ret:
+            ret[key].append(el)
+        else:
+            ret[key] = [el]
+    return ret.values()
```
cgccuser, line 370:
```suggestion
    return list(ret.values())
```

cgccuser (approved):

PR #979, opened 2022-05-13
Fix product of `[]`

### LINKED ISSUE #978: `Π` returns `0` for an empty list
By mathematical definition, the product of [] = 1.

Steffan153:
Closes #978


cgccuser (changes requested):
This won't work when the product is *supposed* to be 0

vyxal/elements.py
```diff
@@ -3376,7 +3376,7 @@ def product(lhs, ctx):
     """
 
     if all(vy_type(x) == NUMBER_TYPE for x in lhs):
-        return vy_reduce(multiply, lhs, ctx=ctx)
+        return vy_reduce(multiply, lhs, ctx=ctx) or 1
```
cgccuser, line 3379:
```suggestion
        return foldl(multiply, lhs, initial=1, ctx=ctx)
```

cgccuser (changes requested):

cgccuser:
Good catch, maybe sum and other elements also need to be fixed?

cgccuser (approved):

Lyxal (approved):

vyxal/helpers.py
```diff
@@ -165,9 +165,6 @@ def foldl(
     ctx: Context,
 ) -> Any:
     """Reduce vector by function"""
-    if len(vector) == 0:
-        return 0
```
chunkybanana, line 169: This breaks other elements that use foldl.

chunkybanana (changes requested):

vyxal/helpers.py
Steffan153: It doesn't, notice the `initial=None`, and if `initial` is `None`, it defaults to `0`. (I don't know why user didn't just do `initial=0`).

cgccuser: The fault is in those other elements, not `foldl`. They should give an initial value or check for empty lists themselves.

PR #984, opened 2022-05-14
Implement all indices

### LINKED ISSUE #971: add element to get all indices of an element
multidimensional would also be nice

(created by Steffan [here](https://chat.stackexchange.com/transcript/message/61112404))

cgccuser:
Will eventually close #971. Also modifies `enumerate_md` to act more like Python's enumerate (yields tuples `(index, item)` instead of just indices)

### DIFF
--- a/documents/knowledge/elements.md
+++ b/documents/knowledge/elements.md
@@ -3221,6 +3221,15 @@ Index a list of coordinates into a value.
 
 - lst a, lst b: `a[b[0]][b[1]][b[2]]... Reduce by indexing with a as initial value`
 -------------------------------
+## `` ÞI `` (All Indices)
+
+All indices of element in list (multidimensional)
+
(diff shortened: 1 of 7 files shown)


Lyxal (dismissed):
Part 1 lgtm

Steffan153 (dismissed):
Should we have a separate one for non-multi-dimensional?

cgccuser:
@Steffan153 Sure

cgccuser:
While this seems to work okay right now, `enumerate_md` is not safe to use with a finite list of infinite lists, let alone an infinite list of infinite lists. Once we get it working with infinite lists of infinite lists, implementing other builtins will also be easier.

Seggan (approved):
*pretends like I totally understood this*

Lyxal (approved):

PR #1008, opened 2022-05-24
Add modular exponentation

### LINKED ISSUE #1006: exponentiation modulo builtin
A triad similar to python's `pow` - a builtin that exponentiates two numbers, modulo a third, efficiently.

Maybe `∆%`?

Steffan153:
Closes #1006


vyxal/elements.py
```diff
@@ -5572,6 +5572,16 @@ def separate_runl_encode(lhs, ctx: Context):
...
 
 
+def mod_pow(lhs, rhs, other, ctx: Context):
+    """Element ∆%
+    (any, any, any) -> a ** b mod c
+    """
+    ts = vy_type(lhs, rhs, other, simple=True)
+    if list in ts:
+        return vectorise(mod_pow, lhs, rhs, other, ctx=ctx)
+    return pow(int(lhs), int(rhs), int(other))
```
Lyxal, line 5582: Make sure this is converted to a Sympy type - raw python ints will confuse the printing and typing system

Lyxal (changes requested):

vyxal/elements.py
Steffan153:
```suggestion
    return sympy.nsimplify(pow(int(lhs), int(rhs), int(other)), rational=True)
```

Lyxal (approved):
Looks good to me

chunkybanana (approved):

PR #1009, opened 2022-05-24
Fix `-1` indice with Ȧ

### LINKED ISSUE #1007: `Ȧ` does not work on `-1`
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLin6gxfDJ8M3w04p+pIHUgNSDIpiIsIiIsIiJd)

It applies it, but then concatenates the whole original array to itself.

Steffan153:
It was actually not a probelm with map at indices, it was a problem with this.

Fixes #1007.


vyxal/elements.py
```diff
@@ -419,7 +423,8 @@ def assign_iterable(lhs, rhs, other, ctx):
         def gen():
             yield from lhs[:rhs]
             yield other
-            yield from lhs[rhs + 1 :]
+            if rhs != -1:
```
chunkybanana, line 426: Does this work for all negative numbers?

Steffan153: Yes, it already worked with other negative numbers. Only `-1` did not work.

chunkybanana (approved):

Lyxal (approved):

PR #1011, opened 2022-05-25
Fixed lift on infinite lists

chunkybanana:
It used to hang, now it works


Lyxal (approved):

cgccuser (approved):
LGTM but how about this?
```python
@lazylist
def lift(lhs, ctx):
    """Element Þż
    (any) -> a * 1...a.length
    """
    i = 1
    for item in iterable(lhs, ctx=ctx):
        yield item * i
        i += 1
```

PR #1037, opened 2022-05-28
Added 0-indexed fibonacci

### LINKED ISSUE #1027: Fibonacci starts with 1
False. The fact that in Vyxal Fibonacci starts with 1 has already cost bytes. Let's move that to 0.

(created by Seggan [here](https://chat.stackexchange.com/transcript/message/61231750))

chunkybanana:
Fixes #1027


Lyxal (approved):
Heh, sorry about that. Bad language design on my part.

SE-Nitro (approved):
nice, but won't that break existing answers?

chunkybanana:
> nice, but won't that break existing answers?

It's a different builtin - `∆f` vs `∆F`

PR #1052, opened 2022-05-31
Fix group by

Steffan153:
It used to not preserve order. this also makes a string return a list of strings


Lyxal (approved):

Lyxal:
just make sure to update the tests

Seggan (approved):

PR #1072, opened 2022-06-08
Add explicit STDIN element

### LINKED ISSUE #1070: Explicit STDIN element
Like `?` but reads STDIN even if there's argv

Lyxal:
Closes #1070

And tests aren't coming because there's no good way to easily test the usage of STDIN


Steffan153 (approved):
Not sure how useful this is but LGTM

PR #1075, opened 2022-06-08
Add string interpolation

### LINKED ISSUE #1073: Add string interpolation
Basically we need to find a character that's not yet used by the short dictionary, and every time it is used in a string, it will pop from the stack and interpolate. (This can be done now, not in v3.)

Steffan153:
Closes #1073


Seggan (approved):

Seggan:
This is gonna be harder in Myxal

chunkybanana (changes requested):
We need to remove `Π` from the compression dictionary

Steffan153:
It can still be used as a second character, like `∑Π`, but it's not currently used as a first character. I will remove that

Steffan153:
But that's gonna be a *pain* to change to make it never be used as a first character, but still as a second to not break old code. The dictionary probably won't be changed in v2 to be big enough to use `Π` anyway...

cgccuser (approved):
Nice feature

chunkybanana (approved):
Fine, but this will break dictionay compression

PR #1085, opened 2022-06-09
Fix coords deepmap and others

### LINKED ISSUE #1083: The new monograph overload for fill by coordinates does not work
My bad

```
  File "/home/Vyxal/mysite/vyxal/main.py", line 164, in execute_vyxal
    exec(code, locals() | globals())
  File "<string>", line 91, in <module>
  File "/home/Vyxal/mysite/vyxal/elements.py", line 4279, in split_on
    return coords_deepmap(lhs, rhs)
TypeError: coords_deepmap() missing 1 required positional argument: 'ctx'
```

### LINKED ISSUE #1084: `ÞZ` is broken with `LazyList`s
it uses `isinstance(b, list)` which obviously does not work.

A quick Ctrl+F finds that `Þj`, `Þ…`, and `ÞU` have the same problem.

Steffan153:
Fixes #1083
Fixes #1084


chunkybanana (dismissed):
Your tests are sus.

Looking at the details, the process terminated after 53 minutes with a SIGKILL, meaning there's probably an infinite / inefficient loop somewhere in there.

vyxal/elements.py
```diff
@@ -859,8 +859,12 @@ def gen():
...
 
             diag_num = 0
-            lhs_max = len(lhs) - 1 if isinstance(lhs, list) else None
-            rhs_max = len(rhs) - 1 if isinstance(rhs, list) else None
+            lhs_max = (
+                len(lhs) - 1 if vy_type(lhs, simple=True) == list else None
+            )
+            rhs_max = (
+                len(rhs) - 1 if vy_type(rhs, simple=True) == list else None
+            )
```
cgccuser, line 867: The `len(lhs) if isinstance(lhs, list)` is so that `len` is only called on lists, not `LazyList`s. This change'll break infinite it for `LazyList`s.

cgccuser (changes requested):
Breaking change in cartesian product

vyxal/elements.py
```diff
@@ -4926,7 +4932,7 @@ def uniquify_mask(lhs, ctx):
     """
     lhs = iterable(lhs, ctx=ctx)
     # TODO (user/cgccuser): Reduce code duplication here?
-    if isinstance(lhs, list):
+    if vy_type(lhs, simple=True) == list:
```
cgccuser, line 4935: This will again break it for infinite lazylists. However, instead of going back to the old code, you could use the `def gen()` thing for finite Python `list`s too.

cgccuser (dismissed):

vyxal/elements.py
Steffan153, line 867:
```suggestion
            lhs_max = len(lhs) - 1 if isinstance(lhs, list) else None
            rhs_max = len(rhs) - 1 if isinstance(rhs, list) else None
```

cgccuser (approved):
LGTM

PR #1096, opened 2022-06-12
Fixed chr_ord on []

### LINKED ISSUE #1095: `chr_ord` fails for empty string
the specification `ord(a) if length 1 else list of ordinals` means the result of `chr_ord` on an empty string should be the empty list, but right now it errors.

chunkybanana:
Fixes #1095


cgccuser (approved):
Formatting kinda sus but lgtm

Lyxal (approved):
Black do be like that sometimes

PR #1099, opened 2022-06-14
Make Matrix Exponentiation work properly

### LINKED ISSUE #1098: Matrix exponentiation never worked
[Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLhuo86ds61PTpcbjrDnuG5gCxcbjoyw55lLCIsIiIsIlsyLCAxLCAxLCAxXSJd) They should be equivalent

This might be a mutability-related bug.

***THIS IS WHY WE HAVE TESTS***

Lyxal:
Closes #1098

It seems that when I first implemented it, I didn't actually get the right definition of matrix exponentiation. I had the matrix plain multiplied by itself n times instead of using the actual matrix multiplication function.

### DIFF
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -1660,7 +1660,7 @@
   arity: 1
   overloads:
       num: round(a)
-      complex: '[real(a), imag(a)]'
+      complex: "[real(a), imag(a)]"
       str: quad palindromize with overlap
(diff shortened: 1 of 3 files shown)


cgccuser (approved):
LGTM, good catch

chunkybanana (approved):

PR #1100, opened 2022-06-14
Make the arguments of `ß` go in the right order

Lyxal:
No related issue, just something that came up in chat.


cgccuser (approved):

chunkybanana (approved):
This works for triads, right?

Lyxal:
yes

PR #1103, opened 2022-06-17
Write more element tests

cjquines:
there are too few tests.

we should probably either fix the failures, or clarify the spec and remove/update inapplicable tests.

todo:
- [x] actually get tests running (i don't think they're failing, they're just not running)
- [x] more tests up to digits
- [x] more tests up to uppercase
- [x] more tests up to lowercase
- [x] more tests up to dotted lowercase
- [x] more tests up to sorting lambda
- [x] more tests up to dotted uppercase
- [x] more tests up to scan by
- [x] more tests up to monographs
- [x] more tests up to math digraphs
- [x] more tests up to str digraphs
- [x] more tests up to list digraphs


Lyxal:
Okay so there's a little problem where it turns out that some of the tests don't actually pass because well um turns out there's bugs that need to be fixed.

We're gonna have to figure out how to add these tests + fix bugs at the same time

Lyxal (commented):
Okay so there's a little problem where it turns out that some of the tests don't actually pass because well um turns out there's bugs that need to be fixed.

We're gonna have to figure out how to add these tests + fix bugs at the same time

cgccuser:
This is good. We can merge these tests with admin access even if they don't pass, then get to the task of fixing bugs. Thanks for making these tests @cjquines

cjquines:
okay rl is getting in the way but i'll commit to finishing this commit by end of week

Lyxal (approved):
After fixing some of the test cases, LGTM

PR #1127, opened 2022-06-22
Fix dyadic modifiers when used monadically

Steffan153:
It simply appends the second element as an identity function.


Lyxal (approved):

cgccuser (approved):
Looks like this test didn't fail, LGTM

PR #1137, opened 2022-06-23
Fix str-str overload of `ġ`

cgccuser:
It wasn't handling the case where there wasn't a common suffix at all (the intersection of the sets of the suffixes of the two strings was empty, so no max could be found). I decided to get rid of that and write a more efficient function for finding the longest common suffix of either strings or lists, should we decide to add this behavior for lists too later on.


Lyxal (approved):
😎

That's me when I have the smartest code golfers all working together on the same project

Very cool

Issue #1119, opened 2022-06-22
Bugs that arose from the new test cases


Lyxal:
_Format: `element`: `test case that caused error (before fixing)` (comments)_

# Bugs
- [x] `•`: `[[1,2,3], [[4], [], [6]]] -> [[1], [], [2]]` (#1167)
- [x] `%`: `[5,0] -> 0` (#1118)
- [x] `/`: `["abcde",3] : ["a","b","c","de"]` (#1148)
- [x] `K`: `[-1] : [-1]` (#1133)
- [x] `i`: `[1357,[1,3]] : 35` (#1145)
- [x] `β`: `["43",5] : 23` (#1153)
- [x] `ġ`: `["asdf","fdsa"] : ""` (#1137)
- [x] `ŀ`: `["a", 1, 2] : "a\na"` (Numbers don't seem to be swapping properly) (#1141)
- [x] `ȯ`: `[123,1] : 23` (#1128) 
- [x] `ẏ`: `[-0.25] : [0, 1, 2, 3]` (Should actually be `[0, 1, 2, 3, 4]`) (#1139)
- [x] `ż`: `[-0.25] : [1,2,3,4]` (Should actually be `[1, 2, 3, 4, 5]`) (#1139)
- [x] `§`: `[123456] : "1\n2\n3\n4\n5\n6"` (#1163)
- [x] `¦`: `[""] : []` (#1143)
- [x] `Ḋ`: `[15, 0] : 0` (#1144)
- [x] `Ḣ`: `[-123] : 123` (#1154)
- [x] `Ŀ`: `[-120, [2], [5]] : [-150]` (#1134)
- [x] `Ṫ`: `[-1234.56] : -1234.5` (discussed in #1147)
- [x] `Ẇ`: `[-1231234.5, 3] : [12, 3, 12, 3, 4.5]` (#1155)
- [x] `℅`: `[""] : ""` (#1142)
- [x] `∩`: `[1234] : [[1], [2], [3], [4]]` (#1135)
- [x] `⊍`: `["12",12] : []` (#1157)
- [x] `ø.`: `[123, 4] : 41234` (#1156)
- [x] `Þj`: `[[]] : 1` (#1149)
- [x] `ÞẊ`: `["ab",1] : ["a","b"]` (#1151)
- [x] `ÞB`: _all tests it seems_ (#1161)
- [x] `Þǔ`: `[[]] : []` (#1160)
- [x] `Þ℅`: `[1] : 1` (#1146)
[...]

PR #1143, opened 2022-06-23
Improve the behaviour of `¦`

Lyxal:
Goes towards closing #1119


cgccuser (approved):

vyxal/elements.py
```diff
@@ -1051,6 +1051,10 @@ def cumulative_sum(lhs, ctx):
     """Element ¦
     (any) -> cumulative sum of a
     """
+    if lhs == "":
+        return []
```
cgccuser, line 1055:
```suggestion
        return ""
```
Maybe a string?

Steffan153 (approved):

chunkybanana:
@Lyxal Why the frickity frick did you think taking the length of a potentially infinite list was a good idea?

PR #1144, opened 2022-06-23
Improve the behaviour of `Ḋ`

Lyxal:
Goes towards closing #1119

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -2397,7 +2397,9 @@ def is_divisible(lhs, rhs, ctx):
     def helper(lhs, rhs):
         ts = vy_type(lhs, rhs)
         return {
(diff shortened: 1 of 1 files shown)


vyxal/elements.py
```diff
@@ -2390,7 +2390,9 @@ def is_divisible(lhs, rhs, ctx):
     def helper(lhs, rhs):
         ts = vy_type(lhs, rhs)
         return {
-            (NUMBER_TYPE, NUMBER_TYPE): lambda: int(lhs % rhs == 0),
+            (NUMBER_TYPE, NUMBER_TYPE): lambda: int(lhs % rhs == 0)
```
cgccuser, line 2393: If we were golfing I'd suggest `rhs and int(lhs ...)` but that'd make this unclear :(

cgccuser (approved):

Steffan153 (approved):

PR #1148, opened 2022-06-23
Fix behavior of `/` and `Þ÷` when splitting strings/lists

cgccuser:
I couldn't see how to get `wrap` to do exactly what we want, so I made a helper for it. Hope that's okay. Goes towards closing #1119


Steffan153 (approved):

Lyxal (approved):
It's probably time there was a dedicated function for it anyway

PR #1149, opened 2022-06-23
Make depth (`Þj`) work with empty children

cgccuser:
It was erroring because `max` was being passed an empty list.


Lyxal (approved):
Good

Steffan153 (approved):

PR #1151, opened 2022-06-23
Fix tests for `Ẋ` and `ÞẊ`

cgccuser:
`ÞẊ` now turns string arguments into lists of characters and turns numbers into ranges. The docs and tests for both now reflect this.


Lyxal (approved):
Looks gaming to me

Steffan153 (approved):

PR #1153, opened 2022-06-24
Fix from custom base

Steffan153:
Goes towards closing #1119.


vyxal/elements.py
```diff
@@ -1750,12 +1750,19 @@ def from_base(lhs, rhs, ctx):
...
     if ts == (str, str):
         return from_base_alphabet(lhs, rhs)
     elif ts[-1] == NUMBER_TYPE:
-        lhs = [chr(x) if type(x) is str else x for x in iterable(lhs)]
-        return from_base_digits(lhs, rhs)
+        if ts[0] == str:
+            return from_base(
+                lhs, (string.digits + string.ascii_lowercase)[:rhs], ctx
+            )
+        return from_base_digits(iterable(lhs), rhs)
```
Lyxal, line 1765:
```suggestion
        return from_base_digits(iterable(lhs, ctx=ctx), rhs)
```

Lyxal: Maybe add `ctx` here? Probably not needed, but doesn't hurt.

Lyxal (approved):

PR #1155, opened 2022-06-24
Fix split and keep delimiter

Steffan153:
Goes towards closing #1119.

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -4452,7 +4452,14 @@ def split_keep(lhs, rhs, ctx):
             if temp:
                 yield temp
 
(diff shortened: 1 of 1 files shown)


Lyxal (approved):

vyxal/elements.py
```diff
@@ -4429,20 +4429,29 @@ def gen():
...
 
-        return LazyList(gen())
+        return LazyList(
+            (
+                sympy.nsimplify("".join(map(str, x)), rational=True)
+                for x in gen()
+            )
+            if is_num
+            else gen()
+        )
```
cgccuser, line 4454:
```suggestion
        if is_num:
            return LazyList(
                sympy.nsimplify("".join(map(str, x)), rational=True)
                for x in gen()
            )
        else:
            return gen()
```

cgccuser (approved):

PR #1159, opened 2022-06-24
Fix matrix determinant

Steffan153:


Lyxal (approved):
I'mma gon have to trust you with that, because I don't know much about the math of matrices.

cgccuser (approved):

PR #1161, opened 2022-06-24
Fix random bits

Steffan153:


Lyxal (approved):
There's a 99% chance I wrote the original function so 10/10 for me doing a great job on remembering how to lambda. Thanks for fixing my (probable) stupidity lol

cgccuser (approved):

PR #1162, opened 2022-06-24
Fix the molding functions

Lyxal:
Goes towards closing #1119

### DIFF
--- a/vyxal/helpers.py
+++ b/vyxal/helpers.py
@@ -653,23 +653,29 @@ def mold(
     The content, molded into the shape.
     """
     final = []
-    original, content = itertools.tee(content)
+    original, content = map(LazyList, itertools.tee(content))
     for item in shape:
         temp = []
         if isinstance(item, str) or is_sympy(item):
             item = [item]
         for _ in item:
-            obj = next(content, None)
-            if obj is None:
-                content = itertools.tee(original)[1]
-                obj = next(content)
-            temp.append(obj)
+            if type(_) in [list, LazyList]:
+                temp = mold(content[: len(_)], _)
+                content = content[len(_) :]
+            else:
+                obj = next(content, None)
+                if obj is None:
+                    content = itertools.tee(original)[1]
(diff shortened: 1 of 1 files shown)


cgccuser (approved):
I'm too smol brain to understand this code but the tests for this command seem to pass so LGTM

Lyxal:
> you could copy this from jelly. it's just 9 lines [https://github.com/DennisMitchell/jellylanguage/blob/master/jelly/interpreter.py#L578-L586](https://github.com/DennisMitchell/jellylanguage/blob/master/jelly/interpreter.py?rgh-link-date=2022-06-24T14%3A31%3A14Z#L578-L586)

Mutability and lazylists say otherwise.

cgccuser:
Wasn't it already copied from Jelly and then modified?

Lyxal:
> Wasn't it already copied from Jelly and then modified?

That's right. In theory, it should still be the same(ish) algorithm just better suited for LazyLists and immutability.

vyxal/helpers.py
```diff
@@ -696,15 +702,25 @@ def mold_without_repeat(
...
-        if isinstance(item, (int, str)):
+        if isinstance(item, str) or is_sympy(item):
             item = [item]
         for _ in item:
-            obj = next(content, None)
-            if obj is None:
-                break
-            temp.append(obj)
+            if type(_) in [list, LazyList]:
+                temp = mold(content[: len(_)], _)
```
chunkybanana, line 709: Should this not be recursing mold_without_repeat?

vyxal/helpers.py
```diff
@@ -653,23 +653,29 @@ def mold(
     The content, molded into the shape.
     """
     final = []
-    original, content = itertools.tee(content)
+    original, content = map(LazyList, itertools.tee(content))
```
chunkybanana, line 656: This will only map at depth 1.

Lyxal:
Replaced by #1167

PR #1167, opened 2022-06-25
Fixed mold (I think?)

chunkybanana:
This should work without mutable lists. All tests work.


chunkybanana:
Should also work for infinite lists. Some small modifications may be required for it to work for modular indexing.

chunkybanana:
Btw, helps fix #1119

Lyxal (approved):
:+1:

cgccuser (approved):
Awesome job

Issue #1164, opened 2022-06-24
Ambiguities that arose from the test cases


Lyxal:
Round 2 of everyone's favourite game. This time, the failed tests are to do with decisions regarding things like typing or edge cases that weren't previously thought of. That means that some things can be ignored/left as-is.

- [x] `½`: `[""] -> ["", ""]` (what to do with empty string?) (#1168)
- [x] `F`: `["12345",["2","4","5"]] : "13"` (should lhs be converted to list if rhs is already list?) (#1169)
- [x] `O`: `[1213121, "1"] : 4` (should string be considered same as int?) (#1170)
- [x] `U`: `["1",1] : ["1",1]` <s>(should `"1"` and `1` be considered equal in this context?)</s> Someone forgot to format the test correctly.
- [x] `V`: `[234,"2",4] : 234` (should `"2"` and `2` be considered equal in this context?) (#1171)
- [x] `ḭ`: `["a",3] : ""` <s>(Why wouldn't it be `["a"]`?) (#1172)</s> Looks like it's valid after all
- [x] `Ȧ`: `[[], [1, 2], 2] : [2, 2]` (What even _should_ happen here? What does assigning to an empty list do?) (#1173)
- [x] `İ`: `[[1,2,3], [-1,0.1,1.9,"0"]] : [3,1,2,1]` (That's a tricky one. What if there's a string in the index list?) (#1298)
- [x] `Ṁ`: `[134,1,2] : 1234` (Should the result be a list or a number? What if it was `"a"` instead of `2`?)
- [x] `Ẋ`: `[12,34] : [13,14,23,24]` (Should those be numbers or lists?)
- [x] `Ẏ`: `[123,2] : 12` (Number or list?) (#1199)
- [x] `Ż`: `[123,3] : 23` (Same as above) (#1199)
- [x] `⌊`: `["-123"] : -123` (Should `-` be counted as part of the number?) (#1198)
- [x] `∪`: `["12",12] : ["1","2"]` (How should strings and numbers interplay here?) (#1299)
- [x] `Ǒ`: `[1.125, 2] : -3` and `[1.125, 3] : 2` <s>(What rules for divisibility did you use? /srs is there a way of determining decimal divisibility?)</s> I think integer order is fine here. (#1301)
- [x] `øḞ`: `["234",2,4] : "234"` <s>(Depends on how strings + numbers should interact)</s> current behaviour is fine. (#1312)
[...]

PR #1168, opened 2022-06-25
Improve the behaviour of `½`

Lyxal:
Goes towards closing #1164


vyxal/elements.py
```diff
@@ -1962,7 +1962,9 @@ def halve(lhs, ctx):
     ts = vy_type(lhs)
     return {
         NUMBER_TYPE: lambda: sympy.Rational(lhs, 2),
-        str: lambda: wrap(lhs, math.ceil(len(lhs) / 2), ctx=ctx),
+        str: lambda: wrap(lhs, math.ceil(len(lhs) / 2), ctx=ctx)
+        if len(lhs) > 1
+        else [lhs, ""],
     }.get(ts, lambda: vectorise(halve, lhs, ctx=ctx))()
```
cgccuser, line 1968: Do you want to make this whole thing just `return divide(lhs, 2)` lol

Lyxal: No, because it wouldn't work lol

Steffan153 (approved):

cgccuser:
I have to say, it feels weird that `½` isn't the same as `2 /`. We may want to rethink the behavior of `/`. I guess it can't be changed now because it'd be breaking, but we should take a look at it again for v3.

chunkybanana (approved):
`/` is still sus tho

PR #1172, opened 2022-06-25
Fix a test case for `ḭ`

Lyxal:
Goes towards closing #1164


cgccuser (commented):
The implementation also needs to be fixed, because this test case doesn't pass

cgccuser (approved):

Lyxal:
Looks like it's correct behaviour

Lyxal:
Not deleting just in case this ever gets discussed again

PR #1175, opened 2022-06-25
Death to this ÞR (Matrix Row Reduce)

### LINKED ISSUE #1150: Two different implementations of `ÞR`
There are currently two implementations of `ÞR`
1. Remove the last item of the cumulative sums of a list and prepend 0.
2. Reduce the rows of a by function b

I think the second is more general and worth keeping. The first is extremely specific.

(this was mentioned in the big bugs issue (test case `[[]] : [0]`))

cgccuser:
Closes #1150. Matrix Row Reduce is dead, long live Cumul Sums Sans Last Prepend Zero!

### DIFF
--- a/documents/knowledge/elements.md
+++ b/documents/knowledge/elements.md
@@ -3571,14 +3571,6 @@ Diagonal of a matrix
 
 - lst a: `diagonal(a)`
 -------------------------------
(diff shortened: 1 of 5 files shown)


Lyxal (approved):
`vƒ` amiright?

Steffan153 (approved):

Issue #1116, opened 2022-06-21
Online interpreter tooltips are slow on mobile


Lyxal:
So say you load the interpreter on your phone.

[This is what happens](https://youtube.com/shorts/WyE_BfvTkiM)

cjquines:
i think this is a problem with popper. on mobile, when you tap a button:

- the element gets focused, so popper starts computing the tooltip based on the current viewport,
- the keyboard appears, changing the viewport, so popper has to recompute,
- the input is focused, which makes most browsers scroll to it, so popper has to recompute.

i want to propose an alternate ui for tooltips on mobile: we disable popper entirely. pressing and holding on an element brings up a modal, which contains the tooltip content. there's a button to close the modal.

PR #1188, opened 2022-06-26
refactor frontend

cjquines:
in preparation for #1116, we:
- split the frontend js into several files
- remove jquery (it's 2022, seriously)
- defer scripts to decrease blocking loading time


chunkybanana (approved):
***DIE DIE DIE JQUERY***

Lyxal (approved):

PR #1194, opened 2022-06-26
Fix fixed-point

### LINKED ISSUE #1193: `fixed_point` should not use `simplify`
This causes integers to turn into `float`s, and since `vy_type` asserts that it is not a float, this causes all kinds of errors.

Steffan153:
Fixes #1193


chunkybanana (approved):
Thought this was sympy for a moment

Lyxal (approved):

PR #1197, opened 2022-06-28
There is no such thing as `types.NoneType`

Lyxal:


vyxal/helpers.py
```diff
@@ -893,7 +893,7 @@ def simplify(value: Any) -> Union[int, float, str, list]:
     Simplify values.
     Turns sympy values into floats, including sympy values in lists
     """
-    if isinstance(value, (int, float, str, types.NoneType)):
+    if isinstance(value, (int, float, str)) or value is None:
```
cgccuser, line 896:
```suggestion
    if value is None or isinstance(value, (int, float, str)):
```
Doesn't make a huge difference but still

cgccuser (approved):

Steffan153:
There is such a thing as `types.NoneType`:
image

Lyxal:
`types.NoneType` was introduced in Python 3.10. PythonAnywhere uses 3.8 or 3.7

PR #1198, opened 2022-06-28
Make `⌊` account for decimals and negatives

Lyxal:
Goes towards closing #1164

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -5659,17 +5659,6 @@ def vy_floor(lhs, ctx):
     (str) -> integer part of a
     """
 
-    def vy_floor_str_helper(item):
-        temp = ""
-        for char in item:
-            if char == "-" and temp == "":
-                temp += char
-            elif char.isdigit():
-                temp += char
-            elif char == "." and "." not in temp:
-                temp += char
-        return sympy.nsimplify(temp)
-
     ts = vy_type(lhs)
     return {
         (NUMBER_TYPE): lambda: lhs.real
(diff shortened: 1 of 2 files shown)


vyxal/elements.py
```diff
@@ -5658,6 +5658,20 @@ def vy_floor(lhs, ctx):
     (num) -> floor(a)
     (str) -> integer part of a
     """
+
+    def vy_floor_str_helper(item):
```
cgccuser, line 5662: Wanna move this to helpers.py?

vyxal/elements.py
```diff
@@ -5658,6 +5658,20 @@ def vy_floor(lhs, ctx):
...
+        temp = ""
+        for char in item:
+            if char == "-" and temp == "":
+                temp += char
+            elif char.isdigit():
+                temp += char
+            elif char == "." and temp.count(".") == 0:
```
cgccuser, line 5669:
```suggestion
            elif char == "." and "." not in temp:
```

vyxal/elements.py
```diff
@@ -5658,6 +5658,20 @@ def vy_floor(lhs, ctx):
...
+                temp += char
+            else:
+                continue
```
cgccuser, line 5672:
```suggestion
```
No need to continue if you're already at the end.

vyxal/elements.py
cgccuser: Might want to make sure you don't get a string like `-.` by adding a zero.

cgccuser (changes requested):

cgccuser (approved):

Steffan153 (approved):

PR #1207, opened 2022-06-29
Add element to get binary string

### LINKED ISSUE #1179: Add binary string digraph
When trying to create binary strings out of negative numbers, it's just a pain. Also the fact that join by nothing does not vectorize makes this very commonly 3 bytes anyway.

Steffan153:
Closes #1179


Lyxal (approved):
Maybe it could be an overload of product?

Steffan153:
Good idea, I'll do that later.

Steffan153:
actually no, because then it couldn't vectorize.

chunkybanana (approved):

PR #1208, opened 2022-06-29
Add test cases

### LINKED ISSUE #1177: Add "test cases" to online interpreter
It would basically just run each test case and output it and its result. I would say a format like this:

Test-cases will be separated by a newline. Empty lines are ignored. A test-case would be a comma-separated list of inputs. Spaces around commas are optional.
Optionally, each test case can have an expected output at the end, by adding `=>` or `->` (optional spaces around it) and the expected output, which will make it verify the output.
If a test case ends with a comma, `=>`, or `->`, it should wrap to the next line.

Steffan153:
Closes #1177

### DIFF
--- a/vyxal/main.py
+++ b/vyxal/main.py
@@ -9,6 +9,7 @@ import traceback
 import types
 
 import vyxal.encoding
+import itertools
 from vyxal.context import Context
 from vyxal.elements import *
 from vyxal.helpers import *
@@ -60,6 +61,7 @@ FLAG_STRING = """ALL flags should be used as is (no '-' prefix)
     ?    If there is empty input, treat it as 0 instead of empty string.
     2    Make the default arity of lambdas 2
     3    Make the default arity of lambdas 3
+    A    Run test cases on all inputs
     …    Limit list output to the first 100 items of that list
     5    Make the interpreter timeout after 5 seconds (online interpreter only)
     b    Make the interpreter timeout after 15 seconds (online interpreter only)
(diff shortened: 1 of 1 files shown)


Lyxal (approved):

vyxal/main.py
```diff
@@ -9,6 +9,7 @@
 import types
 
 import vyxal.encoding
+import itertools
```
cgccuser, line 12: What's this for?

vyxal/main.py
```diff
@@ -84,6 +86,23 @@ def execute_vyxal(file_name, flags, inputs, output_var=None, online_mode=False):
...
+        for inp in inputs:
+            inps = ast.literal_eval(inp)
+            if isinstance(inps, tuple):
+                inps = list(inps)
+            else:
+                inps = [inps]
+            print(inp, end=" => ")
+            execute_vyxal(
+                file_name,
+                "".join(x for x in flags if x != "A"),
```
cgccuser, line 99:
```suggestion
                flags.replace("A", ""),
```

cgccuser (changes requested):

vyxal/main.py
Steffan153: Used it in a previous version of code, but forgot to remove it
```suggestion
```

cgccuser (approved):
LGTM, nice feature

Steffan153:
I didn't add the feature so you could put an expected result because that means manually lexing the string so you know where the arrow and commas actually are and stuff, so it would be super complicated.

cgccuser:
Makes sense yeah

PR #1216, opened 2022-06-29
Make `gen_from_fn` (`Ḟ`) work with LazyLists

cgccuser:
It was trying to call the `append` method, which only works for lists.


Steffan153 (approved):
Why is this still draft?

cgccuser:
@Steffan153 Because I went to eat

chunkybanana (approved):

PR #1227, opened 2022-06-30
cleaned up vectorise

chunkybanana:
The code's a lot cleaner and less repetitive now.

Everything in test_complex.py passes, and it seems to work overall, but I may have broken some other stuff. Review with caution.

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -5325,10 +5325,7 @@ def unwrap(lhs, ctx):
 
 def vectorise(
     function,
-    lhs,
-    rhs=None,
-    other=None,
-    fourth=None,
+    *args,
     explicit=False,
     ctx: Context = None,
 ):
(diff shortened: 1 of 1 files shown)


vyxal/elements.py
```diff
@@ -5338,199 +5335,50 @@ def vectorise(
...
-                )
-            ),
-        }
-
-        if explicit:
+    if explicit:
+        # explicit vectorisation
+        # Algorithm: If there are no lists, vectorise over lhs.
+        # Else, vectorise over the first list.
+        ts = [primitive_type(x) for x in args]
```
Lyxal, line 5342: Rip vectorisation over infinite lists

Lyxal (approved):

vyxal/elements.py
```diff
@@ -5338,199 +5335,50 @@ def vectorise(
...
-            lhs = iterable(lhs, ctx=ctx)
+        # regular vectorisation
+        @lazylist
+        def vectorise_helper():
+            # Like zip, but repeats scalars
+            @lazylist
+            def row_helper(index):
+                for item in args:
+                    if primitive_type(item) is list:
+                        yield item[index]
```
Steffan153, line 5365: This fails when doing something like `[1,2,3,4] + [5,6,7]`.

Steffan153 (changes requested):
Whoops forgot to request changes.

chunkybanana:
Well, frick.

vyxal/elements.py
chunkybanana: Why? This is iterating over each argument to vectorise. Are you referring to another line?

Lyxal: Oh.


I thought it was going over items in the list that's passed in

Lyxal (approved):
Gg

Steffan153 (approved):
lgtm

chunkybanana:
If this fails, blame me.

PR #1229, opened 2022-07-01
Make divide sensible

### LINKED ISSUE #1212: Make `/` always split into n parts, without an extra part
So `` `hello` 2/ `` should not be `["he", "ll", "o"]`, but `["hel", "lo"]`. And `` `h` 2/ `` should be `["h", ""]`, not `["", "", "h"]`.

chunkybanana:
Fixes #1212


vyxal/helpers.py
```diff
@@ -42,17 +42,6 @@ def case_of(value: str) -> int:
     return -1
 
 
-@lazylist
-def chop(it: VyIterable, n: int) -> LazyList:
```
Steffan153, line 46: Why remove from helpers? It's the definition of a helper.

chunkybanana: Because at one point it needed something from elements. It doesn't now.

Lyxal (changes requested):
Don't forget to update the description of the element in the YAML

Lyxal (approved):

vyxal/helpers.py
```diff
@@ -44,13 +44,13 @@ def case_of(value: str) -> int:
...
     chunk_len = len(it) // n
-
+    left_over = len(it) % n
     for i in range(n):
-        yield it[i * chunk_len : (i + 1) * chunk_len]
-    if n * chunk_len < len(it):
-        yield it[n * chunk_len :]
+        length = chunk_len + (1 if left_over > i else 0)
+        yield it[:length]
+        it = it[length:]
```
cgccuser, line 53: If you want to golf it a bit, something like [this](https://tio.run/##jY5BCsIwEEX3PcVfJhopaVWk4ElKEWlTO1imIcZFTx/HglBwoX8zMP/NY/wch4nLkw8pda5HO0xeUTRgXWWQtMOT75fRMc5QMqTU2IKxg9XIc/CC9VMAgRjhyjenPtfvzOTGDhRrwmalq6BIRCJZbZss84E4qpEeUS3P1I3BXmv9XViDwqCU2uAglP2LKn5TBkcBSwFTegE) would work too, I think:
```python
chunk_len = (len(it) + n - 1) // n
for i in range(n):
    yield it[i * chunk_len : (i + 1) * chunk_len]
```

chunkybanana: I don't think it would, as this builtin divides into variable length chunks.

Steffan153 (approved):

PR #1231, opened 2022-07-01
Fix bugs

chunkybanana:
Many bugs fix


chunkybanana:
Specifically, fixed all (I think) the non-ambiguous tests.

Lyxal (approved):
I didn't think I missed any tests in my list but I guess I must have

documents/knowledge/elements.yaml
```diff
@@ -1024,7 +1024,7 @@
       - '["abc",[1,2,3]] : [["a",1],["b",2],["c",3]]'
       - '["abc","d"] : [["a","d"], ["b", 0], ["c", 0]]'
       - '["d","abc"] : [["d", "a"], [0, "b"], [0, "c"]]'
-      - '[123,"45"] : [[1,"4"],[2,"5"]]'
+      - '[123,"45"] : [[1,"4"],[2,"5"],[3,0]]' # Filling with 0s, consistent with the rest
```
cgccuser, line 1027: These comments explaining the rationale for the test are helpful +1

vyxal/elements.py
```diff
@@ -1016,8 +1020,7 @@ def cumul_sum_sans_last_prepend_zero(lhs, ctx):
     """Element ÞR
     Remove the last item of the cumulative sums of a list and prepend 0.
     """
-
-    return prepend(scanl(add, iterable(lhs, ctx=ctx), ctx)[:-1], 0, ctx)
+    return prepend(cumulative_sum(lhs[:-1], ctx=ctx), 0, ctx)
```
cgccuser, line 1023: Btw, I think the reason `scanl` wasn't working is because you need to give it the initial value `0`

vyxal/elements.py
```diff
@@ -4473,7 +4477,7 @@ def shuffle(lhs, ctx):
     """Element Þ℅
     (lst) -> Return a random permutation of a
     """
-    temp = list(deep_copy(lhs))
+    temp = list(deep_copy(iterable(lhs)))
```
cgccuser, line 4480:
```suggestion
    temp = list(deep_copy(iterable(lhs, ctx=ctx)))
```
Just in case

cgccuser (approved):
Great work

PR #1232, opened 2022-07-01
Make `Þİ`  work with swapped args

chunkybanana:
Fixes #1128


Lyxal (approved):
Type reversibility 👍

cgccuser (approved):
I'm starting to wonder if we should have something to automatically swap stuff for us. Maybe a decorator?

PR #1233, opened 2022-07-01
Fix conditional execute

### LINKED ISSUE #1182: Conditional execute is still broken
https://chat.stackexchange.com/transcript/message/61449880#61449880

(created by Steffan [here](https://chat.stackexchange.com/transcript/message/61449901))

chunkybanana:
By making the parser parse it as an if statement.  Fixes #1182


Lyxal (approved):
it'll do for now

cgccuser (approved):
LGTM. It's mildly worrying we can't write it like the other modifiers. Should we change how modifiers work?

PR #1234, opened 2022-07-02
Make tooltips unselectable

### LINKED ISSUE #1217: Keys / tooltips can be selected
A simple `user-select: none` with the appropriate prefixes should fix this.

chunkybanana:
Fixes #1217

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -78,6 +78,10 @@ summary {
   color: white;
   width: 1em;
   text-align: center;
+  user-select: none; /* Frick compatibility */
+  -webkit-user-select: none;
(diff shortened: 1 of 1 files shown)


Steffan153 (approved):
Honestly it's 2022 and you don't really need -webkit and -moz, and definitely not -ms, that's IE only. But whatever.

cjquines:
the big demographic you'll miss by not having prefixes is [safari on ios](https://caniuse.com/?search=user-select), and therefore any device on ios, but i think the fact we prevent the default long press on mobile anyway means they shouldn't be able to select in the first place

cjquines (approved):

PR #1236, opened 2022-07-02
Fix ¨i modifier with the stack and stuff

Steffan153:


chunkybanana (approved):
As with the other one, this is probably a temporary fix.

cgccuser (approved):

PR #1238, opened 2022-07-02
Add drop while overload

### LINKED ISSUE #1218: add drop while element overload
(created by Steffan [here](https://chat.stackexchange.com/transcript/message/61479479))

Steffan153:
Closes #1218

### DIFF
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -319,7 +319,7 @@
       str-num: choose b random items from a
       str-str: are the set of characters in the strings the same?
       any-fun: remove each item x from the beginning of a until b(x) returns false
-      fun-any: remove items from the beginning of b until a(x) returns false
+      fun-any: remove each item x from the beginning of b until a(x) returns false
   vectorise: true
   tests:
       - "[5,3] : 10"
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -318,7 +318,7 @@
       num-str: choose a random items from b
       str-num: choose b random items from a
       str-str: are the set of characters in the strings the same?
-      any-fun: remove items from the beginning of a until b(x) returns false
+      any-fun: remove each item x from the beginning of a until b(x) returns false
       fun-any: remove items from the beginning of b until a(x) returns false
   vectorise: true
   tests:
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -310,14 +310,16 @@
   description: Used for list-related digraphs
 
 - element: "ƈ"
-  name: Choose / random choice / set same
-  description: Binomial coefficient / choose a random items from b / same except duplicates
+  name: Choose / random choice / set same / drop while
+  description: Binomial coefficient / choose a random items from b / same except duplicates / drop while
   arity: 2
   overloads:
       num-num: a choose b (binomial coefficient)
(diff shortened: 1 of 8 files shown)


documents/knowledge/elements.yaml
```diff
@@ -310,14 +310,16 @@
...
-  description: Binomial coefficient / choose a random items from b / same except duplicates
+  name: Choose / random choice / set same / drop while
+  description: Binomial coefficient / choose a random items from b / same except duplicates / drop while
   arity: 2
   overloads:
       num-num: a choose b (binomial coefficient)
       num-str: choose a random items from b
       str-num: choose b random items from a
       str-str: are the set of characters in the strings the same?
+      any-fun: remove items from the beginning of a until b(x) returns false
```
chunkybanana, line 321: What is `x`? The whole array? The dropped item?

vyxal/elements.py
```diff
@@ -3184,6 +3184,17 @@ def n_choose_r(lhs, rhs, ctx):
     (str, str) -> does a have the same characters as b
     """
     ts = vy_type(lhs, rhs)
+
+    @lazylist
+    def drop_while(vec, fun, ctx):
```
chunkybanana, line 3189: This should probably go in helpers.

documents/knowledge/elements.yaml
cgccuser: Each element of the array, I guess

cgccuser: How about
```suggestion
      any-fun: remove each item x from the beginning of a until b(x) returns false
```

cgccuser (changes requested):
What emanresu said

documents/knowledge/elements.yaml
```diff
@@ -310,14 +310,16 @@
...
+  name: Choose / random choice / set same / drop while
+  description: Binomial coefficient / choose a random items from b / same except duplicates / drop while
   arity: 2
   overloads:
       num-num: a choose b (binomial coefficient)
       num-str: choose a random items from b
       str-num: choose b random items from a
       str-str: are the set of characters in the strings the same?
+      any-fun: remove each item x from the beginning of a until b(x) returns false
+      fun-any: remove items from the beginning of b until a(x) returns false
```
cgccuser, line 322:
```suggestion
      fun-any: remove each item x from the beginning of b until a(x) returns false
```

vyxal/elements.py
```diff
@@ -3193,6 +3194,8 @@ def n_choose_r(lhs, rhs, ctx):
             random.choice(lhs) for _ in range(abs(int(rhs)))
         ],
         (str, str): lambda: int(set(lhs) == set(rhs)),
+        (types.FunctionType, ts[1]): lambda: n_choose_r(rhs, lhs, ctx),
```
cgccuser, line 3197: Why is this one `n_choose_r` and not `drop_while`?

cgccuser (changes requested):

vyxal/elements.py
Steffan153: `n_choose_r` is the parent function, so it doesn't really matter which, but I'll switch that around.

```suggestion
        (types.FunctionType, ts[1]): lambda: drop_while(rhs, lhs, ctx),
```

cgccuser (approved):
LGTM

chunkybanana (approved):

PR #1239, opened 2022-07-02
Actually fix A flag

Steffan153:
I actually tested this with the flask_app.py and it works.


cgccuser (approved):
Third time's the charm

chunkybanana (approved):
Woohoo!

PR #1245, opened 2022-07-03
Make Ǒ check for 0 and 1

### LINKED ISSUE #1244: `Ǒ` multiplicity loops infinitely with abs(rhs)==1
When `Ǒ` was [first implemented](https://github.com/Vyxal/Vyxal/commit/b02567f02f62a994ef985974e3948d00341d8fe3) in v2.0.0, it had a check for this:

```
    if types == (Number, Number):
        if rhs == 0 or abs(rhs) == 1:
            return "Infinite"
        elif lhs == 0:
            return 0
        temp, remainder = lhs, 0
        count = 0
        while True:
            temp, remainder = divmod(temp, rhs)
            if remainder:
                break
            count += 1
        return count
```

It is rather odd to me that it used a string literal for infinity instead of, say, `math.inf`, but it worked: [Try it Online!](https://lyxal.pythonanywhere.com?flags=&code=1%C7%91&inputs=&header=&footer=)

But when it was [reimplemented](https://github.com/Vyxal/Vyxal/commit/4f50082e0c1742100f9ff1f3e69aceb5f7582ad2) in v2.6.0, there was no such check:

```
    if ts == (NUMBER_TYPE, NUMBER_TYPE):
        count = 0
        while lhs % rhs == 0:
            lhs /= rhs
            count += 1
        return count
```

and as a result it loops infinitely: [Try it Online!](https://vyxal.pythonanywhere.com/#WyIiLCIiLCIxx5EiLCIiLCIiXQ==)

cgccuser:
Closes #1244

For `lhs == 0`, returns 0. For `rhs == 0`, returns `math.inf`. For `abs(rhs) == 1`, returns `abs(lhs)`.


Steffan153 (changes requested):
We don't use `math.inf` in Vyxal. It complicates things, so we replace it with `0`. If `inf` is ever encountered, it's a bug

vyxal/elements.py
```diff
@@ -3076,8 +3076,14 @@ def multiplicity(lhs, rhs, ctx):
     (any, fun) -> Index of the first truthy item in a under b(x)
     """
 
-    ts = vy_type(lhs, rhs, simple=True)
+    ts = vy_type(lhs, rhs)
     if ts == (NUMBER_TYPE, NUMBER_TYPE):
+        if lhs == 0:
+            return 0
+        if rhs == 0:
+            return math.inf
```
cgccuser, line 3084:
```suggestion
        if lhs == 0 or rhs == 0:
            return 0
```

chunkybanana (approved):

Steffan153 (approved):
why is this still draft?

PR #1249, opened 2022-07-05
Fuzzy search

cjquines:
see title

unfortunately had to switch from preact to react in order to use react-popper, but at least the tooltips are more performant now


cjquines:
god does our precommit not use 3.9+? sigh

chunkybanana (approved):
This looks like you know what you're doing, so have an approval.

cgccuser (commented):
Could you include a screenshot/video of what this looks like for those of us who are too lazy to run it ourselves and see?

PR #1257, opened 2022-07-06
Add buttons to insert digraphs in tooltips

Steffan153:


cjquines (approved):
had to check it didnt break mobile

chunkybanana (approved):

PR #1274, opened 2022-07-09
Preserve infinite lists state

### LINKED ISSUE #670: Infinite lists need to be marked as so
Builtins that take and return infinite lists need to mark their output as infinite. Currently, strip and possibly others have different behaviour on finite and infinite lists.

### LINKED ISSUE #1273: `J`/`p` do not preserve the state that a list is infinite

Steffan153:
Fixes #1273 and fixes #670. Whoa diff be much big


Steffan153:
A lot of tests failing so marked as draft

chunkybanana (approved):
Nice, this is a lot simpler than I had in mind. There might be a couple of edgecases though...

Lyxal (approved):
looks good to me

PR #1281, opened 2022-07-11
Fix linting issues with nonlocals

cgccuser:
Lifted iterablification of nonlocal variables up in `multiset_intersection` and a few other places. Added a `nonlocal rhs` to the `gen` function inside `multiset_intersection` because `rhs` has to be mutated inside it.

Also modified the linter workflow to exclude `build`, `dist`, and `.git`

### DIFF
--- a/.github/workflows/linter.yaml
+++ b/.github/workflows/linter.yaml
@@ -29,6 +29,8 @@ jobs:
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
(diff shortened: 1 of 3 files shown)


Steffan153 (approved):
I did the same thing on a few other elements as well tho, so you might check them too. Just Ctrl+F for `nonlocal`

Edit: powerset, and I think something in helpers.py too. (Edit: several in helpers.py)

cgccuser:
Ah okay. I'll add them to this PR later. I see now the linter is complaining about another one of those.

Lyxal (approved):

PR #1290, opened 2022-07-14
Added letter-to-number

chunkybanana:
Does #1286


Lyxal (changes requested):
What if I give it a string of length more than 1? How about length 0?

chunkybanana:
Will do

Lyxal (approved):

Steffan153 (changes requested):
This won't work on uppercase letters.

PR #1291, opened 2022-07-14
Add "truthy under" overload to T updot

### LINKED ISSUE #1147: Need number overload for `Ṫ` (Tail Remove)
Currently, `Ṫ` has no overload for numbers, so it just removes the last character of the stringified version of the number and converts it to an integer. This is somewhat useful for integers, since it's basically integer division by 10 for them, but I can't see too many use cases for decimals. I think there needs to be a different overload for numbers. Any suggestions?

Lyxal:
Closes #1147

### DIFF
--- a/documents/knowledge/elements.md
+++ b/documents/knowledge/elements.md
@@ -1599,12 +1599,13 @@ Reverse a value
 Sum of each item in a list
 
 -------------------------------
-## `` Ṫ `` (Tail Remove)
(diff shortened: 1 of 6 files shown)


chunkybanana (approved):

Steffan153 (approved):
Is this seriously useful

chunkybanana (approved):

PR #1298, opened 2022-07-14
Improve the behaviour of `İ`

lyxal:
Goes towards closing #1164


vyxal/helpers.py
```diff
@@ -815,6 +815,7 @@ def gen():
 def primitive_type(item: Any) -> Union[str, type]:
     """Turns int/Rational/str into 'Scalar' and everything else
     into list"""
+    print(item, is_sympy(item))
```
Steffan153, line 818: Umm you might want to remove that now
```suggestion
```

Steffan153 (approved):

PR #1308, opened 2022-07-14
Allow overriding inputs

### LINKED ISSUE #1306: Add digraph element to replace the input list with the list at the top of the stack
This would be very useful to put in headers/footers of test harnesses. Input acts differently than the stack (it's not popped, and it responds to `⁰` `¹` `?` `□`), so emulation of input using the stack (by pushing multiple copies of the emulated input) only can work for a subset of Vyxal programs.

Even if this were implemented as a trigraph, it'd be fine, as byte counts don't matter in headers/footers.

Steffan153:
Closes #1306

It's *that* simple? It's that simple.

### DIFF
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -6097,6 +6097,16 @@
   description: Read from STDIN, even if there are arguments
   arity: 0
 
+- element: "¨S"
+  name: Override Inputs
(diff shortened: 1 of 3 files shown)


Lyxal (approved):
Hey look I did something modular for once! lgtm

PR #1312, opened 2022-07-15
Change some tests for `øḞ`

Lyxal:
Goes towards closing #1164

Okay so I think that given things like `O` consider number and string the same, that should be precedence unless types should logically be different


chunkybanana (approved):
lgbt

Steffan153 (approved):

PR #1315, opened 2022-07-15
Make ↔ work on infinite lists differently

chunkybanana:
I'm an idiot, turns out the behaviour of `↔` is decidedly not set intersection, although it's close.This changes `↔`'s behaviour when given infinite lists.

Also, moved `vy_type` to helpers.

Also, turns out `deep_copy` didn't preserve infinite state.


Lyxal (approved):
One day we'll have to change deep copy and more to account for infinite status

cgccuser (approved):
LGTM. Might want to update the description to say set intersection too (just feels a bit clearer to me)

PR #1323, opened 2022-07-16
Allow tab string compress

Steffan153:
Already? Already.

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -872,7 +872,7 @@ function initCodeMirror() {
                         return;
                     }
                 }
-                const str = line.match(/`([a-z ]|\\`)+`$/)?.[0]?.slice(1, -1);
+                const str = line.match(/`[a-z ]+`$/)?.[0]?.slice(1, -1);
                 if (str) {
                     let r = 0n;
                     for (const c of str)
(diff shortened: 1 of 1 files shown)


static/main.js
```diff
@@ -869,7 +869,24 @@ function initCodeMirror() {
                     compressed = '»' + compressed + '»';
                     if (compressed.length <= num.length) {
                         cm.replaceRange(compressed, { line: cur.line, ch: cur.ch - num.length }, { line: cur.line, ch: cur.ch });
-                        // Suggested by Copilot, again. Works.
+                        return;
+                    }
+                }
+                const str = line.match(/`([a-z ]|\\`)+`$/)?.[0]?.slice(1, -1);
```
chunkybanana, line 875: Why are we matching backticks within strings when the strings we're compressing can't contain backticks?

Steffan153: smth I forgot to change

```suggestion
                const str = line.match(/`[a-z ]+`$/)?.[0]?.slice(1, -1);
```

Lyxal (approved):
When dictionary compression?

chunkybanana (approved):

PR #1325, opened 2022-07-16
fix while loops

### LINKED ISSUE #1324: While loops push two values to the context var stack, but only pop one
[Try it Online!](https://vyxal.pythonanywhere.com/#WyJjIiwiIiwiYGFiY2RgKHtYfW7igrQiLCIiLCIiXQ==)

Steffan153:
Closes #1324


chunkybanana (approved):
Huh.

vyxal/transpile.py
```diff
@@ -277,13 +277,16 @@ def transpile_structure(
             transpile_ast(struct.condition, indent, options=options)
             + indent_str("condition = pop(stack, 1, ctx=ctx)", indent)
             + indent_str("counter = ctx.range_start", indent)
+            + indent_str("temp_len = len(ctx.context_values)", indent)
```
chunkybanana, line 280: Nested while loops may break this

chunkybanana (changes requested):

Lyxal:
Why not just pop the second context item it pushes? Or why not remove one of the context items being pushed?

Steffan153:
> Why not just pop the second context item it pushes? Or why not remove one of the context items being pushed?

The condition being pushed was not needed, so I removed that. The rest of the code was because if `break` is used, then the context var is never popped.

vyxal/transpile.py
Steffan153: Good catch, fixed.

chunkybanana (approved):
LGTM

Lyxal (approved):

PR #1360, opened 2022-07-26
When generating a CGCC post template, don't show some flags

### LINKED ISSUE #1358: Make the online interpreter not put certain flags in generated submissions
Specifically, flags like `5` / `B` / `T` (timeout), `A` (testcases), `…` (limiting length) and `P` (list format)

Steffan153:
Closes #1358

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -608,16 +608,17 @@ function generateURL() {
 
 // onclick event listener for sharing buttons
 function shareOptions(shareType) {
-    var code = e_code.doc.getValue()
-    var url = generateURL()
-    var flags = document.getElementById("flag").value
-    var flag_appendage = ","
-    if (flags) {
-        flag_appendage = " `" + flags + "`,"
+    const code = e_code.doc.getValue()
+    const url = generateURL()
+    const flags = document.getElementById("flag").value
+    let flagAppendage = ","
+    const flagsThatMatter = flags.replace(/[5bBTAP…]/g, "");
+    if (flagsThatMatter) {
+        flagAppendage = " `" + flagsThatMatter + "`,"
     }
-    var output = ""
-    var utfable = [...code].every(x => (codepage + ' ' + '\n').includes(x))
-    var len = utfable ? code.length : new Blob([code]).size
+    let output = ""
+    const utfable = [...code].every(x => (codepage + ' ' + '\n').includes(x))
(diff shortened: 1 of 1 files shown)


Lyxal (changes requested):
The `a` and `S (updot)` flags are also flags that don't require stating by default - one is for when you just want multiple lines as a list and one is for when you want everything to be taken as a string anyway, but you don't know if people testing the program will know to wrap in quotes.

Steffan153:
That's not fully true - some questions have stupid I/O that require newline separated input. In that case, the `a` flag probably should be specified

Lyxal:
Well if you're using it as an advantage, you can include it manually. It's really for cases where the challenge has given test data as a bunch of stuff separated by newlines

Steffan153:
Then you might as well just put `□` in the header or use the `A` flag. But, fine.

static/main.js
```diff
@@ -608,16 +608,17 @@ function generateURL() {
...
-    var url = generateURL()
-    var flags = document.getElementById("flag").value
-    var flag_appendage = ","
-    if (flags) {
-        flag_appendage = " `" + flags + "`,"
+    const code = e_code.doc.getValue()
+    const url = generateURL()
+    const flags = document.getElementById("flag").value
+    let flagAppendage = ","
+    const flagsThatMatter = flags.replace(/[5bBTAP…]/g, "");
```
Steffan153, line 615:
```suggestion
    const flagsThatMatter = flags.replace(/[5bBTAP…aṠ]/g, "");
```

Lyxal (approved):

Lyxal:
From experience, those 2 flags have only ever been used for convenience for test cases only challenges with relaxed io in the first place.

Steffan153:
See here where `a` flag was useful: https://codegolf.stackexchange.com/questions/1775/additive-persistence/250227#250227

Lyxal:
Well that's a case where it should be manually included. Every time I've used them, I've removed them from the header, and I suspect others have too

Lyxal:
> See here where `a` flag was useful: https://codegolf.stackexchange.com/questions/1775/additive-persistence/250227#250227

The a flag doesn't need to be included - people are just taking a normal list of integers as input anyway there.

chunkybanana (approved):

PR #1399, opened 2022-08-13
Make `İ` make complex numbers

### LINKED ISSUE #1390: Add element to get complex number from real and imag part
So take `x` and `y` and return `x + 1j * y`.

Lyxal:
Closes #1390


Steffan153 (approved):
Don't know if we need this 1 byte but ok

Lyxal:
I wanted it to be 1 byte because normal complex syntax is 1 byte, and the num-num overload of I updot was useless anyway.

cgccuser (approved):
LGTM, I like that this is an element and not part of the parser

PR #1410, opened 2022-08-31
Make ȯ work with integers that would give a slice starting with 0

Lyxal:


cgccuser (approved):

cgccuser:
I'd ask for tests or add myself but feeling lazy and let's be honest I never test things myself

PR #1414, opened 2022-09-01
Add partition after truthy indices and partition before indices elements

### LINKED ISSUE #1396: Add element to split a on truthy indices in b
05AB1E `Å¡`, Jelly `œṗ`. Also, split a on indices in b, Jelly `œṖ`.

Lyxal:
Closes #1396


Steffan153 (approved):

Steffan153 (commented):
actually why do you have these as helper functions? no one's gonna use them as helpers

Steffan153 (changes requested):
^ and also you ned to use `@lazylist_from` to make sure infinite list state is preserved.

Lyxal:
> actually why do you have these as helper functions? no one's gonna use them as helpers

same could be said about half the functions in the file.

They are in the helper file just in case they need to be used for other split on functions - Jelly's are used for two different purposes, so I figured there could be more use cases.

Steffan153 (approved):
fine, but if we need them again, we could just move them to helpers when we actually need it.

PyGamer0 (approved):

PR #1423, opened 2022-09-04
Add several "all powers" built-ins

### LINKED ISSUE #1419: All powers of a number
Like `¨*` but with `e` if that makes sense. Also, common powers like 2, 3, 10 should have their own digraph

(created by lyxal [here](https://chat.stackexchange.com/transcript/message/61925167))

Lyxal:
Closes #1419

### DIFF
--- a/documents/knowledge/elements.md
+++ b/documents/knowledge/elements.md
@@ -3920,6 +3920,22 @@ Distance matrix of undirected graph
 
 - lst a: `distance matrix of an undirected graph (where a = [[i, j] for each edge i to j])`
 -------------------------------
+## `` ÞṖ `` (Split Before Indices)
+
+Split a list before indices in another list
+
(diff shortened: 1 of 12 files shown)


Steffan153 (commented):
Not sure if all powers of 3 is useful but whatever. Why isn't this under list digraphs?

Lyxal:
> Not sure if all powers of 3 is useful but whatever. Why isn't this under list digraphs?

I placed it in the miscellaneous digraph space for consistency with all multiples.

Do you think there's a power more suitable than 3? If so, it can be changed to that. I just thought 3 was a sensible one to have.

Steffan153 (approved):
No, I think `3` just isn't needed

Lyxal:
Fair enough. I'll remove it a little bit later

PR #1424, opened 2022-09-04
Add `)` as lambda to newline, and make only `}` close all structures

### LINKED ISSUE #1342: Add lambda to newline
We can use `)`, so if no `(` is opened, then it does lam to newline

Lyxal:
Closes #1342

### DIFF
--- a/tests/test_complex.py
+++ b/tests/test_complex.py
@@ -882,7 +882,7 @@ def test_context_var_while():
 
 
 def test_close_multiple_structs():
-    stack = run_vyxal("λ0[2|3(n⅛;†¾")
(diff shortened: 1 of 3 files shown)


Steffan153 (changes requested):
Keep the rest of the closing characters functioning as before.

Steffan153 (changes requested):
This still isn't how it was before - `}];` should not necessarily close *all* structures, so `[{(...}` will only close the for-loop.

chunkybanana:
^

Steffan153 (approved):
👍

PR #1429, opened 2022-09-05
`R` now puts two things on the stack regardless of original types

Lyxal:


Lyxal:
Nvm I bamboozled myself here

PR #1458, opened 2022-10-03
Make ¨U work NOT online but offline

Lyxal:
<insert disapproving lenny face here>


Steffan153 (approved):
Ok but why can't it just always work?

Lyxal:
> Ok but why can't it just always work?

So that people can't use the online interpreter to do things like commit crimes or access illegal data. Basically it's to ensure we ain't liable for anything

PR #1480, opened 2022-10-25
Make sure to cast numbers to iterable before getting max/min

Lyxal:


Steffan153:
That's useless, do something else

Lyxal:
Not useless at all - https://codegolf.stackexchange.com/questions/128775/the-best-base-is-10-lets-reach-it/253730#253730

I'd get a byte shave with this PR

Steffan153:
You can't call it useful because you use it in one single answer.

PR #1483, opened 2022-11-08
Update version-3 with basic parser

ysthakur:
Now that a basic parser's been made, the version-3 branch can be updated so other parts can be implemented (also, there's some miscellaneous commits in version-3-parser that aren't related to the parser and need to be in version-3 too)


Lyxal (approved):
Now that my finals are almost over, we can start going full steam ahead

PR #1484, opened 2022-11-08
Allow compiling for the JVM and JS

ysthakur:
Had to make quite a few changes but it's mostly just copying from VyxalS. Everything now has an extra level of nesting, unfortunately: `src` became `shared/src` and all `js`- and `jvm`-specific code will go under those directories. I really, really wanted to also put everything inside a `vyxal` folder and maybe make separate folders for the parser and stuff just to put y'all through the torture of `shared/src/main/scala/vyxal/parser` but I settled for putting a `package vyxal` at the top of each file. `build.sbt` now looks like a Groovy programmer puked all over it. Minor changes to `.gitignore` and `scalafmt.conf` that really shouldn't be part of this PR but whatever.


Lyxal:
Thank you for keeping nesting to a minimum

PR #1485, opened 2022-11-09
Write docs next to the implementations

ysthakur:
This branch is for playing around with adding docs right next to the implementations instead of separately in elements.yaml. Might be ready to merge now.

### DIFF
--- a/shared/src/main/scala/Elements.scala
+++ b/shared/src/main/scala/Elements.scala
@@ -1,128 +1,166 @@
 package vyxal
 
-type Overload = (Tuple, String)
+import spire.math.Number
+
+type Overload = String
 
 case class Element(
     symbol: String,
+    name: String,
     arity: Option[Int],
-    desc: String,
     overloads: List[Overload],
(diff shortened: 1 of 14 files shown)


Steffan153 (approved):
pog

shared/src/main/scala/Elements.scala
```diff
@@ -0,0 +1,173 @@
...
+        "a: str, b: num -> a + b",
+        "a: str, b: str -> a + b"
+      )
+    ) {
+      case (a: Number, b: Number) => a + b
+      case (a: String, b: Number) => s"$a$b"
+      case (a: Number, b: String) => s"$a$b"
+      case (a: String, b: String) => s"$a$b"
+      case _ => throw NotImplementedError("Can't add functions :(")
+      // todo consider doing something like APL's forks
```
Steffan153, line 170: what?

Steffan153 (commented):
also why is precommit trying to run

Lyxal (approved):

PR #1487, opened 2022-11-10
Make simple repl

ysthakur:
Make a simple REPL that just reads a line of code and executes it forever. Also added practically all the flags from Vyxal 2 (I assume they'll change for Vyxal 3 but it's good to have them here for now).


Lyxal (commented):
Quick question, how does it handle multiple flags at once?

Lyxal:
That's perfectly fine. Hell, I actually prefer that and would have done that from the start in version 1 if I had known how to

Lyxal (approved):
Merge when you're ready

PR #1492, opened 2022-11-25
Assorted changes

fredwaldo:
Sorry, this PR is a bunch of unrelated changes:
- Made the code a little more idiomatic in places
- Fix parser tests with trailing nilad moving to front, add simple tests for specific cases of reduce and mold
- Vectorise the add and multiply helpers themselves instead of the element in `Impls`
- Allow setting context vars in `Interpreter.executeFn`, use that for executing `VFun`s
- Get the test-build Github Actions workflow to run on pushes to all branches instead of just version-3

### DIFF
--- a/.github/workflows/test-build.yaml
+++ b/.github/workflows/test-build.yaml
@@ -1,8 +1,6 @@
 name: Test and upload jar
 
 on:
-  push:
-    branches: ['*']
   pull_request:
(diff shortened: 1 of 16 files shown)


Lyxal (approved):
Thank you for this - I greatly appreciate the time and effort you put into making sure I'm doing this correctly!

PR #1493, opened 2022-11-27
Make provisions for globals

fredwaldo:
Made a `Globals` class with inputs, settings, and a register. Every `Context` can hold a reference to the same `Globals` object to share global inputs and stuff.


Lyxal (approved):
Excellent. This'll simplify `?` grinders.

PR #1495, opened 2022-11-27
Rename files dir to examples and add some examples

Lyxal:


Lyxal:
pre-commit.ci autofix

examples/fibo.vy
```diff
@@ -0,0 +1,4 @@
+1…1{~+…}
```
Steffan153, line 1: This is improper Fibonacci. It only prins one `1`.

Maybe `1{…~+}` or `0 1{…~+}`? Or `1 0{…~+}` to print a zero too.

examples/fibo_long.vy
```diff
@@ -0,0 +1,8 @@
+1 : →a ,
+1 : →b ,
+
+{
+  ←a ←b + :, →temp
+  ←b →a
+  ←temp →b
+}
```
Steffan153, line 8: Why not print the 1s in the loop? This is much simpler

```
1 →a
1 →b

{
  ←a :, ←b + →temp
  ←b →a
  ←temp →b
}
```

PR #1502, opened 2022-12-08
Use custom parser instead of combinator parser

Lyxal:
Benefits:

- Lambdas and lists are able to be arity grouped (instead of just numbers and strings)
- Arity grouping is nicer altogether (no need for composite nilads/monads/dyads etc
- No funky workarounds to have non-left recursive rules
- Allows for future syntax extensions to be added that would otherwise be painful with combinators
- Less code (451 lines with formatting compared to the 499 of the current parser)

Drawbacks:

- Parser code no longer closely related to EBNF
- More prone to bugs that would otherwise be mitigated by a real-world tested library
- Could be harder to understand

I think a custom parser is the best way to go forward, but a little conversation/discussion would be good.


Lyxal:
> That was _fast_.

Well the infrastructure was already there (the lexer and interpreter still exist, plus v2 parser for inspiration) + I had help from a very smart person.

Lyxal:
I forgor to do it 💀

shared/src/main/scala/Parser.scala
```diff
@@ -1,494 +1,430 @@
...
-          val temp = value.map(tok => {
-            val inner = VyxalParser.parse(List(tok))
-            inner match {
-              case Right(ast) => ast
-              case Left(error) =>
-                throw RuntimeException(s"Error while parsing $tok: $error")
+          if (
+            topLevel && program.nonEmpty && program.front == VyxalToken.StructureAllClose
+          ) {
+            program.dequeue()
```
Lyxal, line 59: This can just be setting the structure depth to 0 and placing a close on the top of the queue

Lyxal: In a previous version of the structure parsing logic, I was using an integer variable to keep track of how many structures the program was in.

It seems to be gone now though

Lyxal: It's all good I'll just have to make sure I give everything a good look over so I know I understand it when I have the chance

Lyxal:
After looking at the changes, I think I understand them. Merge when ready

Lyxal:
On second thought, don't merge yet - I need to figure out how to complex numbers first.

PR #1504, opened 2022-12-11
Make modifiers fancy macros

Lyxal:


Lyxal:
I think this is good to go - this was mainly to do the infrastructure changes, rather than ensure it's 100% bug free. Bugs can be squashed back on the custom parser branch

PR #1506, opened 2022-12-11
Deploy Scaladocs to GitHub Pages

ysthakur:
This PR is for v3, not v2. It'll deploy Scaladocs to `vyxal.github.io/Vyxal/docs` (shouldn't mess with the interpreter because the interpreter will be at `vyxal.github.io/Vyxal`). It's not actually needed because Vyxal's an application, not a library, but it might help people new to v3. I'll admit that I'm mainly doing this because it'd be kinda cool.

<s>I think GitHub Pages needs to enabled first, though.</s> Thanks lyxal

### DIFF
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -28,6 +28,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          ref: 'scaladoc'
 
       - name: Setup Pages
         uses: actions/configure-pages@v2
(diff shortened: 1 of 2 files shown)


Lyxal:
I've gone to repo settings and made it so that workflows can deploy pages. Hopefully it should work nicely with this PR

Lyxal:
Reopen if ever this becomes not useless

Steffan153:
This is useful

Lyxal:
> This is useful

user
8:29
Okay, now it kinda sorta works. If you search for VNum => VNum => VList, it should show you NumberHelpers.range
However, if you try to search for a signature that includes anything from the stdlib, it'll just say it can't resolve it
So, not terribly useful
And I just realized it wouldn't be useful for people trying to find Vyxal elements either because every element is basically (VAny, ..., VAny) => VAny

Steffan153:
Yeah but its useful for people working on vyxal to know the structure and stuff. I mean, might as well merge

PR #1511, opened 2022-12-15
Make øṄ account for overlapping matches

### LINKED ISSUE #1510: øN. seems to be broken
See chat transcript for context

(created by lyxal [here](https://chat.stackexchange.com/transcript/message/62561106))

Lyxal:
Closes #1510

### DIFF
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -5129,6 +5129,10 @@
       - '[234,"2",4,1] : 434'
       - '["234",2,4,1] : "434"'
       - '["234","2",4,1] : "434"'
+      - '["abbbabbb", "bb", "c", 1] : "acbabbb"'
+      - '["abbbabbb", "bb", "c", 2] : "abcabbb"'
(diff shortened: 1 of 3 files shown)


Steffan153 (commented):
Why is it supposed to account for overlapping matches? I don't think it should.

Lyxal:
> Why is it supposed to account for overlapping matches? I don't think it should.

Because say you have:

```
abbbabbb` `bb` `c` 1 øṄ ,
abbbabbb` `bb` `c` 2 øṄ ,
abbbabbb` `bb` `c` 3 øṄ ,
abbbabbb` `bb` `c` 4 øṄ ,
```

You would expect it to give `acbabbb`, `abcabbb`, `abbbacb`, `abbbabc`, as those are the 4 different occurrences. Currently, it only gives `acbabbb` and `abbbacb` for the first and second occurrences. This PR simply makes the behaviour a little more consistent.

[Context](https://chat.stackexchange.com/transcript/message/62558284#62558284)

Steffan153 (approved):
Ok fine.

PR #1512, opened 2022-12-17
Get rid of nonlocal returns warning

ysthakur:
Scala is going to stop supporting letting you return from a for loop because they're syntactic sugar for `.foreach`, so this PR changes the list parsing to use the `parseBranches` function that was already being used for structures.


Lyxal (approved):
lgtm (even though that site is now archived lol)

PR #1516, opened 2022-12-17
Update more-variable-options from base

Lyxal:


Lyxal:
I'm sure this won't cause any weird merge conflicts down the line

PR #1531, opened 2023-01-01
Add elements to return a more sensible diagonals order

### LINKED ISSUE #1447: Make diagonals in a more sensible order
Currently, the diagonals and antidiagonals builtins output the main (antidiagonal), all the ones down from there, and then the rest. This is annoying and often requires an `IRƒJ` or similar to fix. It would be much more intuitive to have it start at the shortest top diagonal and work its way down, i.e. for:

```
1 1 0 1
0 1 0 1
1 0 1 0
0 0 0 1
```

Diagonals would be `[[1], [0, 1], [1, 0, 0], [1, 1, 1, 1], [0, 0, 0], [1, 0], [0]]` and antidiagonals woudl be `[[1], [1, 0], [0, 1, 1], [1, 0, 0, 0], [1, 1, 0], [0, 0], [1]]

Lyxal:
Keep the original ones for backwards compatibility

Closes #1447


vyxal/elements.py
```diff
@@ -193,6 +193,28 @@ def all_antidiagonals(lhs, ctx):
     return all_diags
 
 
+def all_antidiagonals_ordered(lhs, ctx):
```
ysthakur, line 196: Do we want to add support for LazyLists?

vyxal/elements.py
```diff
@@ -248,6 +270,31 @@ def all_diagonals(lhs, ctx):
     return all_diags
 
 
+def all_diagonals_ordered(lhs, ctx):
```
ysthakur, line 273: Do we want to add support for LazyLists? We can't make this work for an infinite list of infinite lists, but I think it could be modified to work for an infinite list of finite lists.

PR #1539, opened 2023-01-15
Make `ß` retain the original parent

Lyxal:


Steffan153 (approved):

Steffan153 (commented):
What about other modifiers like `¨i`?

Lyxal:
It was only `ß` that needed special casing because it's transpiled not as a modifier but as an actual if statement. The rest of the system works fine.

Lyxal:
Okay why is `¨i` special cased in the parser?

PR #1543, opened 2023-01-18
Make Context#push variadic

ysthakur:
Steffan suggested making it variadic so we won't have to do things like `ctx.push(a); ctx.push(b); ctx.push(c)`. Also fixed a minor issue in `Context.makeFnCtx` where if you explicitly gave arguments for a function to use (instead of popping them from the stack), it would give them in the reverse order.


Lyxal (approved):
I like it. A good idea by Steffan. Nice little quality of life thing.

Issue #1270, opened 2022-07-09
Add a nice `__repr__` to lazylists


chunkybanana:
Currently, Vyxal's lazylists are printed in debugging as the default python thing. It'd be nice for their values to be printed.

Personally,  I'd prefer to have them printed with the `["1", 2, [3, 4]]` syntax since it can be pasted in again. Also, limiting their size would be nice.

cjquines:
i do like the idea of printing the first x elements (for some x), generating them if they dont exist. some questions, though:
- generating them might change program behavior if it's defined in an effectful manner?
- how would this work with the debugger? immediately evaluating the list means jumping into the code that computes it in the first place, would that be confusing?

cjquines:
actually the way you might approach that in the debugger is probably "generate lazylist elements on demand", so reprs show [...] at first for a lazylist, but then you can get more elements (stepping through code if necessary of course)

cjquines:
also, https://docs.python.org/3/library/reprlib.html?

chunkybanana:
This is more on the python side, when debugging manually, e.g. `print(LazyList([1, 2]))`.

If it changes code behaviour, we've got more bugs.

PR #1546, opened 2023-01-19
Override __repr__ for LazyLists

ysthakur:
Implements #1270. Any idea how to actually test this?


Lyxal (commented):
I'd click approve, but testing needs to be discussed first. I would assume it'd just be a call to `repr` (or use raw formatting codes in an f-string).

Steffan153:
Yeah just test with `repr(LazyList())`

Lyxal (approved):

PR #1549, opened 2023-01-19
fix incorrect behavior of multiline comment

stasoid:
`#` in `}#` does not get eaten, that's why:
This program prints 1, should print 0:
```
#{}##{
1
```
This program prints 1, should hang:
```
#{}#{ }#
1
```
**Warning**: I did not test this PR because I could not figure out how to run development version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Vyxal
 
-<!-- secret hidden comment from a bot test--> 
+<!-- secret hidden comment from a bot test-->
 
 ![Vyxal Logo](./documents/logo/vylogo.png)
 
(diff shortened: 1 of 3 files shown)


Lyxal:
I'll test this out and add some test cases before approving

Lyxal (approved):

ysthakur:
@stasoid The [README](https://github.com/Vyxal/Vyxal/blob/main/README.md#installation) has instructions on how to run Vyxal. You'll need Poetry if you want to use the Vyxal CLI, but if you only want to run tests, you can add a test named `test_foobar` to `tests/test_complex.py` and then run `pytest tests -k "foobar"` to run only that test.

stasoid:
@ysthakur Thanks, but I only needed a quick fix. I don't plan to contribute to Vyxal in the future.

PR #1550, opened 2023-01-19
Make mode work on lists of lists

### LINKED ISSUE #1548: Mode breaks on a list of lists
For example, it can't be used [here](https://codegolf.stackexchange.com/a/256816/78850) instead of `Ċ↑h` because it complains that lazylists aren't hashable

(created by lyxal [here](https://chat.stackexchange.com/transcript/message/62802727))

Lyxal:
Closes #1548


ysthakur (approved):
LGTM

PR #1552, opened 2023-01-19
Fix `IndexError: deque index out of range` while lexing unclosed multiline comment

stasoid:
This PR fixes these programs: `#{#` and `#{}`


Lyxal (approved):
lgtm

Issue #1398, opened 2022-08-12
Make replace and replace first occurrence loop over lists


Steffan153:
Like the behavior in 05AB1E.

Lyxal:
What are some examples of differences in behaviour?

Steffan153:
When given lists, it runs it for each item. Like vectorize, but instead of returning a list, running each replacement in order and then returning the final string.

PR #1554, opened 2023-01-19
Make replace and replace_first deep

ysthakur:
Trying to implement #1398. I'm not entirely sure what @Steffan153 wanted, but I made `replace` deep, hopefully that's fine.


Steffan153:
Rip this is not at all what I wanted lmao.

Steffan153:
so like `"abcdef", ["b", "d"], "z"` => `azczef`, and `"abcdef", ["b", "d"], ["x", "y"]` => `axcyef`

vyxal/helpers.py
```diff
@@ -927,6 +927,16 @@ def primitive_type(item: Any) -> Union[str, type]:
     return list
 
 
+def replace_deep(
+    lst: VyList, target: Any, replacement: Any, ctx: Context
+) -> VyList:
```
Steffan153, line 932: what's this unused useless helper lmao

ysthakur: Yeah I forgot to delete that

ysthakur:
bruh I made a whole generic replace method that could recurse into lists and start at any occurrence and have a max number of occurrences. What you want is much easier though, I'll implement that

Steffan153:
what you did could be useful too though, you cuold put it into a digraph

ysthakur:
Okay, if it should be working now. I made it so `"abcd", ["a", "b"], ["foo"]` is equivalent to `"abcd", ["a", "b"], ["foo", ""]` if that's okay (the replacements list gets extended with empty strings).

documents/knowledge/elements.yaml
```diff
@@ -5124,7 +5127,7 @@
   description: Replace the first instance of an item with another item
   arity: 3
   overloads:
-    any-any-any: a.replace_first(b, c)
+    any-any-any: a.replace(b, c, count=1). See "V" (Replace First) for specifics.
```
Steffan153, line 5130:
```suggestion
    any-any-any: a.replace(b, c, count=1). See "V" (Replace) for specifics.
```

Steffan153 (commented):
It doesn't look like this works if `lhs` is a list

ysthakur:
@Steffan153 You mean like `[1, 2, 3, 4], [1, 2], [5, 6] -> [5, 6, 3, 4]`? I don't think it should work like that for lists, in case someone wants to replace something in a nested list like `[[1, 2], [0, 1], [3, 4]], [[1, 2], [3, 4]], [[1, 0], [2, 3]] -> [[1, 0], [0, 1], [2, 3]]`

Steffan153 (commented):
What if somebody wants to replace a whole list in a list? I'd say put the recursive behavior into a digraph

ysthakur:
> What if somebody wants to replace a whole list in a list?

What about it? This PR doesn't change the behavior for lists

> I'd say put the recursive behavior into a digraph

I'll get around to it after we merge this :)

Lyxal (approved):
This approval is for when the intended behaviour is reached. The overall infrastructure of the change lgtm

Lyxal:
No merging yet

Steffan153:
> What about it? This PR doesn't change the behavior for lists

Okay so why not implement this the same way you implementented strings?

ysthakur:
@Steffan153 In case someone wants to replace a whole list inside a nested list. What if someone wants to replace an entire list in a nested list like `[[1, 2], [0, 1], [3, 4]], [1, 2], [1, 0] -> [[1, 0], [0, 1], [3, 4]]`? If it worked for lists the way it did for strings, it would instead do nothing. Do we want that behavior? I guess we could make a digraph for the current behavior on lists.

Steffan153 (approved):
I guess that makes sense.

PR #1556, opened 2023-01-20
Make scoping work like Python

ysthakur:
This PR makes the parent context of an executing lambda the context that it was originally created in instead of the context it's executing in, so that it has full access to the variables from there. This doesn't work *exactly* like Python, since it's a bit like `nonlocal x` is automatically thrown in whenever you use a variable `x` from the outer context, but I think it's the same as Vyxal v2's behavior.

After discussing with lyxal in chat, context variable N is now last condition value in while loops and context variable M is the while loop's index (they were swapped). I've updated comments and docs a bit, let me know if I put it wrong.

Also added some miscellaneous changes, since we don't have too many people to review PRs and I'm lazy.

### DIFF
--- a/shared/src/test/scala/InterpreterTests.scala
+++ b/shared/src/test/scala/InterpreterTests.scala
@@ -38,7 +38,9 @@ class InterpreterTests extends AnyFunSuite:
   test("Can the interpreter execute named functions?") {
     given ctx: Context = Context()
     ctx.push(3, 4)
-    Interpreter.execute(AST.FnDef("f", AST.Lambda(2, List.empty, AST.Command("-"))))
+    Interpreter.execute(
+      AST.FnDef("f", AST.Lambda(2, List.empty, AST.Command("-")))
+    )
     Interpreter.execute(AST.GetVar("f"))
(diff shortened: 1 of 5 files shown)


Lyxal:
I'll review this later when I get a chance to properly understand the changes

shared/src/test/scala/InterpreterTests.scala
```diff
@@ -126,4 +128,37 @@ class InterpreterTests extends AnyFunSuite:
...
+    Interpreter.executeFn(
+      VFun.fromLambda(AST.Lambda(1, Nil, AST.AugmentVar("x", AST.Command("+"))))
+    )
+    assertResult(VNum(6))(ctx.getVar("x"))
+  }
+
+  test("References to variables in other contexts") {
+    val ctx1 = Context()
+    ctx1.setVar("x", 5)
+    Interpreter.execute(AST.Lambda(1, Nil, AST.AugmentVar("x", AST.Command("+"))))(using ctx1)
```
Lyxal, line 155: Clever usage of variable augmenting to make the test cases more concise.

Lyxal (approved):
lgtm

PR #1558, opened 2023-01-21
Make decimals and complex literals work better

Lyxal:
Before decimals just didn't work, because of a call to `.toInt`.

Now, all sorts of hanging literals are handled.

### DIFF
--- a/shared/src/main/scala/VNum.scala
+++ b/shared/src/main/scala/VNum.scala
@@ -46,14 +46,16 @@ object VNum:
     // Spits into real and imaginary parts
     // todo handle empty real part
     val parts = s.split("ı")
-    val real = parts(0)
-    val imag = parts.lift(1).getOrElse("0")
+    val real = parts.lift(0).getOrElse("0")
+    val imag = if s.endsWith("ı") then "1" else parts.lift(1).getOrElse("0")
 
-    val realNum = (if real.last == '.' then real + "5" else real).toInt
-    val imagNum = (if imag.last == '.' then imag + "5" else imag).toInt
+    var realNum: String = if real.last == '.' then real + "5" else real
+    realNum = if realNum.startsWith(".") then "0" + realNum else realNum
+    var imagNum: String = if imag.last == '.' then imag + "5" else imag
+    imagNum = if imagNum.startsWith(".") then "0" + imagNum else imagNum
 
-    if imagNum == 0 then complex(realNum, 0)
-    else complex(realNum, imagNum)
+    if Real(imagNum) == 0 then complex(Real(realNum), 0)
+    else complex(Real(realNum), Real(imagNum))
   end from
 
   /** Allow pattern matching like `VNum(r, i)` */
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/VNum.scala
```diff
@@ -46,14 +46,17 @@ object VNum:
...
     // todo handle empty real part
     val parts = s.split("ı")
-    val real = parts(0)
-    val imag = parts.lift(1).getOrElse("0")
+    val real = parts.lift(0).getOrElse("0")
+    val imag = if s.endsWith("ı") then "1" else parts.lift(1).getOrElse("0")
 
-    val realNum = (if real.last == '.' then real + "5" else real).toInt
-    val imagNum = (if imag.last == '.' then imag + "5" else imag).toInt
+    // Apologies for not using val - I need to use the same variable twice
```
ysthakur, line 52: You don't need to apologize for using vars lol, if they're the better way to do things, they should be used

Steffan153 (approved):

shared/src/main/scala/VNum.scala
```diff
@@ -46,14 +46,17 @@ object VNum:
     // Spits into real and imaginary parts
     // todo handle empty real part
     val parts = s.split("ı")
-    val real = parts(0)
-    val imag = parts.lift(1).getOrElse("0")
+    val real = parts.lift(0).getOrElse("0")
+    val imag = if s.endsWith("ı") then "1" else parts.lift(1).getOrElse("0")
```
ysthakur, line 50: Not necessary but there's [this](https://scastie.scala-lang.org/XjPy7l5sSLi7KQac0naayA) cool feature that lets you match on strings

shared/src/main/scala/VNum.scala
```diff
@@ -46,14 +46,17 @@ object VNum:
...
-    val imagNum = (if imag.last == '.' then imag + "5" else imag).toInt
+    // Apologies for not using val - I need to use the same variable twice
+    var realNum: String = if real.last == '.' then real + "5" else real
+    realNum = if realNum.startsWith(".") then "0" + realNum else realNum
+    var imagNum: String = if imag.last == '.' then imag + "5" else imag
+    imagNum = if imagNum.startsWith(".") then "0" + imagNum else imagNum
 
-    if imagNum == 0 then complex(realNum, 0)
-    else complex(realNum, imagNum)
+    if Real(imagNum) == 0 then complex(Real(realNum), 0)
```
ysthakur, line 58: May I ask what this check is for? Isn't the 0 passed explicitly to `complex` going to be wrapped in a `Real` anyway?

shared/src/test/scala/InterpreterTests.scala
```diff
@@ -161,4 +164,20 @@ class InterpreterTests extends AnyFunSuite:
     assertResult((VNum(6), "foo"))((ctx1.getVar("x"), ctx2.getVar("x")))
   }
 
+  test("Numeric literals are handled correctly") {
```
ysthakur, line 167: We may have to find a different testing [style](https://www.scalatest.org/user_guide/selecting_a_style) so that each of these asserts can be its own test. Or we could use [`Checkpoints`](https://www.scalatest.org/scaladoc/3.2.15/org/scalatest/Checkpoints.html) for this particular test

shared/src/main/scala/VNum.scala
Lyxal: I don't know - it was around from before the PR, so I figured I needed to keep it.

PR #1560, opened 2023-01-21
Add a __version__ attribute

GingerIndustries:
What it says on the tin.


ysthakur (changes requested):
I don't think this is necessary. See https://stackoverflow.com/a/67097076

PR #1562, opened 2023-01-21
Write some more docs for v3

ysthakur:
Made a file for explaining how to use sbt. Need to update element implementation docs and add docs on how to write tests.

### DIFF
--- a/documentation/ElementDocumentation.md
+++ b/documentation/ElementDocumentation.md
@@ -1,8 +1,7 @@
 # Element documentation
 
 This doc will go over implementing elements. The elements are all
-implemented in [Elements.scala](/shared/src/main/scala/Elements.scala), while
-the modifiers are in [Modifiers.scala](/shared/src/main/scala/Modifiers.scala).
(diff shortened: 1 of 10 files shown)


ysthakur:
I think I'll just mark this ready for review and we can write docs on writing tests later once we've finalized how they'll work.

documentation/ElementDocumentation.md
```diff
@@ -150,6 +172,45 @@ val add = addDyadFull(
...
+```
+
+Because we used `addVect`, the element will be vectorised and work on numbers
+and lists of numbers. However, when it meets anything other than that, it will
+throw an error saying it doesn't work for those arguments.
+
+### `addDirect` example
+
+Suppose we want to make an element to pop the top 69 items off the stack and
+then push `"420"` to it.
```
Lyxal, line 199: Easter egg trigraph coming to a v3 implementation near you sometime soon

Lyxal (approved):
Thank you. I appreciate all the detail here, and how thorough the explanations of the add functions are.

PR #1563, opened 2023-01-22
Mill support for v3

ysthakur:
Since sbt more or less makes you use its shell and requires a lot of nesting in its project structure, we could maybe use [Mill](https://com-lihaoyi.github.io/mill/mill/Intro_to_Mill.html) as an alternative. However, we may get to a point and then find out Mill doesn't actually work very well with Scala.js or doesn't have a plugin we need and end up switching to sbt, so we definitely can't drop sbt yet. As a compromise, we could have both a build.sbt for sbt and a build.sc for Mill and when the project stabilizes and we're sure that Mill can do everything we want, we can get rid of sbt and make the project structure less nested.

Another nice thing about Mill is that it doesn't require people to install it globally. There's a script called `mill` that you can add to the root directory of your project and use it directly like `./mill shared.test`. Although the script I included only works on Unix, there's [millw](https://github.com/lefou/millw), which also supports Windows and is apparently recommended. However, I haven't tried out `millw` yet to see if it has problems or anything.

<s>Do **not** merge this yet: it requires a lot of testing to make sure Scala.js works and that an uber JAR can be made for the JVM version. The only thing I've checked so far is that it can compile the shared and JVM code.</s> `./mill jvm.assembly` and `./mill js.fastLinkJS` seem to work, and when you open `index.html` in your browser and open up the console, you should be able to run `Vyxal.execute`.


Lyxal (approved):
lgtm. Should be interesting to see if mill gets better support.

PR #1571, opened 2023-01-30
Add desc of ṁ for num in yml

ysthakur:
Pointed out by 97.100.97.109 [in chat](https://chat.stackexchange.com/transcript/message/62873860#62873860). `ṁ`'s num overload isn't documented in elements.yaml (but the docstring does have it).


mathcat4 (approved):
TIL

PR #1579, opened 2023-01-31
Add element decorator

ysthakur:
Adds an element decorator so that elements that have a function for them and a fixed arity don't have to be manually added to the elements dictionary. Perhaps eventually, we could also use this to completely replace elements.yaml (other than the tests)


ysthakur:
@Lyxal Well, this only moves the dictionary up. It doesn't make the file shorter (in fact, because of the new `element` function, it makes it slightly *longer*).

PR #1580, opened 2023-02-01
Allow R to handle up to base 62

### LINKED ISSUE #1568: Extend Valid Digit Range Of R (String Base Conversion From Decimal)
Currently, `R` encodes base 10 numbers in a string using the symbols `0-9A-Z`, which limits it to conversions up to base 36. I suggest extending the symbol set to `0-9A-Za-z` allowing conversions up to base 62.

ysthakur:
Fixes #1568 by letting R use a-z to go up to base 62.


Lyxal (approved):
yes

PR #1582, opened 2023-02-01
Make `o` use all integers

ysthakur:
<s>What the linked issue suggests. Since `ȯ`'s description doesn't say just positive integers, it's been changed to be all integers now to avoid duplicating the functionality of `o`.</s> Other way round now, `o` will be all integers, `ȯ` will be positive integers.

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -5126,7 +5126,7 @@ def remove(lhs, rhs, ctx):
     ts = vy_type(lhs)
     if set(vy_type(lhs, rhs)) == {types.FunctionType, NUMBER_TYPE}:
         func, count = (rhs, lhs) if ts is types.FunctionType else (lhs, rhs)
-        return vy_filter(func, infinite_positives(None, ctx), ctx)[:count]
+        return vy_filter(func, infinite_all_integers(None, ctx), ctx)[:count]
     if ts == str:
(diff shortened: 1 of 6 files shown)


Lyxal:
This needs to be the other way around. Changing `ȯ` would break too many answers, but `o` wouldn't

Lyxal (changes requested):
see the comment I left.

Lyxal:
> This needs to be the other way around. Changing `ȯ` would break too many answers, but `o` wouldn't

This is because the behaviour was originally on `ȯ`, not `o`

Lyxal:
Don't forget to change the title and description of the PR :p

Lyxal (approved):
Much better

vyxal/elements.py
```diff
@@ -5125,19 +5125,8 @@ def remove(lhs, rhs, ctx):
...
-            value = 1
-            found = 0
-            while found != rhs:
-                if lhs(value):
-                    yield value
-                    found += 1
-                value += 1
-
-        return gen()
+        func, count = (rhs, lhs) if ts is types.FunctionType else (lhs, rhs)
```
ysthakur, line 5128:
```suggestion
        func, count = (lhs, rhs) if ts is types.FunctionType else (rhs, lhs)
```

PR #1587, opened 2023-02-02
Add more v3 tests

ysthakur:
v3 sorely needs more tests. This is just a draft for now.


Lyxal:
That's right. I'm nowhere near done yet with adding tests for everything currently implemented.

shared/src/test/scala/ElementTests.scala
```diff
@@ -55,54 +174,55 @@ class ElementTests extends AnyFunSpec:
...
-        assertResult(3: VNum)(Impls.execute("1 2 + D"))
+        testCode(""" "1 2 + D" Ė """, 3)
       }
 
       it("should use the same context for executing the code") {
-        given ctx: Context = Context(inputs = List(3, 4))
+        // Doesn't use the test helpers because of context handling
+        given ctx: Context = Context(inputs = List(3, 4), testMode = true)
+        // I think I (user) may have screwed this test up at some point, since
+        // there's no Ė here anymore
```
ysthakur, line 214:
```suggestion
```
Silly billy

shared/src/main/scala/Context.scala
```diff
@@ -27,8 +29,8 @@ import scala.io.StdIn
   */
 class Context private (
     private var stack: mut.ArrayBuffer[VAny],
-    private var _contextVarN: Option[VAny] = None,
-    private var _contextVarM: Option[VAny] = None,
+    private var _contextVarPrimary: Option[VAny] = None,
```
ysthakur, line 32: Great idea making them more visually distinctive. Do we want to call them `ctxVarPrimary/Secondary` so it's a little shorter?

shared/src/main/scala/Elements.scala
```diff
@@ -189,15 +189,21 @@ object Elements:
       "a: lst -> int(a, 2), using list of digits"
     )(NumberHelpers.fromBinary)
 
-    addFull(
+    val tobinary = addVect(
```
ysthakur, line 192:
```suggestion
    val toBinary = addVect(
```

shared/src/main/scala/Context.scala
Lyxal: I think that would be fine

shared/src/main/scala/Elements.scala
```diff
@@ -189,15 +189,21 @@ object Elements:
...
-      true,
       "a: num -> convert a to binary",
       "a: str -> bin(ord(x) for x in a)"
-    )(NumberHelpers.toBinary)
+    ) {
+      case a: VNum => NumberHelpers.toBinary(a)
+      case a: String =>
+        VList(
+          a.map(x => StringHelpers.chrord(x.toString))
+            .map(NumberHelpers.toBinary(_))*
```
ysthakur, line 204:
```suggestion
          a.map(x => NumberHelpers.toBinary(StringHelpers.chrord(x.toString)))*
```

Lyxal: I could have sworn I tried something like that last night and couldn't get it to compile. Does this compile and mean that I'm just bad at scala? :p

shared/src/main/scala/Elements.scala
```diff
@@ -433,18 +431,24 @@ object Elements:
...
-        case a: VNum => a.toInt.toChar.toString
+        case a: VNum   => StringHelpers.chrord(a)
+        case a: String => StringHelpers.chrord(a)
+        case a: VList =>
+          val temp = a.map(StringHelpers.chrord)
+          if temp.forall(_ match
+              case _: String => true
+              case _         => false
+            )
+          then temp.fold("")(_.asInstanceOf[String] + _.asInstanceOf[String])
```
ysthakur, line 450:
```suggestion
          if temp.forall(_.isInstanceOf[String])
          then temp.mkString
```

Lyxal (approved):
An approval for merging when ready. Obviously, don't merge straight away.

shared/src/main/scala/Elements.scala
```diff
@@ -482,9 +486,10 @@ object Elements:
       "a: str, b: num|str -> does regex pattern b match haystack a?"
     ) {
       case (a: VNum, b: VNum) =>
-        NumberHelpers.range(a, b - 1)
-      case (a: String, b: String) => a.r.findFirstIn(b).isDefined
-      case (a: String, b: VNum)   => a.r.findFirstIn(b.toString).isDefined
+        NumberHelpers.range(a, b).dropRight(1)
```
ysthakur, line 489: May I ask why dropRight is needed instead of the b-1 from before?

Lyxal: Because of cases like `5 -> 1`

Lyxal: A range from 5 to 0 isn't what should be returned. So instead of having conditional swapping, just remove the last item of the range. That works for ranges in both directions.

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -51,9 +51,11 @@ object NumberHelpers:
   end gamma
 
   def multiplicity(a: VNum, b: VNum): VNum =
+    if a == VNum(0) || b == VNum(0) then return VNum(0)
+    if b.vabs == VNum(1) then return a.vabs
     var result = 0
     var current = a.toInt
-    while current % b.toInt == 0 do
+    while VNum(current) % VNum(b.toInt) == VNum(0) do
```
ysthakur, line 58: Since we're working with 32-bit ints here, is it necessary to use `VNum()`?

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -68,8 +70,13 @@ object NumberHelpers:
...
     a match
       case n: VNum =>
-        val binary = n.toInt.toBinaryString
-        VList(binary.map(_.asDigit: VNum)*)
+        val binary = n.toInt.abs.toBinaryString
+        val temp = VList(binary.map(_.asDigit: VNum)*)
+        if n.toInt < 0 then
+          VList(temp.map((_: @unchecked) match
+            case v: VNum => -v
+          )*)
```
ysthakur, line 78:
```suggestion
          temp.vmap(v => -v.asInstanceOf[VNum])
```

shared/src/main/scala/NumberHelpers.scala
Lyxal: Yes. It really is. They're cast to VNum for the `%` behaviour.

Default scala modulo doesn't support floating point numbers and is truncated. We need floored floating point modulo

ysthakur (commented):
I'd approve this if I could lol

Lyxal:
> I'd approve this if I could lol

That's why I clicked approve lol

shared/src/main/scala/Elements.scala
ysthakur: Oh, smart

shared/src/main/scala/NumberHelpers.scala
ysthakur: Ah, right. In that case, we might as well make `current` and the rest `VNum`s too, so we're not restricted to 32-bit ints. `b.toIntegral` should work for that (although we should only call it once and store it in a variable)

Lyxal:
Mmmm compiler errors.

shared/src/main/scala/NumberHelpers.scala
ysthakur: Maybe something like
```scala
  def multiplicity(a: VNum, b: VNum): VNum =
    if a == VNum(0) || b == VNum(0) then return VNum(0)
    if b.vabs == VNum(1) then return a.vabs
    var result = 0
    var current = a.toIntegral
    val bInt = b.toIntegral
    while current % bInt == VNum(0) do
      result += 1
      current /= bInt
    result
```

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -32,8 +70,11 @@ object NumberHelpers:
   def toBinary(a: VAny): VAny =
```
ysthakur, line 70:
```suggestion
  def toBinary(a: VAny)(using Context): VAny =
```

ysthakur:
My bad, I just wanted to use `vmap` to make it more concise

shared/src/main/scala/NumberHelpers.scala
Lyxal: Actually, I think nothing should be cast to int. I was reading the current v2 implementation and it doesn't have any cast to int

Lyxal: Just leave them all as `VNum`. No need to uncast. I probably only keep the to integral because I was keeping some of the original code from before

ysthakur: So maybe something like this?
```scala
  def multiplicity(a: VNum, b: VNum): VNum =
    if a == VNum(0) || b == VNum(0) then return VNum(0)
    if b.vabs == VNum(1) then return a.vabs
    var result = 0
    var current = a
    while current % b == VNum(0) do
      result += 1
      current /= b
    result
```

Lyxal: Exactly like that

PR #1588, opened 2023-02-03
Scala Native support

ysthakur:
This branch is for eventually creating Vyxal executables directly using Scala Native. There's also GraalVM, which we should check out too. Executables made by Scala Native will likely be smaller, since they don't contain all the Java stuff that GraalVM has. Scala Native will also be able to optimize more, since it's aware of the original Scala code, while GraalVM works on bytecode. However, the size and performance difference probably won't be big enough to be noticeable, and GraalVM has been around longer than Scala Native and works on JARs, so we may just want to go with that.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -6,7 +6,7 @@ import mill.scalanativelib._
 import mill.scalanativelib.api._
 
 /** Shared settings for all modules */
-trait VyxalModule extends SbtModule {
+trait VyxalModule extends ScalaModule {
   def platform: String
 
   def scalaVersion = "3.2.1"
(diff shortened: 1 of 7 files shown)


build.sc
```diff
@@ -47,7 +49,7 @@ trait VyxalModule extends SbtModule {
 }
 
 /** Shared and JVM-specific code */
-object jvm extends VyxalModule {
+object jvm extends ScalaModule with VyxalModule {
```
ysthakur, line 52:
```suggestion
object jvm extends VyxalModule {
```

Lyxal (approved):
I think the fact that everything has to be JS compatible means that code we write will fit just fine into the requirements of Native. As in, we won't be using anything Java anyway.

Lyxal:
besides, we can always figure out any issues when we come across them + decide to not support native if need be

Lyxal:
merge when you're ready

PR #1590, opened 2023-02-07
Create LiterateLexer.scala

Lyxal:
Literate mode is a version 3 exclusive. It works by basically transliterating words into their unicode equivalent, and running the resulting program as normal.


Steffan153 (approved):
This gonna be awesome

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -0,0 +1,158 @@
...
+    """\{""".r ~ rep(lambdaBlock | """(#}|[^{}])+""".r) ~ """\}""".r ^^ {
+      case _ ~ body ~ _ => LambdaBlock(body)
+    }
+  def normalGroup: Parser[LiterateToken] =
+    """\(""".r ~ rep(normalGroup | """[^()]+""".r) ~ """\)""".r ^^ {
+      case _ ~ body ~ _ => Word(sbcsify(body.map(recHelp).mkString))
+    }
+
+  def list: Parser[LiterateToken] =
+    "[" ~ repsep(list | """[^\]\[|]+""".r, "|") ~ "]" ^^ { case _ ~ body ~ _ =>
```
ysthakur, line 76:
```suggestion
    "[" ~> repsep(list | """[^\]\[|]+""".r, "|") <~ "]" ^^ { body =>
```
Since we don't care about the []

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  def varSet: Parser[LiterateToken] = """:=([_a-zA-Z][_a-zA-Z0-9]*)?""".r ^^ {
+    value => AlreadyCode("#" + value.substring(1))
+  }
+
+  def augVar: Parser[LiterateToken] = """:>([a-zA-Z][_a-zA-Z0-9]*)?""".r ^^ {
+    value => AlreadyCode("#>" + value.substring(2))
+  }
+
+  def unpackVar: Parser[LiterateToken] = ":=" ~ list ^^ { case _ ~ value =>
```
ysthakur, line 97:
```suggestion
  def unpackVar: Parser[LiterateToken] = ":=" ~> list ^^ { value =>
```

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -0,0 +1,158 @@
...
+    case value: String      => litLex(value)
+
+def getRight(
+    either: Either[VyxalCompilationError, List[LiterateToken]]
+): List[LiterateToken] =
+  // For when you know it's a right and don't want to have 20 thousand match statements
+  either match
+    case Right(value) => value
+    case Left(value) =>
+      List()
```
ysthakur, line 155:
```suggestion
  either.getOrElse(Nil)
```

ysthakur: Perhaps emanresu A was right in that discussion we had a while back about Eithers vs exceptions. If using Eithers is being very annoying, we could throw `VyxalCompilationError`s instead.

ysthakur (changes requested):
Love it! Just some minor stuff

shared/src/main/scala/Interpreter.scala
```diff
@@ -5,6 +5,10 @@ import vyxal.impls.Elements
 import VNum.given
 
 object Interpreter:
+  def runLiterate(code: String)(using ctx: Context): Unit =
```
ysthakur, line 8: Do we want to add a `literate: Boolean = false` parameter to `Interpreter.execute` itself instead of making a separate method?

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -0,0 +1,158 @@
...
+  // Object instead of String like the normal lexer because it's way easier
+  case Word(override val value: String) extends LiterateToken(value)
+  case AlreadyCode(override val value: String) extends LiterateToken(value)
+  // This is for strings that are already in SBCS form
+  case LitComment(override val value: String) extends LiterateToken(value)
+  case LambdaBlock(override val value: List[Object])
+      extends LiterateToken(value.toString)
+
+  case ListToken(override val value: List[Object])
+      extends LiterateToken(value.toString)
```
ysthakur, line 40: Why not `LiterateToken(value)`? `value.toString` will look like `List(..., ..., ...)`

ysthakur, line 40:
```suggestion
enum LiterateToken:
  // Object instead of String like the normal lexer because it's way easier
  case Word(value: String)
  case AlreadyCode(value: String)
  // This is for strings that are already in SBCS form
  case LitComment(value: String)
  case LambdaBlock(value: List[Object])
  case ListToken(value: List[Object])

  def value: Any
```

ysthakur: Or we could only have a `TokenType` enum and then use `(TokenType, Any)` tuples to represent tokens

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -0,0 +1,155 @@
...
+    LitComment(value)
+  }
+
+  def lambdaBlock: Parser[LiterateToken] =
+    """\{""".r ~ rep(lambdaBlock | """(#}|[^{}])+""".r) ~ """\}""".r ^^ {
+      case _ ~ body ~ _ => LambdaBlock(body)
+    }
+  def normalGroup: Parser[LiterateToken] =
+    """\(""".r ~ rep(normalGroup | """[^()]+""".r) ~ """\)""".r ^^ {
+      case _ ~ body ~ _ => Word(sbcsify(body.map(recHelp).mkString))
```
ysthakur, line 72:
```suggestion
    "(" ~> rep(normalGroup | """[^()]+""".r) <~ ")" ^^ {
      case body => Word(sbcsify(body.map(recHelp).mkString))
```

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -0,0 +1,155 @@
...
+    AlreadyCode(value)
+  }
+
+  def comment: Parser[LiterateToken] = """##[^\n]*""".r ^^ { value =>
+    LitComment(value)
+  }
+
+  def lambdaBlock: Parser[LiterateToken] =
+    """\{""".r ~ rep(lambdaBlock | """(#}|[^{}])+""".r) ~ """\}""".r ^^ {
+      case _ ~ body ~ _ => LambdaBlock(body)
```
ysthakur, line 68:
```suggestion
    "{" ~ rep(lambdaBlock | """(#}|[^{}])+""".r) ~ "}" ^^ {
      body => LambdaBlock(body)
```

shared/src/main/scala/Interpreter.scala
Lyxal: I guess it depends on what would be more concise to call in other files. I'm fine either way I think - I just made it a separate method because I wanted to avoid fricking up everything else.

shared/src/main/scala/LiterateLexer.scala
Lyxal: Copilot!

Lyxal: Honestly I was just going off the token type enum from the og lexer and modifying it. I just went with whatever looked like it would make the types happy lol.

ysthakur (approved):
On phone but looks like suggestions were applied we can work put other stuff and write docs later

PR #1595, opened 2023-02-08
Create tests.yaml and an accompanying test suite

ysthakur:
Putting tests in a tests.yaml will let HN get examples for free, and I guess it'll be more concise than what we have now. This PR adds a `YamlTests` test suite, but that needs some work, so this is only a draft for now.


jvm/src/test/scala/YamlTests.scala
```diff
@@ -0,0 +1,216 @@
...
+      criteria += Criterion.Contains(javaListToVyxal(contains))
+    val containsMonotonic = testInfo.get("contains-monotonic")
+    if containsMonotonic != null then
+      criteria +=
+        Criterion.Contains(javaListToVyxal(containsMonotonic), monotonic = true)
+
+    criteria.toSeq
+  end getOutputCriteria
+
+  private def javaToVyxal(obj: Any): VAny = obj match
```
Lyxal, line 182: is this going to work with the native tests?

ysthakur: No, Snakeyaml is a Java library. That's why I stuck it in the jvm/ folder instead of the shared/ folder

ysthakur: We could use a multiplatform Scala library instead, but I think it's fine to run these tests only on the JVM. If an element does have potentially different behavior depending on the platform, we can stick into the `ElementTests` suite in the `shared/` folder

Steffan153: I used circe here: https://github.com/Vyxal/VyxalS/pull/14

Would that work? Idk

ysthakur: It almost certainly would, I was just too lazy to use something new and figured we wouldn't need to run the tests on other platforms. I'll move it to the shared folder and use circe-yaml later (or [snake-yaml](https://github.com/VirtusLab/scala-yaml) if you want)

Lyxal (dismissed):

ysthakur:
Converting to draft since it's got an approval and I don't want to accidentally merge it

jvm/src/test/scala/YamlTests.scala
Steffan153: 🤷  do what you want

Steffan153: I don't see a problem with sticking with snake yaml

Steffan153: Idc

shared/src/main/scala/VList.scala
```diff
@@ -100,10 +101,10 @@ class VList private (val lst: Seq[VAny])
...
 
 object VList extends SpecificIterableFactory[VAny, VList]:
-  def from(it: Seq[VAny]): VList = 
-    if it.isInstanceOf[VList] then it
-    else new VList(it)
-  
+  def from(it: Seq[VAny]): VList =
+    it match
+      case temp: VList => temp
+      case _           => new VList(it)
```
Lyxal, line 107: Will this cause merge conflicts with #1741?

Lyxal: Guess we'll see upon merging lol

Lyxal (approved):

Lyxal:
> We might want to have a way for entire test groups to specify custom code to be run, rather than individual tests. Not sure how useful that'd be, though, so maybe we can wait for a use case to arise before actually implementing it.

Feel like that'd be better as just using a testing class. The yaml should be for data in data out tests imo

PR #1596, opened 2023-02-08
Make v3 Modifiers Great Again

Lyxal:
They've kinda gone unnoticed. Now they have tests, and are actually included in the elements documentation. Plus their behaviours have been fixed.


shared/src/main/scala/FuncHelpers.scala
```diff
@@ -10,24 +10,51 @@ object FuncHelpers:
       fn.arity match
         case 1 =>
           val a = ctx.pop()
-          Monad.vectoriseNoFill { a =>
+          VList(ListHelpers.makeIterable(a).map { a =>
             Interpreter.executeFn(fn, args = Some(List(a)))
-          }(a)
+          }*)
```
ysthakur, line 15: I can't make a suggestion here, but this can be
```scala
ListHelpers.makeIterable(a).vmap { a =>
  Interpreter.executeFn(fn, args = Some(List(a)))
}
```

shared/src/main/scala/FuncHelpers.scala
```diff
@@ -10,24 +10,51 @@ object FuncHelpers:
...
         case 2 =>
           val b, a = ctx.pop()
-          Dyad.vectoriseNoFill { (a, b) =>
-            Interpreter.executeFn(fn, args = Some(List(a, b)))
-          }(a, b)
+          VList(ListHelpers.makeIterable(a).zipWithIndex.map { (a, i) =>
+            Interpreter.executeFn(
+              fn,
+              args = Some(List(a, b))
```
ysthakur, line 21: Perhaps it was a bad idea to take an `Option[List[VAny]]` for the arguments. We could use `List[VAny] | Null` instead to avoid the `Some()`. What do you think?

shared/src/main/scala/FuncHelpers.scala
```diff
@@ -10,24 +10,51 @@ object FuncHelpers:
       fn.arity match
         case 1 =>
           val a = ctx.pop()
-          Monad.vectoriseNoFill { a =>
+          VList(ListHelpers.makeIterable(a).map { a =>
```
ysthakur, line 13: Does this new implementation vectorise deeply, by the way? Or does it only do one level?

ysthakur (approved):
LGTM, just confused about how the new implementation works

shared/src/main/scala/FuncHelpers.scala
Lyxal: It goes as deep as individual elements choose.

Lyxal:
> LGTM, just confused about how the new implementation works

It brings v3 vectorisation to the same way v2 vectorisation works.

shared/src/main/scala/FuncHelpers.scala
Lyxal: Honestly, I'm just going off what copilot suggested. How would this look with `List[VAny]|Null`?

Lyxal:
Why are you like this pre-commit ci?

Lyxal:
pre-commit.ci autofix

Lyxal:
No? Still unhappy?

shared/src/main/scala/FuncHelpers.scala
ysthakur: Oh, so `v` does the same job as `M`? I thought it went deep, while `M` only went one level in.

Lyxal:
I better not catch you acting goofy again you dingus

ysthakur:
You can leave the `trailing-whitespace` rule and just make sure the elements.txt file is generated without trailing whitespace

jvm/src/main/scala/Main.scala
```diff
@@ -90,12 +90,21 @@ object Main:
...
+          println(s"Keywords: ${keywords.mkString(", ")}")
           overloads.foreach { overload =>
             println(s"- $overload")
           }
           println("---------------------")
       }
 
+    Modifiers.modifiers.foreach { case (name, info) =>
+      print(s"$name\n")
+      println(s"Keywords: ${info.keywords.mkString(", ")}")
```
ysthakur, line 102:
```suggestion
      println(s"Keywords:${info.keywords.mkString(" ", ", ", "")}")
```

jvm/src/main/scala/Main.scala
```diff
@@ -90,12 +90,21 @@ object Main:
           print(
             s"$symbol ($name) (${if vectorises then "" else "non-"}vectorising)\n"
           )
+          println(s"Keywords: ${keywords.mkString(", ")}")
```
ysthakur, line 93:
```suggestion
          println(s"Keywords:${keywords.mkString(" ", ", ", "")}")
```

shared/src/main/scala/FuncHelpers.scala
ysthakur: There wouldn't be a `Some()`, we'd just pass `args = List(a, b)`

Lyxal: That does sound good

shared/src/main/scala/FuncHelpers.scala
Lyxal: It's a sort of quick map thing that's good for when something doesn't vectorise by default

ysthakur: Ah, I thought all this time that `v` meant you vectorised an element deeply.

shared/src/main/scala/FuncHelpers.scala
```diff
@@ -9,52 +9,24 @@ object FuncHelpers:
...
           }
         case 2 =>
           val b, a = ctx.pop()
-          VList(ListHelpers.makeIterable(a).zipWithIndex.map { (a, i) =>
-            Interpreter.executeFn(
-              fn,
-              args = Some(List(a, b))
-            )
-          }*)
+          ListHelpers.makeIterable(a).vmap { a =>
```
ysthakur, line 17: I gotta say, dyad vectorisation still confuses me. Why isn't `b` also turned into an iterable and zipped with `a`?

Lyxal: Because if you wanted that behaviour, you wouldn't need `v` for things that already vectorise. And for things that don't automatically vectorise, zipping is only needed maybe 10% of the time at most.

However, I'm thinking of having another modifier that uses the behaviour of vectorises from before this PR.

Lyxal: In another PR that is - it'd have to be discussed for which codepage item it'd take, etc.

PR #1597, opened 2023-02-08
Add literate mode documentation

Lyxal:
+ make it so that only `i` can be used in complex literals (the parser change is to make life easier - isolated `i`s are turned into `ı`, even inside for loops, so they are quietly turned back into `i` in for loop variable names)


shared/src/main/scala/Parser.scala
```diff
@@ -18,7 +18,10 @@ object Parser:
   type ParserRet[T] = Either[VyxalCompilationError, T]
 
   private def toValidName(name: String): String =
-    name.filter(_.isLetterOrDigit).dropWhile(!_.isLetter)
+    name
+      .filter(c => c.isLetterOrDigit || c.toString == "ı")
```
ysthakur, line 22:
```suggestion
      .filter(c => c.isLetterOrDigit || c == 'ı')
```

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -43,10 +44,10 @@ object LiterateLexer extends RegexParsers:
 
   private def decimalRegex = raw"((0|[1-9][0-9]*)?\.[0-9]*|0|[1-9][0-9]*)"
   def number: Parser[LiterateToken] =
-    raw"($decimalRegex[ij]$decimalRegex?)|([ij]$decimalRegex)|$decimalRegex|([ij]( |$$))".r ^^ {
+    raw"($decimalRegex[i]$decimalRegex?)|([i]$decimalRegex)|$decimalRegex|([i]( |$$))".r ^^ {
```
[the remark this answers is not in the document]
Lyxal: Ah yeah, forgot that the square brackets aren't necessary now lol

shared/src/main/scala/Parser.scala
Lyxal: The joys of still getting used to `char` and `String` not being interchangeable :p

ysthakur (approved):

ysthakur:
Nice, lots of details

PR #1598, opened 2023-02-09
Improve v3 site functionality + internal code

ysthakur:
Get rid of DOM stuff in JSVyxal.scala + make programs stoppable.


js/src/main/scala/JSVyxal.scala
```diff
@@ -1,42 +1,48 @@
...
-    // cast output object to textarea
-    val textarea = output.asInstanceOf[dom.html.TextArea]
-    textarea.value = ""
-    val settings =
-      Settings(online = true)
+    val settings = flags.foldLeft(Settings(online = true))(_.withFlag(_))
     val globals = Globals(
-      settings = settings
+      settings = settings,
+      printFn = (x: VAny) =>
```
Lyxal, line 24: This cursed conglomerate of a print function is so that programs can live stream their input to the output box and still be interrupted. It's not just a neat feature, it's a consequence of basically the only way I could think of allowing programs to be stopped.

pages/main.js
```diff
@@ -713,27 +723,61 @@ window.addEventListener("DOMContentLoaded", e => {
     const stdin = document.getElementById("inputs")
     const flags = document.getElementById("flag")
     const output = document.getElementById("output")
-    const extra = document.getElementById("extra")
+    const extra = document.getElementById("debug")
     const filter = document.getElementById("filterBox")
 
     async function do_run() {
+        timeout = 10000
```
Lyxal, line 730: dynamically set timeout from flags

js/src/main/scala/JSVyxal.scala
```diff
@@ -1,7 +1,6 @@
 package vyxal
 
-import org.scalajs.dom
-import org.scalajs.dom.document
+import scala.annotation.static
```
ysthakur, line 3: Was this left in from a previous edit?

js/src/main/scala/JSVyxal.scala
```diff
@@ -10,33 +9,35 @@ import scala.scalajs.js.JSConverters.*
 @JSExportTopLevel("Vyxal")
 object JSVyxal:
   @JSExport
-  def execute(code: String, inputs: String, flags: String): String =
+  def execute(
+      code: String,
+      inputs: String,
+      flags: String,
+      printFunc: String
```
ysthakur, line 16: Why not take a `js.Function1[String, Unit]`? That can easily be converted to a Scala function (`String => Unit`)

pages/main.js
```diff
@@ -1,4 +1,5 @@
 const $ = x => document.getElementById(x)
+
 var codepage = "λƛ¬∧⟑∨⟇÷×«␤»°•ß†€"
```
ysthakur, line 3: We might want to consider adding a method inside the Scala.js code to get the codepage so it won't have to be duplicated here

pages/worker.js
```diff
@@ -1,5 +1,11 @@
-
+importScripts("../js/lib/scalajs-3.0.0.js")
```
ysthakur, line 1: One problem with this is that it's hard-coding the version. We'll have to either rename the generated file to not include the Vyxal version in its name, or generate this pages/worker.js file itself.

js/src/main/scala/JSVyxal.scala
```diff
@@ -10,33 +9,35 @@ import scala.scalajs.js.JSConverters.*
...
-    val textarea = output.asInstanceOf[dom.html.TextArea]
-    textarea.value = ""
-    val settings =
-      Settings(online = true)
+    val settings = flags.foldLeft(Settings(online = true))(_.withFlag(_))
     val globals = Globals(
-      settings = settings
+      settings = settings,
+      printFn = (x: VAny) =>
+        val f = js
```
ysthakur, line 24: This evals `printFunc` every time you call it. Is that desired?

js/src/main/scala/JSVyxal.scala
```diff
@@ -10,33 +9,35 @@ import scala.scalajs.js.JSConverters.*
...
+          .eval(s"$printFunc")
+          .asInstanceOf[js.Function1[String, Unit]]
+        f.apply(if x.toString.isEmpty then "" else x.toString)
```
ysthakur, line 27: You used Copilot here, didn't you?

js/src/main/scala/JSVyxal.scala
Lyxal: I think so

js/src/main/scala/JSVyxal.scala
Lyxal: Yes, it is

js/src/main/scala/JSVyxal.scala
Lyxal: This is the one time I searched through docs to write a cursed mess :p

js/src/main/scala/JSVyxal.scala
Lyxal: This whole thing was painful enough that I didn't want to try experimenting with forcing it to scala types explicitly

js/src/main/scala/JSVyxal.scala
ysthakur: Oh ok, just asking because `f.apply(if x.toString.isEmpty then "" else x.toString)` is the same as `f.apply(x)`

pages/main.js
Lyxal: Oh yeah we can do that now. Very cool

js/src/main/scala/JSVyxal.scala
Lyxal: The empty string check is an artifact from something else I tried lol

pages/worker.js
Lyxal: I think not including version is a better idea.

chunkybanana (approved):

chunkybanana (approved):

Steffan153 (approved):

PR #1599, opened 2023-02-10
Make literate mode generator golfier

Lyxal:


shared/src/main/scala/LiterateLexer.scala
```diff
@@ -125,12 +131,66 @@ def recHelp(token: Object): String =
...
     case LitComment(value)  => value
+    case Number(value)      => value
+    case Variable(value)    => value
     case LambdaBlock(value) => value.map(recHelp).mkString("λ", " ", "}")
     case ListToken(value)   => value.map(recHelp).mkString("[", "|", "]")
     case value: String      => value
 
 def sbcsify(tokens: List[LiterateToken]): String =
-  tokens.map(sbcsify).mkString("", " ", "")
+  val out = mutable.StringBuilder()
```
ysthakur, line 141:
```suggestion
  val out = StringBuilder()
```
This should work afaik. If that doesn't work, see if `new StringBuilder()` does

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -33,11 +34,16 @@ enum LiterateToken:
...
   // This is for strings that are already in SBCS form
+  case Number(value: String)
+  case Variable(value: String)
+
+  case Group(value: List[Object])
   case LitComment(value: String)
   case LambdaBlock(value: List[Object])
   case ListToken(value: List[Object])
 
   def value: Any
```
ysthakur, line 45: Perhaps this method could be removed

ysthakur (approved):
LGTM, this is very useful

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -1,6 +1,7 @@
 package vyxal
 
 import java.util.regex.Pattern
+import scala.collection.mutable
```
ysthakur, line 4:
```suggestion
```

PR #1600, opened 2023-02-10
Make common CLI

ysthakur:
This branch is for making a common CLI for JVM and Native, instead of duplicating the code in both places. <s>I think we can commands to the REPL before merging, so I'll let this be a draft for now.</s> Looks like Ginger wants to work on the REPL, so we can just merge this and let them implement the REPL in a later PR.

I ended up also modifying the workflows a bit: there are now 2 separate workflows for native and JVM stuff (and they only upload artifacts if on version-3). The scalafix workflow also runs the formatter now.


ysthakur:
idk why the workflows aren't running, will figure out later

Lyxal (approved):

PR #1601, opened 2023-02-11
Update deploy-pages.yaml

Steffan153:

PR #1602, opened 2023-02-11
undo dumb thing i did

Steffan153:

PR #1603, opened 2023-02-11
Don't delete pages

ysthakur:
Reverting [this](https://github.com/Vyxal/Vyxal/pull/1601) well-intentioned PR that removes the pages directory, which contains important files like main.js


ysthakur:
Duplicate of #1602

PR #1606, opened 2023-02-11
Make everything compile in JS & Native

Lyxal:
+ make the print function not println
+ make the native workflow actually run
  (for some reason, it didn't want to recognise the `env`)


.scalafmt.conf
```diff
@@ -18,6 +18,8 @@ rewrite.imports.sort = original
 rewrite.imports.groups = [["vyxal\\..*"]]
 rewrite.imports.contiguousGroups = no
 
+rewrite.trailingCommas.style = keep # frick you removing my commas.
```
ysthakur, line 21: We could even make this always

native/src/main/scala/Main.scala
```diff
@@ -4,4 +4,8 @@ import vyxal.cli.CLI
 
 object Main:
   def main(args: Array[String]): Unit =
-    CLI.run(args)
+    if args.nonEmpty && args(0).charAt(0) != '-' then
```
ysthakur, line 7: This logic should be handled in `CLI`

shared/src/main/scala/cli/CLI.scala
```diff
@@ -31,7 +28,7 @@ object CLI:
       settings: Settings = Settings()
   )
 
-  def run(args: Array[String]): Unit =
+  def run(args: Array[String], program: Option[String]): Unit =
```
ysthakur, line 31: Is scopt not parsing code given directly correctly?

ysthakur:
Sorry, I didn't even realize that the workflows weren't running

ysthakur:
If I understand correctly, the only reason to split up the code was because `java.io.File` didn't compile in JS, so I modified it to take the file name as a string instead of taking a `File`. Also modified `execute` to take a boolean `literate` parameter to avoid all the `if`s everywhere.

jvm/src/main/scala/Main.scala
```diff
@@ -2,6 +2,8 @@ package vyxal
 
 import vyxal.cli.CLI
 
+import java.io.File
```
ysthakur, line 5:
```suggestion
```

jvm/src/main/scala/Main.scala
```diff
@@ -2,6 +2,7 @@ package vyxal
 
 import vyxal.cli.CLI
 
+
```
ysthakur, line 5:
```suggestion
```

ysthakur:
Do vyxal.js and vyxal.js.map have to be added to git? I feel like we could upload them as artifacts or something

ysthakur (approved):
My bad, didn't see that the JS didn't compile. I don't know why the native workflow doesn't evaluate the matrix.os variable, but maybe that happens only if it's on version-3?

Steffan153 (approved):

PR #1608, opened 2023-02-12
Automatic release workflow

Lyxal:
Make workflow to automatically create releases when a tag is made. Assets will also be made (jar and executables for Ubuntu, Windows, and MacOS)


ysthakur (approved):
Approving this in case lyxal wants to merge later

ysthakur:
Screw it, I'm going to merge

PR #1611, opened 2023-02-13
Make Lambdas Great (for the first time really)

Lyxal:
Multi-branch lambdas + parameter parsing


shared/src/main/scala/AST.scala
```diff
@@ -55,11 +55,12 @@ enum AST(val arity: Option[Int]):
     case If(thenBody, elseBody)  => s"[$thenBody|$elseBody}"
     case For(loopVar, body)      => s"(${loopVar.getOrElse("")}|${body.toVyxal}"
     case While(cond, body) => s"{${cond.fold("")(_.toVyxal)}|${body.toVyxal}}"
-    case Lambda(arity, params, body) => s"λ${body.toVyxal}}"
-    case FnDef(name, lam)            => ???
-    case GetVar(name)                => s"#<$name"
-    case SetVar(name)                => s"#>$name"
-    case ast                         => ast.toString
+    case Lambda(arity, params, body) =>
+      s"λ${body.map(_.toVyxal).mkString("|")}}"
```
ysthakur, line 59:
```suggestion
      body.map(_.toVyxal).mkString("λ", "|", "}")
```

shared/src/main/scala/ListHelpers.scala
```diff
@@ -5,6 +5,28 @@ import VNum.given
...
+    val filtered = iterable.zipWithIndex.filter((item, index) =>
+      var temp = true
+      for branch <- branches do
+        temp = temp && MiscHelpers.boolify(
+          VFun
+            .fromLambda(AST.Lambda(1, List.empty, List(branch)))
+            .execute(item, index, List(item))
+        )
+      temp
+    )
```
ysthakur, line 25:
```suggestion
    val filtered = iterable.zipWithIndex.filter { (item, index) =>
      branches.forall { branch =>
        MiscHelpers.boolify(
          VFun
            .fromLambda(AST.Lambda(1, List.empty, List(branch)))
            .execute(item, index, List(item))
        )
      }
    }
```
The braces are just a personal style thing, we can keep the parens if you want.

shared/src/main/scala/ListHelpers.scala
```diff
@@ -34,9 +56,21 @@ object ListHelpers:
...
+          f.execute(item, index, List(item))
+        }*)
+    var temp: VList = to
+    for branch <- branches do
+      temp = VList(temp.zipWithIndex.map { (item, index) =>
+        VFun
+          .fromLambda(AST.Lambda(1, List.empty, List(branch)))
+          .execute(item, index, List(item))
+      }*)
+    temp
```
ysthakur, line 72:
```suggestion
    branches.foldLeft(to) { (mapped, branch) =>
      VList(mapped.zipWithIndex.map { (item, index) =>
        VFun
          .fromLambda(AST.Lambda(1, List.empty, List(branch)))
          .execute(item, index, List(item))
      }*)
    }
```

shared/src/main/scala/ListHelpers.scala
```diff
@@ -67,6 +101,20 @@ object ListHelpers:
     moldHelper(content, shape, 0)
   end mold
 
+  def sortBy(iterable: VList, key: VFun)(using ctx: Context): VList =
+    return VList(
```
ysthakur, line 105:
```suggestion
    VList(
```

shared/src/main/scala/MiscHelpers.scala
```diff
@@ -28,6 +28,26 @@ object MiscHelpers:
     case (a: VNum, b: String)   => a.toString.compareTo(b)
     case (a: String, b: String) => a.compareTo(b)
 
+  def compareExact(
+      a: VAny,
+      b: VAny
+  )(using ctx: Context): Int = (a, b) match
+    case (a: VVal, b: VVal) => compare(a, b)
```
ysthakur, line 35: I'm thinking we might want to get rid of `VVal` and `VAtom`. It's not immediately clear to someone new to our codebase what that means (in fact, I have to take a few seconds to remember what they are every time I see them, and I wrote them)

shared/src/main/scala/MiscHelpers.scala
```diff
@@ -28,6 +28,26 @@ object MiscHelpers:
...
+
+      if aIter.length != bIter.length then
+        return aIter.length.compare(bIter.length)
+
+      var ind = 0
+      var result = -1
+      while ind < aIter.length && result != 0 do
+        result = compareExact(aIter(ind), bIter(ind))
+        ind += 1
+      return 0
```
ysthakur, line 49: Was this supposed to be `return result`?

shared/src/main/scala/ListHelpers.scala
Lyxal: This is going to be changed anyway lol

ysthakur:
For the `compare` method, you might be able to use [`Ordering`](https://www.scala-lang.org/api/2.13.5/scala/math/Ordering$.html) or something. I would try to help but I am too sleepy

Lyxal:
I've already made a `compareExact`

shared/src/main/scala/ListHelpers.scala
```diff
@@ -67,6 +99,54 @@ object ListHelpers:
...
+              key.executeResult(b(0), b(1), List(b(0)))
+            )
+          )
+          .map(_._1)*
+      )
+
+    val out = iterable.zipWithIndex
+      .sortWith { (a, b) =>
+        val (aRes, bRes) =
+          branches
```
ysthakur, line 132:
```suggestion
          branches
            .view
```
Needs to be lazy

shared/src/main/scala/ListHelpers.scala
```diff
@@ -67,6 +99,55 @@ object ListHelpers:
...
+            .view
+            .map { branch =>
+              val f = VFun.fromLambda(AST.Lambda(1, List.empty, List(branch)))
+              (
+                f.execute(a(0), a(1), List(a(0))),
+                f.execute(b(0), b(1), List(b(0)))
+              )
+            }
+            .dropWhile(_ == _)
+            .head
```
ysthakur, line 142:
```suggestion
            .find(_ != _)
            .getOrElse(ctx.settings.defaultValue)
```

Lyxal: What does this do?

Lyxal:
That funky stuff with `Impls.exec` is to make it so that lambdas acting on the stack don't push an extra 0.

Also, no clue why the test isn't passing for it.

Lyxal:
The scala fix thing is because the method wasn't showing in vs code lol

Lyxal:
There are probably better ways of making `!` arity lambdas operate on the whole stack, but this was the best I could come up with

ysthakur:
@Lyxal I meant for the compareExact method. We could turn those last 6 lines into like 2 lines

shared/src/test/scala/ElementTests.scala
```diff
@@ -121,7 +121,9 @@ class ElementTests extends VyxalTests:
           AST.Lambda(
             1,
             List.empty,
-            AST.makeSingle(AST.Number(8), AST.Command("-"))
+            List(
+              AST.makeSingle(AST.Number(8), AST.Command("-"))
```
ysthakur, line 125:
```suggestion
              AST.Number(8), AST.Command("-")

```

shared/src/main/scala/ListHelpers.scala
ysthakur: If they compare equal for all branches, `.dropWhile(_ == _)` would return an empty list, so `.head` would have thrown an exception. This one finds the first branch where they're not equal, and if such a branch can't be found, gives `(0, 0)` (or whatever the default value is set to be)

shared/src/main/scala/Parser.scala
```diff
@@ -2,6 +2,7 @@ package vyxal
 
 import vyxal.impls.Elements
 
+import java.sql.Struct
```
ysthakur, line 5: ...

Steffan153: Autoimport go brr

Lyxal:
Test comment

shared/src/main/scala/Context.scala
```diff
@@ -88,6 +88,8 @@ class Context private (
     if useStack then getTopCxt().push(items*)
     else stack ++= items
 
+  def length(): Int = stack.length
```
ysthakur, line 91:
```suggestion
  def length: Int = stack.length
```
It's totally a matter of personal preference, but since this is Scala, figured we could do it like this. It's a constant time operation anyway

shared/src/main/scala/Elements.scala
```diff
@@ -287,6 +288,21 @@ object Elements:
         case list: VList => list.vmap(execHelper)
         case _ => throw new Exception("Can't exec on functions in lists")
 
+    val execNotPop = addDirect(
+      "Ḃ",
+      "Execute lambda without popping | Evaluate as Vyxal without popping",
```
ysthakur, line 293:
```suggestion
      "Execute lambda without popping arguments | Evaluate as Vyxal without popping arguments",
```

Lyxal: This is only a temporary element for testing anyway. The description was copilot generated and if anything, this will probably be a digraph

shared/src/main/scala/Interpreter.scala
```diff
@@ -133,48 +133,41 @@ object Interpreter:
...
+                popped += termCount
+                val terms = ctx.pop(termCount)
+                popped ++= terms
+                temp += VList(terms*)
+              else
+                val top = ctx.pop()
+                vars(name) = top // set variable
+                popped += top
+        end for
+        if !popArgs then ctx.push(popped.toList.reverse*)
```
ysthakur, line 169: One problem with this is that if you ran out of elements to pop off the stack, it would've gotten some inputs, so now this line would push those inputs onto the stack. We need to use that `ctx.length` method you added to only push the necessary number of elements. Maybe something like this?
```scala
// At the top
val origLength = ctx.length
val inputs = ...
if !popArgs then
  ctx.push(popped.toList.take(origLength).reverse*)
```

shared/src/main/scala/ListHelpers.scala
```diff
@@ -61,10 +61,14 @@ object ListHelpers:
         return VList(to.zipWithIndex.map { (item, index) =>
           f.execute(item, index, List(item))
         }*)
+
+    val params = f.originalAST match
+      case Some(lam) => lam.params
+      case None      => List.empty
```
ysthakur, line 67:
```suggestion
    val params = f.originalAST.fold(Nil)(_.params)
```
I'm not sure how complex this would seem to someone new to Scala, so don't accept if you think this makes the code less understandable

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -141,20 +138,22 @@ object LiterateLexer extends RegexParsers:
     AlreadyCode("!")
   }
 
+  def colon: Parser[LiterateToken] = ":" ^^ { value =>
+    AlreadyCode("|")
+  }
+
+  def comma: Parser[LiterateToken] = "," ^^ { value =>
+    AlreadyCode(",")
+  }
```
ysthakur, line 147:
```suggestion
  def colon: Parser[LiterateToken] = ":" ^^^ AlreadyCode("|")

  def comma: Parser[LiterateToken] = "," ^^^ AlreadyCode(",")
```

shared/src/main/scala/Parser.scala
```diff
@@ -335,7 +335,10 @@ object Parser:
 
   private def parseParameters(params: AST): (List[String | Int], Int) =
     val paramString = params.toVyxal
-    val components = paramString.split(",")
+    val components = paramString.split(
+      ","
+    ) // may leave extra spaces, but that's okay, because
+    // spaces are removed when converting to a valid name
```
ysthakur, line 341:
```suggestion
    val components = paramString.split(",")
    // ^ may leave extra spaces, but that's okay, because
    // spaces are removed when converting to a valid name
```
Just wanted to get that `split(",")` all on one line

shared/src/main/scala/Elements.scala
ysthakur: Ah, ok

shared/src/main/scala/Elements.scala
```diff
@@ -560,6 +639,17 @@ object Elements:
...
     }
 
+    val wrap = addDirect(
+      "W",
+      "Wrap",
+      List("wrap"),
+      None,
+      "a, b, c, ..., -> [a, b, c, ...]"
+    ) { ctx ?=>
+      val args = ctx.pop(ctx.length())
```
ysthakur, line 649:
```suggestion
      val args = ctx.pop(ctx.length)
```

shared/src/main/scala/Interpreter.scala
Lyxal: Somehow, despite thinking about this when I was doing my thinking, I forgot to do this

shared/src/main/scala/ListHelpers.scala
Lyxal: If I don't know what it does exactly, then there's a chance others won't too :p

shared/src/main/scala/Interpreter.scala
```diff
@@ -126,24 +129,84 @@ object Interpreter:
...
-      else if popArgs then ctx.pop(arity)
-      else ctx.peek(arity)
+      if args != null && params.isEmpty then args
+      else if arity == -1 then List.empty // operates on entire stack
+      else if params.isEmpty then // no params, so just pop the args
+        if popArgs then ctx.pop(arity) else ctx.peek(arity)
+      else
+        var argIndex: Int = 0
+        val origLength = ctx.length
+        val popFunction = (n: Option[Int]) =>
```
ysthakur, line 143:
```suggestion
        def popFunction(n: Option[Int]): VAny | Seq[VAny] =
```
Since the result type is a union type, I think you need to explicitly give the result type or it'll infer `Any`.

shared/src/main/scala/Context.scala
```diff
@@ -35,18 +35,19 @@ class Context private (
     private val parent: Option[Context] = None,
     val globals: Globals = Globals(),
     val testMode: Boolean = false,
-    var onlineOutput: String = ""
+    val useStack: Boolean = false
```
ysthakur, line 38: Remind me again why we need to use the top stack and not just the parent's stack?

shared/src/main/scala/Elements.scala
```diff
@@ -259,20 +261,46 @@ object Elements:
...
+    }
+
+    def execHelper(value: VAny)(using ctx: Context): VAny =
+      value match
+        case code: String =>
+          Interpreter.execute(code)
+          ctx.pop()
+        case n: VNum     => 10 ** n
+        case list: VList => list.vmap(execHelper)
+        case _ => throw new Exception("Can't exec on functions in lists")
```
ysthakur, line 289: If we can exec code in lists, why not functions too?

shared/src/main/scala/Elements.scala
```diff
@@ -319,17 +347,47 @@ object Elements:
...
       "n",
       "Get Context Variable N",
-      List("get-context-n", "context-n", "c-var-n", "ctx-n", "ctx-primary"),
+      List(
+        "get-context-n",
+        "context-n",
+        "c-var-n",
+        "ctx-n",
+        "ctx-primary",
+        "n"
```
ysthakur, line 389: Are `n` and `m` are necessary to give as keywords? If someone writes `n` in literate mode, it's as if it's already been sbcsified, right?

shared/src/main/scala/ListHelpers.scala
```diff
@@ -5,6 +5,27 @@ import VNum.given
 
 object ListHelpers:
 
+  def filter(iterable: VList, predicate: VFun)(using ctx: Context): VList =
+    val branches = predicate.originalAST match
+      case Some(lam) => lam.body
+      case None =>
+        return VList(iterable.zipWithIndex.map { (item, index) =>
```
ysthakur, line 12: The `return` here confused me a bit, I thought you were setting branches to `VList(iterable.zipWith...` at first. I think all the code below (`val filtered = ...`) can go into the `Some` branch above

shared/src/main/scala/ListHelpers.scala
```diff
@@ -34,9 +55,24 @@ object ListHelpers:
         else VList(num.toString.map(x => VNum(x.toString))*)
 
   def map(f: VFun, to: VList)(using ctx: Context): VList =
-    VList(to.zipWithIndex.map { (item, index) =>
-      f.execute(item, index, List(item))
-    }*)
+    val branches = f.originalAST match
+      case Some(lam) => lam.body
+      case None =>
+        return VList(to.zipWithIndex.map { (item, index) =>
```
ysthakur, line 61: Same comment as for `filtered`

shared/src/main/scala/ListHelpers.scala
```diff
@@ -67,6 +103,54 @@ object ListHelpers:
...
+            .map { branch =>
+              val f = VFun.fromLambda(AST.Lambda(1, List.empty, List(branch)))
+              (
+                f.execute(a(0), a(1), List(a(0))),
+                f.execute(b(0), b(1), List(b(0)))
+              )
+            }
+            .find(_ != _)
+            .getOrElse((ctx.settings.defaultValue, ctx.settings.defaultValue))
+        MiscHelpers.compareExact(aRes, bRes) < 0
```
ysthakur, line 146:
```suggestion
            .map { case (aRes, bRes) => MiscHelpers.compareExact(aRes, bRes) < 0 }
            .getOrElse(false)
```
On second thought, there's no need to use the default value because all we want to do is return false in the case that they're equal.

(this suggestion is part of the review but it's not a change I 100% want to happen, it's just a suggestion you can reject)

shared/src/main/scala/Context.scala
Lyxal: Well that's dependent on whether stack lambdas should operate on the global stack or the parent stack

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -103,113 +127,125 @@ object LiterateLexer extends RegexParsers:
         AlreadyCode("#:" + value.map(recHelp).mkString("[", "|", "]"))
   }
 
+  def newline: Parser[LiterateToken] = "\n" ^^ { value =>
+    Newline(value)
+  }
```
ysthakur, line 132:
```suggestion
  def newline: Parser[LiterateToken] = "\n" ^^^ Newline("\n")
```

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -103,113 +127,125 @@ object LiterateLexer extends RegexParsers:
...
+
   def branch: Parser[LiterateToken] = "|" ^^ { value =>
     AlreadyCode("|")
   }
```
ysthakur, line 136:
```suggestion
  def branch: Parser[LiterateToken] = "|" ^^^ AlreadyCode("|")
```

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -103,113 +127,125 @@ object LiterateLexer extends RegexParsers:
...
 
+  def tilde: Parser[LiterateToken] = "~" ^^ { value =>
+    AlreadyCode("!")
+  }
```
ysthakur, line 140:
```suggestion
  def tilde: Parser[LiterateToken] = "~" ^^^ AlreadyCode("!")
```

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -103,113 +127,125 @@ object LiterateLexer extends RegexParsers:
...
+          if value == "0" then out.append("0")
+          else
+            next match
+              case Some(Number(nextNumber)) =>
+                if nextNumber == "." && value.endsWith(".") then
+                  out.append(value)
+                else out.append(value + " ")
+              case Some(Group(items)) =>
+                if items.length == 1 &&
+                  LiterateLexer(items.head.toString)
```
ysthakur, line 198: Perhaps we should instead use `LiterateLexer(value + item.head.toString)` to check if it's interpreted as a single number when the two are together. Otherwise, value might be `1.1` and `item.head` might be `.`, so it might have been totally fine for them to not have a space in between but this would add an unnecessary space. However, if we do use `LiterateLexer(value + item.head.toString)`, we will have to check that the parsed number isn't just `value` but `value` and some part of `item.head` before adding a space.

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -35,13 +52,20 @@ enum LiterateToken:
...
   case LambdaBlock(value: List[Object])
   case ListToken(value: List[Object])
 
 object LiterateLexer extends RegexParsers:
+  lazy val literateModeMappings: Map[String, String] =
+    Elements.elements.values.view.flatMap { elem =>
+      elem.keywords.map(_ -> elem.symbol)
+    }.toMap ++ Modifiers.modifiers.view.flatMap { (symbol, mod) =>
+      mod.keywords.map(_ -> symbol)
+    }.toMap
```
ysthakur, line 67:
```suggestion
    }.toMap ++ hardcodedKeywords
```

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -103,113 +127,125 @@ object LiterateLexer extends RegexParsers:
...
+    for i <- tokens.indices do
+      val token = tokens(i)
+      val next = if i == tokens.length - 1 then None else Some(tokens(i + 1))
+      token match
+        case Word(value) =>
+          out.append(
+            literateModeMappings.getOrElse(
+              value,
+              hardcodedKeywords.getOrElse(value, value)
+            )
```
ysthakur, line 186:
```suggestion
            literateModeMappings.getOrElse(value, value)
```

shared/src/main/scala/LiterateLexer.scala
```diff
@@ -103,113 +127,125 @@ object LiterateLexer extends RegexParsers:
...
-            case Some(Group(items)) =>
-              if items.length == 1 && getRight(
-                  LiterateLexer(
-                    items.head.toString
+            case Some(Word(w)) =>
+              if "[a-zA-Z0-9_]+".r.matches(
+                  literateModeMappings.getOrElse(
+                    w,
+                    hardcodedKeywords.getOrElse(w, "")
                   )
```
ysthakur, line 213:
```suggestion
                  literateModeMappings.getOrElse(w, "")
```

shared/src/main/scala/Parser.scala
```diff
@@ -322,6 +333,38 @@ object Parser:
...
+          val num = component.toInt
+          arity += num
+          paramList += num
+        else if component.startsWith("!") then
+          // operate on entire stack, so set arity to -1 and remove all other parameters
+          // also, process no other parameters
+          arity = -1
+          paramList.drop(paramList.length)
+        else if component == "*" || component == "×" then
+          // varargs - pop n and pop n items onto lambda stack
```
ysthakur, line 356: Is "pop n and pop n" a typo?

shared/src/main/scala/VAny.scala
```diff
@@ -37,17 +38,40 @@ case class VFun(
...
+  )(using ctx: Context): VAny =
+    val res = Interpreter.executeFn(
+      this,
+      Some(contextVarPrimary),
+      Some(contextVarSecondary),
+      args = args
+    )
+
+    res match
+      case f: VFun =>
```
ysthakur, line 65: Why execute the function if a function was returned? That seems dangerous

ysthakur (changes requested):

shared/src/main/scala/Context.scala
ysthakur: If `useStack` is enabled, then it uses the global stack no matter what. Should it be renamed to `useGlobalStack`?

shared/src/main/scala/Elements.scala
Lyxal: Because it was 12am when I made the `execHelper` function and I was way too tired to handle the function case properly.

shared/src/main/scala/Elements.scala
Lyxal: For some reason I thought they weren't already sbcsified.

shared/src/main/scala/VAny.scala
Lyxal: Because when you're sorting a list or filtering a list by a function, you don't want to compare by a function. You want to compare by the result of calling that function.

Imagine you have something like

```
{0 < if negate end} :=abs
"some list of numbers" sort-lambda $abs then negate end
```

You don't want to order by the literal function `$abs`, you want to compare by the result.

shared/src/main/scala/ListHelpers.scala
Lyxal: This seems to error anyway, saying it's a `Unit`, not a `Boolean`

shared/src/main/scala/Context.scala
Lyxal: As I said, that depends on whether we want stack lambdas to use the global stack or the parent stack. The behaviour is subject to change

ysthakur: Ah, ok

shared/src/main/scala/ListHelpers.scala
```diff
@@ -104,8 +106,42 @@ object ListHelpers:
...
+                  (
+                    f.execute(a(0), a(1), List(a(0))),
+                    f.execute(b(0), b(1), List(b(0)))
+                  )
+                }
+                .find(_ != _)
+                .getOrElse(
+                  (ctx.settings.defaultValue, ctx.settings.defaultValue)
+                )
+            MiscHelpers.compareExact(aRes, bRes) < 0
```
ysthakur, line 140:
```suggestion
            branches.view
                .map { branch =>
                  val f =
                    VFun.fromLambda(AST.Lambda(1, List.empty, List(branch)))
                  (
                    f.execute(a(0), a(1), List(a(0))),
                    f.execute(b(0), b(1), List(b(0)))
                  )
                }
                .find(_ != _)
                .map { case (aRes, bRes) => MiscHelpers.compareExact(aRes, bRes) < 0 }
                .getOrElse(false)
```
Didn't remove the `val ... =` part before

shared/src/main/scala/ListHelpers.scala
ysthakur: See the new suggestion below

shared/src/main/scala/Elements.scala
ysthakur, line 289:
```suggestion
      ctx.push(execHelper(ctx.pop())
    }

    def execHelper(value: VAny)(using ctx: Context): VAny =
      value match
        case code: String =>
          Interpreter.execute(code)
          ctx.pop()
        case n: VNum     => 10 ** n
        case list: VList => list.vmap(execHelper)
        case fn: VFun =>
          ctx.push(Interpreter.executeFn(fn))
            if fn.arity == -1 then
              ctx.pop() // Handle the extra value pushed by lambdas that operate on the stack
            ctx.pop()
```

PR #1634, opened 2023-02-18
Expose element and modifier documentation in JS

Steffan153:


GingerIndustries (approved):
LGTM (as long as you've tested it q:)

mathcat4 (approved):
lgtm, compiles without any errors on my machine, I'm pretty sure you don't need `impl` tho :P Would that increase the load time?

Steffan153:
@mathcat4 `impl` isn't part of the end object.

PR #1644, opened 2023-02-24
Implement constant variables

Lyxal:


Lyxal:
We may do a lot of goofy, but PR titles aren't typically goofy things.

GingerIndustries:
💀

Seggan:
... why. There'd literally be no point to this.

Lyxal:
Constants help solidify the prac-lang side of vyxal which is something I want to invest in

Steffan153:
There's plenty of prac-langs without constants

Lyxal:
But the majority of the most popular ones (Java, Scala, C, C++, Rust, C#, JavaScript and more) do.

Constants provide programmer safety, which there isn't much of in a language like this. Every measure counts.

And it may be pointless for golfing, but for longer programs it's the same as using `val` for 40-50% of values in scala.

"But people won't use vyxal for longer programs" maybe, maybe not. I intend for people to be able to do so if they want, and I like having options available to make the prac-lang side something people won't complain about if they use it.

Seggan:
🤷

GingerIndustries (approved):

PR #1646, opened 2023-02-26
Add that SSS string compression algorithm Jelly uses

Lyxal:
Ready to review and merge when satisfied


js/src/main/scala/JSVyxal.scala
```diff
@@ -9,6 +9,10 @@ import scala.scalajs.js.JSConverters.*
 /** A bridge between the interpreter and JS */
 @JSExportTopLevel("Vyxal")
 object JSVyxal:
+
+  var shortDict: Seq[String] = Seq()
+  var longDict: Seq[String] = Seq()
```
ysthakur, line 14:
```suggestion
  @JSExport
  var shortDict: Seq[String] = Seq()
  @JSExport
  var longDict: Seq[String] = Seq()
```

js/src/main/scala/JSVyxal.scala
```diff
@@ -9,6 +9,10 @@ import scala.scalajs.js.JSConverters.*
...
+
   @JSExport
   def execute(
```
ysthakur, line 17: Another way we could do things is by passing in the short and long dictionaries in the `execute` method itself. No idea what the pros/cons of that are

shared/src/main/scala/cli/CLI.scala
```diff
@@ -39,6 +39,16 @@ object CLI:
     OParser.parse(parser, args, CLIConfig()) match
       case Some(config) =>
         val inputList = config.inputs.reverse.map(Parser.parseInput)
+        val shortDict =
+          io.Source.fromFile("resources/ShortDictionary.txt").getLines()
+        val longDict =
+          io.Source.fromFile("resources/LongDictionary.txt").getLines()
```
ysthakur, line 45:
```suggestion
        val shortDict =
          io.Source.fromResource("ShortDictionary.txt").getLines()
        val longDict =
          io.Source.fromResource("LongDictionary.txt").getLines()
```
Files in the `resource/` directory are sorta special, they're bundled into JARs (or native executables, when using Scala Native). Hopefully we can do something like this for Scala.js too

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -1,12 +1,39 @@
 package vyxal
 
 import vyxal.*
+import vyxal.impls.Elements
 
+import scala.annotation.retains
```
ysthakur, line 6:
```suggestion
```

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -1,12 +1,38 @@
...
+      case (n: VNum, _)           => fromBase(b, a)
+      case (a: String, b: String) => fromBaseAlphabet(a, b)
+      case _ => fromBaseDigits(ListHelpers.makeIterable(a), b)
+
+  def fromBaseAlphabet(value: String, alphabet: String): VAny =
+    // Returns value in base 10 using base len(alphabet)
+    // [bijective base]
+    var ret = 0
+    for digit <- value do ret += alphabet.size * ret + alphabet.indexOf(digit)
+    ret
```
ysthakur, line 23:
```suggestion
    value.foldLeft(VNum(0)) { (ret, digit) => alphabet.size * ret + alphabet.indexOf(digit): VNum }
```
In case you like folds more

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -1,12 +1,38 @@
...
+  def fromBaseDigits(digits: VList, base: VAny)(using ctx: Context): VAny =
+    // Returns digits in base 10 using arbitrary base 'base'
+    var ret: VAny = VNum(0)
+    for digit <- digits do
+      ctx.push(digit, ret)
+      Interpreter.execute("×")
+      ctx.push(base, ret)
+      Interpreter.execute("+")
+      ret = ctx.pop()
+    ret
```
ysthakur, line 34:
```suggestion
    digits.foldLeft(VNum(0)) { (ret, digit) =>
      MiscHelpers.add(MiscHelpers.multiply(base, ret), digit)
    }
```
In case you prefer folds

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -1,12 +1,38 @@
...
+
+  def fromBaseAlphabet(value: String, alphabet: String): VAny =
+    // Returns value in base 10 using base len(alphabet)
+    // [bijective base]
+    var ret = 0
+    for digit <- value do ret += alphabet.size * ret + alphabet.indexOf(digit)
+    ret
+
+  def fromBaseDigits(digits: VList, base: VAny)(using ctx: Context): VAny =
+    // Returns digits in base 10 using arbitrary base 'base'
```
ysthakur, line 26:
```suggestion
  /** Returns digits in base 10 using arbitrary base `base` */
  def fromBaseDigits(digits: VList, base: VAny)(using ctx: Context): VAny =
```

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -1,12 +1,38 @@
...
+  def fromBase(a: VAny, b: VAny)(using ctx: Context): VAny =
+    (a, b) match
+      case (a: VNum, b: VNum)     => toInt(a.toString(), b.toInt)
+      case (n: VNum, _)           => fromBase(b, a)
+      case (a: String, b: String) => fromBaseAlphabet(a, b)
+      case _ => fromBaseDigits(ListHelpers.makeIterable(a), b)
+
+  def fromBaseAlphabet(value: String, alphabet: String): VAny =
+    // Returns value in base 10 using base len(alphabet)
+    // [bijective base]
```
ysthakur, line 20:
```suggestion
  /** Returns value in base 10 using base len(alphabet) [bijective base] */
  def fromBaseAlphabet(value: String, alphabet: String): VAny =
```

shared/src/main/scala/StringHelpers.scala
```diff
@@ -15,6 +16,72 @@ object StringHelpers:
...
+        if c.toInt == 10 then 95
+        else if ' ' <= c && c <= '~' then c.toInt - 32
+        else throw new Exception(s"Invalid character $c")
+
+      3 * 96 * z + o
+
+    def dictionary(z: Int, w: String, nonempty: Boolean) =
+      var ts = nonempty
+      var subW = w
+      if w.headOption.exists(_ == ' ') then
```
ysthakur, line 34: Can we assume that `w` is not the empty string? If so, we can do
```suggestion
      if w.charAt(0) == ' ' then
```

shared/src/main/scala/StringHelpers.scala
```diff
@@ -54,4 +124,54 @@ object StringHelpers:
...
+
+  // https://github.com/DennisMitchell/jellylanguage/blob/70c9fd93ab009c05dc396f8cc091f72b212fb188/jelly/interpreter.py#L1055
+  def sss(compressed: String)(using ctx: Context): String =
+    val decompressed = StringBuilder()
+    var integer =
+      NumberHelpers
+        .fromBase(
+          VList.from(
+            compressed.indices.map(x => CODEPAGE.indexOf(compressed(x)))
+          ),
```
ysthakur, line 136:
```suggestion
          VList.from(compressed.map(CODEPAGE.indexOf)),
```

shared/src/main/scala/StringHelpers.scala
```diff
@@ -54,4 +124,54 @@ object StringHelpers:
...
+        val words =
+          if useShort then ctx.globals.shortDictionary
+          else ctx.globals.longDictionary
+        val index = integer % words.length
+        integer = integer / words.length
+        var word = words(index)
+        if flagSwap then
+          word = (
+            word.map(c => if c.isUpper then c.toLower else c.toUpper)
+          )
```
ysthakur, line 169:
```suggestion
          word =
            word.map(c => if c.isUpper then c.toLower else c.toUpper)
```
Do we have a helper for toggling case? We ought to make one

shared/src/main/scala/StringHelpers.scala
Lyxal: I honestly have no clue if we can make that assumption. The python I ported had `w[:1] == " "` as the condition

shared/src/main/scala/StringHelpers.scala
Lyxal: When I tried something like that, it gave a compiler error and complained about having a list of int instead of a list of string

shared/src/main/scala/StringHelpers.scala
Lyxal: Yeah we should make one

shared/src/main/scala/StringHelpers.scala
ysthakur: Should work now

shared/src/main/scala/StringHelpers.scala
ysthakur: Well if we're compressing the empty string, we can probably return early, right? There's no compressing to do if it's already `""`. Not sure where to put that, though

shared/src/main/scala/StringHelpers.scala
```diff
@@ -54,4 +124,52 @@ object StringHelpers:
...
       else mapping((index + 1) % mapping.length)
     }.mkString
+
+  // https://github.com/DennisMitchell/jellylanguage/blob/70c9fd93ab009c05dc396f8cc091f72b212fb188/jelly/interpreter.py#L1055
+  def sss(compressed: String)(using ctx: Context): String =
+    val decompressed = StringBuilder()
+    var integer =
+      NumberHelpers
+        .fromBase(
+          VList.from(compressed.map(CODEPAGE.indexOf(_))),
```
ysthakur, line 134:
```suggestion
          VList.from(compressed.map(CODEPAGE.indexOf(_): VNum)),
```

shared/src/main/scala/StringHelpers.scala
Lyxal: Where's hyper when you need him? :p

shared/src/main/scala/StringHelpers.scala
```diff
@@ -15,6 +16,75 @@ object StringHelpers:
...
+      if f then
+        res = 3 * res + j
+        res = 3 * res + 2
+      else res = 3 * res + 1
+      res
+    end dictionary
+
+    def go(z: Int) =
+      val compressed = StringBuilder()
+      var z1 = z
```
Seggan, line 66: Can't you make the var/method names a bit more... descriptive?

ysthakur: This was ported directly from what Dennis wrote. Someday, we mere mortals will be able to decipher what he wrote. Until then, we'll have to stick with cryptic single-letter names :|

shared/src/main/scala/StringHelpers.scala
```diff
@@ -54,4 +124,52 @@ object StringHelpers:
...
       else mapping((index + 1) % mapping.length)
     }.mkString
+
+  // https://github.com/DennisMitchell/jellylanguage/blob/70c9fd93ab009c05dc396f8cc091f72b212fb188/jelly/interpreter.py#L1055
+  def sss(compressed: String)(using ctx: Context): String =
+    val decompressed = StringBuilder()
+    var integer =
+      NumberHelpers
+        .fromBase(
+          VList.from(compressed.map(CODEPAGE.indexOf(_): VNum)),
```
hyper-neutrino, line 134: No clue how scala works, but is the `+1` not needed here? ([...](https://github.com/DennisMitchell/jellylanguage/blob/70c9fd93ab009c05dc396f8cc091f72b212fb188/jelly/interpreter.py#L1058))

Lyxal: I missed that part of the original. It needs to be added

shared/src/main/scala/StringHelpers.scala
```diff
@@ -15,6 +16,72 @@ object StringHelpers:
...
+
+    val dp = Array.fill(s.length + 1)(0)
+    // scala equivalent of for i in range(len(str) -1,-1,-1)
+    for i <- (s.length - 1) to 0 by -1 do
+      dp(i) = character(dp(i + 1), s(i))
+      for j <- 1 to Math.min(endLength, s.length - i) do
+        try
+          dp(i) = Math.min(
+            dp(i),
+            dictionary(dp(i + j), s.substring(i, i + j), i != 0)
```
Steffan153, line 81: It's fairly easy to tell based on the arguments here that `s.substring(i, i + j)` (`w` in the function) will never be empty.

shared/src/main/scala/StringHelpers.scala
```diff
@@ -15,6 +16,72 @@ object StringHelpers:
...
+          )
+        catch case _: Exception => () // todo (lyxal): is this necessary?
```
Steffan153, line 83: ?

shared/src/main/scala/StringHelpers.scala
```diff
@@ -15,6 +16,75 @@ object StringHelpers:
...
```
Lyxal, line 83: @ysthakur yes it is - without it, words not in the dictionary cause the function to execute with an error. I assume that the errors thrown are for control execution.

Lyxal:
Needs to fix having other string closers in the generated output

Lyxal:
Also tests

Lyxal:
Also also the word lists don't seem to be optimal - eg compressing hello world is a few bytes longer in Vyxal than it is in jelly

shared/src/main/scala/StringHelpers.scala
ysthakur: Then we should check if a word is in the dictionary before accessing it instead of catching all `Exception`s. Silently catching a whole category of exceptions will make debugging quite a bit harder

Lyxal:
Just needs native figuring out

shared/src/main/scala/StringHelpers.scala
Steffan153: `z1` is the number that represents the string, to be converted into our compressed string (you can see the divmod by 250, which is the base 250 compression).

Steffan153: This whole `go` function is just number compression. Could be outsorced to a helper. Also idk our literal syntax but we can probably use a higher base than 250.

Steffan153:
Should we export the whole compression functions to JS or should we manually write those in JS?

Lyxal:
Depends on whether a manual implementation is faster than the exported one

shared/src/main/scala/StringHelpers.scala
Lyxal: I tried changing the base to 252, but it seemed to break compression. Maybe I did something wrong, or maybe the modulos need to change as well

Steffan153: You have to change both 250s, and you might also have to mess with the codepage it's using for indexing, idk

ysthakur:
Wait, why is inporting vyxal.Globals.ShortDictionary necessary to compile? That's kinda wack

ysthakur:
> Should we export the whole compression functions to JS or should we manually write those in JS?

Probably easiest to export to JS. Not much reason to reimplement all that in JS, I think

Lyxal:
> Wait, why is inporting vyxal.Globals.ShortDictionary necessary to compile? That's kinda wack

It's not I don't think. That's just a stray import

GingerIndustries:
<s>hurry up I need my ඞ</s>

pages/index.html
```diff
@@ -31,6 +31,8 @@
   <link rel="stylesheet" href="https://vyxal.github.io/Vylight/mode-vyxal.css">
 
   <script defer src="./vyxal.js"></script>
+  <script defer src="./dictionary.js"></script>
+  </script>
```
ysthakur, line 35:
```suggestion
```
Not sure if this is a stray closing script tag

Lyxal: it is

shared/src/main/scala/StringHelpers.scala
```diff
@@ -54,4 +129,56 @@ object StringHelpers:
       if index == -1 then c
       else mapping((index + 1) % mapping.length)
     }.mkString
+
+  // https://github.com/DennisMitchell/jellylanguage/blob/70c9fd93ab009c05dc396f8cc091f72b212fb188/jelly/interpreter.py#L1055
+  def sss(compressed: String)(using ctx: Context): String =
```
ysthakur, line 134: Might want to either add a doc comment saying that this is for decompression or rename the method itself to `decompress` or something simlar

Lyxal: I'd be fine with renaming it.

PR #1647, opened 2023-03-05
Make `toString` on non-integer `VNum` not just a stringified float

Lyxal:


ysthakur (approved):
Still gonna run into porblems when we exceed the range of a Double but this is good until then

PR #1653, opened 2023-03-07
Use the provided method for real to string

Lyxal:
So turns out there was a method provided by spire this whole time!


shared/src/main/scala/VNum.scala
```diff
@@ -30,11 +30,8 @@ class VNum private (val underlying: Complex[Real]):
   def vabs: VNum = underlying.abs
 
   override def toString =
-    def toStringHelper(n: Real): String =
-      if n.isWhole then n.toString
-      else n.floor.toString + "." + (n - n.floor).toDouble.toString().substring(2)
-    if this.imag == 0 then toStringHelper(this.real)
-    else toStringHelper(this.real) + "ı" + toStringHelper(this.imag)
+    if this.imag == 0 then this.real.getString(40)
```
ysthakur, line 33: Hold on a sec, isn't `Real's` [`toString`](https://github.com/typelevel/spire/blob/main/core/src/main/scala/spire/math/Real.scala#L291) essentially doing the same thing as `getString(40)`? How come we get fractions when doing `toString` but not with `getString`?

Lyxal: that's a real good question

Lyxal:
```scala
  override def toString: String = x match {
    case Exact(n) => n.toString
    case _        => getString(Real.digits)
  }
```

> `Exact(n)`

We get fractions because it's returning them for exact numbers.

PR #1655, opened 2023-03-07
Make input able to be evaluated as vyxal

Lyxal:


tests/test_interpreter.py
```diff
@@ -0,0 +1,33 @@
+"""For testing flags"""
+
+import multiprocessing
+import os
+import sys
+import sympy
```
Steffan153, line 6: ?

tests/test_interpreter.py
```diff
@@ -0,0 +1,33 @@
+"""For testing flags"""
+
+import multiprocessing
```
Steffan153, line 3: ?

PR #1658, opened 2023-03-08
Create pull_request_template.md

GingerIndustries:
So sussy bakas like don't forgor 💀 to link our issues.


Lyxal (approved):
Fun fake fact: lgtm stands for Lyxal golfing trolled me (that's what I decided it stands for ever since lgtm was shut down)

PR #1663, opened 2023-03-08
Improve sort script

Steffan153:
It was quite unpythonic, so I cleaned it up and also made the JS file be way shorter.


Steffan153:
pre-commit.ci run

GingerIndustries (approved):
mmm pymthon

PR #1665, opened 2023-03-09
10k/50k Word Lists for Dictionaries

Lyxal:
5k just wasn't long enough for the short, and 50k was the best for long with 10k short.


dict_scripts/sss_compress.py
```diff
@@ -0,0 +1,80 @@
...
+                    dp[i] = min(
+                        dp[i],
+                        dictionary(dp[i + j], str[i : i + j], i != 0),
+                    )
+                except ValueError:
+                    pass
+
+        return min(
+            [
+                "“{0}”".format(str),
```
mathcat4, line 74:
```suggestion
        return '"{0}”'.format("".join(compressed[::-1]))

    def optimal(str):
        dp = [0] * -~len(str)
        for i in range(len(str) - 1, -1, -1):
            dp[i] = character(dp[i + 1], str[i])
            for j in range(1, min(len(dp) - i, end_len)):
                try:
                    dp[i] = min(
                        dp[i],
                        dictionary(dp[i + j], str[i : i + j], i != 0),
                    )
                except ValueError:
                    pass

        return min(
            [
                '"{0}"'.format(str),
```

wrong apostrophes?

mathcat4 (changes requested):

dict_scripts/sss_compress.py
Steffan153: No, this is Jelly's string delimiters.

dict_scripts/sss_compress.py
```diff
@@ -0,0 +1,80 @@
...
+        compressed = []
+        while z:
+            z, c = divmod(
+                z - 1, 252
+            )  # Taken inspiration from Jelly's bijective base conversion algorithm
+            compressed.append(code_page[c])
+        return "“{0}»".format("".join(compressed[::-1]))
+
+    def optimal(str):
+        dp = [0] * -~len(str)
```
Steffan153, line 60: Come on, this isn't code golf

Steffan153:
```suggestion
        dp = [0] * (len(str) + 1)
```

dict_scripts/sss_compress.py
Lyxal: This is only to get length under compression. The strings produced are irrelevant.

dict_scripts/sss_compress.py
Lyxal: Don't blame me - blame the script I found on a code golf answer.

mathcat4 (approved):
lgtm

dict_scripts/sss_compress.py
```diff
@@ -0,0 +1,80 @@
...
+                        dictionary(dp[i + j], str[i : i + j], i != 0),
+                    )
+                except ValueError:
+                    pass
+
+        return min(
+            [
+                "“{0}”".format(str),
+                go(dp[0]),
+            ],
```
Steffan153, line 76: You don't need the square brackets here

PR #1666, opened 2023-03-09
Add sugar trigraphs for real

Lyxal:


Lyxal:
At some stage, this needs to automatically print out what the converted program is if it uses sugar trigraphs. That can probably go in the apply section with a boolean set to true if a sugar trigraph is ever processed

Lyxal:
Also as a general naming rule thing:

- `#.<letter>` is the updot version of that letter if it exists
- `#,<letter>` is the downdot version of that letter if it exists
- `#^<letter>` is a superscript version of a letter if it exists
- `#.<symbol>` is what one would expect a trigraph to logically turn into
- `#,<symbol>` is what one would expect a trigraph to aesthetically turn into. Alternatively, if it has an underbar, use that.
`#^<symbol>` is superscript version if it exists. Otherwise, it's a sort of miscellaneous namespace.
`#^.`, `#^:` and `#^%` were chosen for the single, double and triple element grouping modifiers respectively because they have one, two and three circles/lines respectively
L

Lyxal:
Also, trigraphs are subject to change. If any of y'all feel like there's something more suitable for a Unicode, feel free to suggest it.

Steffan153:
We should add tab completion to convert these to their unicode too. Except then maybe the `#` shouldn't be required.

shared/src/main/scala/SugarMap.scala
```diff
@@ -0,0 +1,46 @@
+package vyxal
+object SugarMap:
+  def internalMap: Map[String, String] =
+    val out = Map.newBuilder[String, String]
+
+    val lowerUpdot = "abcdefghlmnoprsx" -> "ȧḃċḋėḟġḣŀṁṅȯṗṙṡṫẋ"
+    val upperUpdot = "ABCDEFGHLMNOPRSTWX" -> "ȦḂĊḊĖḞĠḢİĿṀṄȮṖṘṠṪẆẊ"
+    val upperDowndot = "BDHILMNORST" -> "ḄḌḤỊḶṂṆỌṚṢṬ"
```
mathcat4, line 8:
```suggestion
    val lowerUpdot = "abcdefghlmnoprstx" -> "ȧḃċḋėḟġḣŀṁṅȯṗṙṡṫẋ"
    val upperUpdot = "ABCDEFGHILMNOPRSTWX" -> "ȦḂĊḊĖḞĠḢİĿṀṄȮṖṘṠṪẆẊ"
    val upperDowndot = "BDHILMNORST" -> "ḄḌḤỊḶṂṆỌṚṢṬ"
```
well this doesn't seem right

mathcat4 (changes requested):

mathcat4 (approved):

PR #1669, opened 2023-03-10
Sugar trigraphs (take 2)

Lyxal:


Lyxal:
Also as a general naming rule thing:

- `#.<letter>` is the updot version of that letter if it exists
- `#,<letter>` is the downdot version of that letter if it exists
- `#^<letter>` is a superscript version of a letter if it exists
- `#.<symbol>` is what one would expect a trigraph to logically turn into
- `#,<symbol>` is what one would expect a trigraph to aesthetically turn into. Alternatively, if it has an underbar, use that.
- `#^<symbol>` is superscript version if it exists. Otherwise, it's a sort of miscellaneous namespace.
- `#^.`, `#^:` and `#^%` were chosen for the single, double and triple element grouping modifiers respectively because they have one, two and three circles/lines respectively
- `#,<number>` is the subscript version of that number.

shared/src/main/scala/AST.scala
```diff
@@ -56,7 +56,18 @@ enum AST(val arity: Option[Int]):
...
     case Lst(elems)      => elems.map(_.toVyxal).mkString("#[", "|", "#]")
     case Command(value)  => value
-    case Group(elems, _) => elems.map(_.toVyxal).mkString
+    case Group(elems, _) =>
+      // replace instances of Number, Number with Number, Space, Number
+      elems
+        .groupBy(_ match
+          case Number(_) => true
+          case _         => false
+        )
```
ysthakur, line 65:
```suggestion
        .groupBy {
          case Number(_) => true
          case _         => false
        }
```
Or just `.groupBy(_.isInstanceOf[AST.Number])`

shared/src/main/scala/AST.scala
```diff
@@ -56,7 +56,18 @@ enum AST(val arity: Option[Int]):
...
+        .map((k, v) =>
+          if k then v.map(_.toVyxal).mkString(" ")
+          else v.map(_.toVyxal).mkString
+        )
```
ysthakur, line 69:
```suggestion
        .map { (k, v) =>
          if k then v.map(_.toVyxal).mkString(" ")
          else v.map(_.toVyxal).mkString
        }
```

Lyxal: I blame copilot for that :p

shared/src/main/scala/Lexer.scala
```diff
@@ -73,7 +74,7 @@ val SPECIAL_MODIFIERS = "ᵗᵜ"
 object Lexer extends RegexParsers:
   override def skipWhitespace = true
   override val whiteSpace: Regex = "[ \t\r\f]+".r
-
+  var sugarUsed = false
```
ysthakur, line 77:
```suggestion
  /** Whether the code lexed so far has sugar trigraphs */
  private var sugarUsed = false
```

shared/src/main/scala/SugarMap.scala
```diff
@@ -0,0 +1,47 @@
...
+    for i <- 0 until dotExtra._1.length do
+      out += s"#.${dotExtra._1(i).toString}" -> dotExtra._2(i).toString
+
+    out.result()
+  end internalMap
+
+  def apply(trigraph: String): String =
+    internalMap.get(trigraph) match
+      case Some(value) => value
+      case None        => trigraph
```
ysthakur, line 46:
```suggestion
    internalMap.getOrElse(trigraph, trigraph)
```

shared/src/main/scala/SugarMap.scala
```diff
@@ -0,0 +1,47 @@
+package vyxal
+object SugarMap:
+  def internalMap: Map[String, String] =
```
ysthakur, line 3: Should this be a val?

Lyxal: I wasn't sure if it should be a def or a val. If it can be a val and not impact anything, then sure

shared/src/main/scala/SugarMap.scala
```diff
@@ -0,0 +1,45 @@
...
```
ysthakur, line 3:
```suggestion
  val internalMap: Map[String, String] =
```

ysthakur (approved):

PR #1671, opened 2023-03-14
Make the dictionary not error in JS on a fresh context

Lyxal:

### DIFF
--- a/shared/src/main/scala/Dictionary.scala
+++ b/shared/src/main/scala/Dictionary.scala
@@ -1,0 +1,10 @@
+package vyxal
+object Dictionary:
+  var shortDictionary: Seq[String] = Seq()
+  var longDictionary: Seq[String] = Seq()
+  var initalised: Boolean = false
+
+  def initalise(short: Seq[String], long: Seq[String]): Unit =
+    shortDictionary = short
+    longDictionary = long
+    initalised = true
--- a/shared/src/main/scala/Interpreter.scala
+++ b/shared/src/main/scala/Interpreter.scala
@@ -12,6 +12,7 @@ object Interpreter:
   def execute(code: String, literate: Boolean = false)(using
       ctx: Context
   ): Unit =
+    if !Dictionary.initalised then throw new Error("Dictionary not initalised")
     val sbcsified = if literate then LiterateLexer.litLex(code) else code
     Parser.parse(sbcsified) match
       case Right(ast) =>
@@ -23,6 +24,7 @@ object Interpreter:
         then vyPrintln(ctx.peek)
       case Left(error) =>
         throw new Error(s"Error while executing $code: $error")
+  end execute
 
   def execute(ast: AST)(using ctx: Context): Unit =
     if ctx.settings.logLevel == LogLevel.Debug then
--- a/js/src/main/scala/JSVyxal.scala
+++ b/js/src/main/scala/JSVyxal.scala
@@ -28,9 +28,9 @@ object JSVyxal:
       settings = settings,
       printFn = printFunc,
       inputs = Inputs(inputs.split("\n").map(Parser.parseInput).toSeq),
-      shortDictionary = shortDict,
-      longDictionary = longDict
     )
+
+    Dictionary.initalise(shortDict, longDict)
 
     val ctx = Context(
       inputs = inputs.split("\n").map(Parser.parseInput).toIndexedSeq,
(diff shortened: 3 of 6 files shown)


shared/src/main/scala/Interpreter.scala
```diff
@@ -23,6 +24,7 @@ object Interpreter:
         then vyPrintln(ctx.peek)
       case Left(error) =>
         throw new Error(s"Error while executing $code: $error")
+  end execute
```
Steffan153, line 27: Is this necessary?

Steffan153 (approved):

shared/src/main/scala/Interpreter.scala
Lyxal: Scala linter. It just puts it there for stylistic consistency

shared/src/main/scala/Dictionary.scala
```diff
@@ -0,0 +1,10 @@
+package vyxal
+object Dictionary:
+  var shortDictionary: Seq[String] = Seq()
+  var longDictionary: Seq[String] = Seq()
+  var initalised: Boolean = false
+
+  def initalise(short: Seq[String], long: Seq[String]): Unit =
```
ysthakur, line 7: This should be `initialize`. Yes, with a "z." It is my right as an American citizen to not have to use your British "English." I will use "z" and put my periods inside my quotes and you can't stop me.

shared/src/main/scala/Dictionary.scala
```diff
@@ -0,0 +1,10 @@
+package vyxal
+object Dictionary:
+  var shortDictionary: Seq[String] = Seq()
+  var longDictionary: Seq[String] = Seq()
+  var initalised: Boolean = false
```
ysthakur, line 5: This should be `initialised`

shared/src/main/scala/Dictionary.scala
```diff
@@ -0,0 +1,10 @@
+package vyxal
+object Dictionary:
```
ysthakur, line 2:
```suggestion

private[vyxal] object Dictionary:
```
Not a huge fan of this being able to be modified by everyone. I'd have preferred platform-specific implementations, but if we're not doing that, let's at least make this private to the `vyxal` package.

shared/src/main/scala/Dictionary.scala
ysthakur, line 5:
```suggestion
  private var shortDict: Seq[String] = Seq()
  private var longDict: Seq[String] = Seq()
  private var initialised: Boolean = false

  def shortDictionary: Seq[String] = shortDict
  def longDictionary: Seq[String] = longDict
```

native/src/main/scala/Main.scala
```diff
@@ -4,4 +4,18 @@ import vyxal.cli.CLI
...
 object Main:
   def main(args: Array[String]): Unit =
+
+    Dictionary.initalise(
+      readResource(ShortDictionaryFile),
+      readResource(LongDictionaryFile)
+    )
     CLI.run(args)
+
+  private val ShortDictionaryFile = "/ShortDictionary.txt"
```
ysthakur, line 14: I wonder if we can avoid repeating this in the JVM folder, the native folder, and the shared test folder. Perhaps we could move all the shared tests into the JVM folder, so that it'll only be repeated in two places

ysthakur (changes requested):

shared/src/main/scala/Dictionary.scala
Lyxal: Skill issue + Australian english + resolved conversation + not changed

shared/src/main/scala/Dictionary.scala
ysthakur: Wait actually don't accept this suggestion directly, because the `initialise` method would need to be changed to use `shortDict` and `longDict`

shared/src/main/scala/Dictionary.scala
Lyxal: Platform specific implementations was causing too much trouble. A singleton object that's shared between objects with settable dictionaries is much easier.

Lyxal: Also, does this mean only vyxal files can modify the dictionary?

shared/src/main/scala/Dictionary.scala
ysthakur: I was joking about the "z," but you really do need to change the spelling

shared/src/main/scala/Dictionary.scala
ysthakur: Yup

Lyxal: Who's going to be using Vyxal as a library anyway? Most of the helper functions require context, and that just seems silly when context is vyxal specific.

ysthakur: ¯\_(ツ)_/¯

shared/src/main/scala/Dictionary.scala
Steffan153: Are we really that afraid of overwriting shortDict and longDict?

ysthakur: I am afraid of everything

native/src/main/scala/Main.scala
Lyxal: What if there was an `initOffline` and `initOnline` method in dictionary and jvm/native and js could call them respectively

ysthakur (approved):

PR #1674, opened 2023-03-17
Implement H

Lyxal:

### DIFF
--- a/shared/src/main/scala/NumberHelpers.scala
+++ b/shared/src/main/scala/NumberHelpers.scala
@@ -153,7 +153,7 @@ object NumberHelpers:
           if digit < 0 then
             current += 1
             digit -= base.toBigInt
-          VNum(digit * sign) +=: digits
+          digits.prepend(digit * sign)
         VList(digits.toList*)
 
   def toInt(value: VAny, radix: Int)(using ctx: Context): VAny =
--- a/shared/src/main/scala/NumberHelpers.scala
+++ b/shared/src/main/scala/NumberHelpers.scala
@@ -124,10 +124,10 @@ object NumberHelpers:
   /** Returns value in base len(alphabet) using base 10 [bijective base] */
   def toBaseAlphabet(value: VNum, alphabet: VIter)(using ctx: Context): VAny =
     val indexes = toBaseDigits(value, alphabet.iterLength)
-    val alphalist = VList((alphabet match
+    val alphalist = VList.from(alphabet match
       case a: String => a.toString.toList.map(_.toString)
       case l: VList  => l
-    )*)
+    )
 
     val temp = indexes.map(alphalist.index(_).toString())
     alphabet match
--- a/shared/src/main/scala/NumberHelpers.scala
+++ b/shared/src/main/scala/NumberHelpers.scala
@@ -146,7 +146,7 @@ object NumberHelpers:
       var current = sign * value.toBigInt
       if base.toBigInt == 1 then VList(List.fill(current.toInt.abs)(sign)*)
       else
-        val digits = ListBuffer[VAny]()
+        val digits = ListBuffer[VNum]()
         while current > 0 do
           var digit = current % base.toBigInt
           current = current / base.toBigInt
--- a/shared/src/main/scala/NumberHelpers.scala
+++ b/shared/src/main/scala/NumberHelpers.scala
@@ -3,6 +3,7 @@ package vyxal
 import vyxal.*
 import vyxal.impls.Elements
 
+import scala.collection.mutable
 import scala.collection.mutable.ListBuffer
 import scala.math
 
@@ -102,6 +103,58 @@ object NumberHelpers:
           result += VList(binary.map(_.asDigit).map(VNum(_)).toList*)
         VList(result.toList*)
       case _ => throw new Exception("Cannot convert to binary")
+
+  def toBase(a: VAny, b: VAny)(using ctx: Context): VAny =
+    (a, b) match
+      case (a: VNum, b: VNum) =>
+        if b.toBigInt == 0 then 0
+        else toBaseDigits(a, b)
+      case (n: VNum, _)  => toBaseAlphabet(n, ListHelpers.makeIterable(b))
+      case (a: VList, _) => VList(a.map(toBase(_, b))*)
+      case _ => throw new Exception("toBase only works on numbers and lists")
+
+  /** Returns value in base len(alphabet) using base 10 [bijective base] */
+  def toBaseAlphabet(value: VNum, alphabet: VIter)(using ctx: Context): VAny =
+    val indexes = toBaseDigits(value, alphabet.iterLength)
+    val alphalist = VList((alphabet match
+      case a: String => a.toString.toList.map(_.toString())
+      case l: VList  => l
+    )*)
+
+    val temp = indexes.map(alphalist.index(_).toString())
+    alphabet match
+      case a: String => temp.mkString("")
+      case l: VList  => l
+
+  def toBaseDigits(value: VNum, base: VNum): VList =
+    if value == VNum(0) then VList(List(VNum(0))*)
+    else if base.toBigInt == -1 then
+      val out = ListBuffer[VNum]()
+      if value.toBigInt > 0 then
+        out += 1
+        for _ <- 1 until value.toInt do
+          out += 0
+          out += 1
+      else
+        val out = ListBuffer[VNum]()
+        for _ <- 1 until value.toInt.abs do
+          out += 1
+          out += 0
+      VList(out.toList*)
+    else
+      val sign = if value.toBigInt < 0 && base.toBigInt > 0 then -1 else 1
+      var current = sign * value.toBigInt
+      if base.toBigInt == 1 then VList(List.fill(current.toInt.abs)(sign)*)
+      else
+        val digits = ListBuffer[VAny]()
+        while current > 0 do
+          var digit = current % base.toBigInt
+          current = current / base.toBigInt
+          if digit < 0 then
+            current += 1
+            digit = digit - base.toBigInt
+          digits += VNum(digit * sign)
+        VList(digits.reverse.toList*)
 
   def toInt(value: VAny, radix: Int)(using ctx: Context): VAny =
     value match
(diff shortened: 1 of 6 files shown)


shared/src/main/scala/VList.scala
```diff
@@ -56,7 +56,9 @@ class VList private (val lst: Seq[VAny])
       lst(Math.floorMod(ind, lst.length))
     else
       try lst(ind)
-      catch case _: IndexOutOfBoundsException => lst(ind % lst.length)
+      catch
+        case _: IndexOutOfBoundsException =>
+          if lst.isEmpty then 0 else lst(ind % lst.length)
```
ysthakur, line 61:
```scala
    if lst.isEmpty then 0
    else if ind < 0 then
      // floorMod because % gives negative results with negative dividends
      lst(Math.floorMod(ind, lst.length))
    else
      try lst(ind)
      catch
        case _: IndexOutOfBoundsException => lst(ind % lst.length)
```
I think the `if lst.isEmpty` check should go at the top in case the index is negative but the list is empty

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -103,6 +104,58 @@ object NumberHelpers:
...
+      case (a: VList, _) => VList(a.map(toBase(_, b))*)
+      case _ => throw new Exception("toBase only works on numbers and lists")
+
+  /** Returns value in base len(alphabet) using base 10 [bijective base] */
+  def toBaseAlphabet(value: VNum, alphabet: VIter)(using ctx: Context): VAny =
+    val indexes = toBaseDigits(value, alphabet.iterLength)
+    val alphalist = VList((alphabet match
+      case a: String => a.toString.toList.map(_.toString())
+      case l: VList  => l
+    )*)
```
ysthakur, line 122:
```suggestion
    val alphalist = VList.from(alphabet match
      case a: String => a.toList.map(_.toString())
      case l: VList  => l
    )
```

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -103,6 +104,58 @@ object NumberHelpers:
...
+    if value == VNum(0) then VList(List(VNum(0))*)
+    else if base.toBigInt == -1 then
+      val out = ListBuffer[VNum]()
+      if value.toBigInt > 0 then
+        out += 1
+        for _ <- 1 until value.toInt do
+          out += 0
+          out += 1
+      else
+        val out = ListBuffer[VNum]()
```
Steffan153, line 139:
```suggestion
```

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -103,6 +104,58 @@ object NumberHelpers:
...
+        else toBaseDigits(a, b)
+      case (n: VNum, _)  => toBaseAlphabet(n, ListHelpers.makeIterable(b))
+      case (a: VList, _) => VList(a.map(toBase(_, b))*)
+      case _ => throw new Exception("toBase only works on numbers and lists")
+
+  /** Returns value in base len(alphabet) using base 10 [bijective base] */
+  def toBaseAlphabet(value: VNum, alphabet: VIter)(using ctx: Context): VAny =
+    val indexes = toBaseDigits(value, alphabet.iterLength)
+    val alphalist = VList((alphabet match
+      case a: String => a.toString.toList.map(_.toString())
```
Steffan153, line 120:
```suggestion
      case a: String => a.toString.toList.map(_.toString)
```

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -103,6 +104,58 @@ object NumberHelpers:
...
+        out += 1
+        for _ <- 1 until value.toInt do
+          out += 0
+          out += 1
+      else
+        val out = ListBuffer[VNum]()
+        for _ <- 1 until value.toInt.abs do
+          out += 1
+          out += 0
+      VList(out.toList*)
```
Steffan153, line 143:
```suggestion
      VList.from((0 to value.abs * 2 - (if value > 0 then 2 else 3)).map(1 - _ % 2))
```

Something like this.

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -103,6 +104,58 @@ object NumberHelpers:
...
+      var current = sign * value.toBigInt
+      if base.toBigInt == 1 then VList(List.fill(current.toInt.abs)(sign)*)
+      else
+        val digits = ListBuffer[VAny]()
+        while current > 0 do
+          var digit = current % base.toBigInt
+          current = current / base.toBigInt
+          if digit < 0 then
+            current += 1
+            digit = digit - base.toBigInt
```
Steffan153, line 155:
```suggestion
            digit -= base.toBigInt
```

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -103,6 +104,58 @@ object NumberHelpers:
...
+          digits += VNum(digit * sign)
+        VList(digits.reverse.toList*)
```
Steffan153, line 157:
```suggestion
          VNum(digit * sign) +=: digits
        VList(digits.toList*)
```

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -103,6 +104,58 @@ object NumberHelpers:
...
+        for _ <- 1 until value.toInt.abs do
+          out += 1
+          out += 0
+      VList(out.toList*)
+    else
+      val sign = if value.toBigInt < 0 && base.toBigInt > 0 then -1 else 1
+      var current = sign * value.toBigInt
+      if base.toBigInt == 1 then VList(List.fill(current.toInt.abs)(sign)*)
+      else
+        val digits = ListBuffer[VAny]()
```
Steffan153, line 149:
```suggestion
        val digits = ListBuffer[VNum]()
```

shared/src/test/scala/ElementTests.scala
```diff
@@ -354,6 +354,14 @@ class ElementTests extends VyxalTests:
     }
   }
 
+  describe("Element H") {
+    testMulti("H")(
+      List[VAny](VNum(69420)) -> "10F2C",
+      List[VAny](VNum(0)) -> "0",
+      List[VAny](VNum(89)) -> "59"
```
Steffan153, line 361: Might want to add some reverse tests here (string => num)

shared/src/main/scala/NumberHelpers.scala
ysthakur: Nice. Might want to make a separate variable to hold the end for the range but still much shorter and clearer

shared/src/main/scala/NumberHelpers.scala
ysthakur: If we're prepending, we might as well do `var digits = List.empty[VNum]` and then do `digits = VNum(digit * sign) :: digits`, so at the end we can do `VList.from(digits)`

Steffan153: Is that as efficient as using a mutable list?

Steffan153: Okay, I did a bad benchmark and using a List is faster than a MutableList
https://scastie.scala-lang.org/U3UMcxCyTEGBaQBCin4xRg

ysthakur: I figured it'd be more efficient since you don't have to create a new list at the end, it just gets built up normally (prepending is an O(1) operation on Lists). Efficiency doesn't matter too much here though, imo

ysthakur: Hmm, I'm consistently getting the second ListBuffer approach to be faster than the List approach. Perhaps we should stick to ListBuffer then lol

shared/src/main/scala/NumberHelpers.scala
Lyxal: For some reason, when I tried this, it didn't want to compile

shared/src/test/scala/ElementTests.scala
Lyxal: Can't believe I forgot those

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -137,32 +137,24 @@ object NumberHelpers:
...
         while current > 0 do
           var digit = current % base.toBigInt
           current = current / base.toBigInt
           if digit < 0 then
             current += 1
-            digit = digit - base.toBigInt
-          digits += VNum(digit * sign)
-        VList(digits.reverse.toList*)
+            digit -= base.toBigInt
+          VNum(digit * sign) +=: digits
```
ysthakur, line 156:
```suggestion
          digits.prepend(digit * sign)
```
Can we use `prepend` here instead? It's easy to guess what `+=:` but someone new to Scala might have to look it up

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -111,6 +112,50 @@ object NumberHelpers:
         VList(result.toList*)
       case _ => throw new Exception("Cannot convert to binary")
 
+  def toBase(a: VAny, b: VAny)(using ctx: Context): VAny =
+    (a, b) match
+      case (a: VNum, b: VNum) =>
+        if b.toBigInt == 0 then 0
```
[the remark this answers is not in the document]
Lyxal: "Values of types vyxal.VNum and Int cannot be compared with == or !="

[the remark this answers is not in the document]
Lyxal: Apply that in #1699

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -111,6 +112,50 @@ object NumberHelpers:
...
+        else toBaseDigits(a, b)
+      case (n: VNum, _)  => toBaseAlphabet(n, ListHelpers.makeIterable(b))
+      case (a: VList, _) => VList(a.map(toBase(_, b))*)
+      case _ => throw new Exception("toBase only works on numbers and lists")
```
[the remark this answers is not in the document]
Lyxal: Applied in #1699

shared/src/main/scala/NumberHelpers.scala
```diff
@@ -111,6 +112,50 @@ object NumberHelpers:
...
+
+  /** Returns value in base len(alphabet) using base 10 [bijective base] */
+  def toBaseAlphabet(value: VNum, alphabet: VIter)(using ctx: Context): VAny =
+    val indexes = toBaseDigits(value, alphabet.iterLength)
+    val alphalist = VList.from(alphabet match
+      case a: String => a.toString.toList.map(_.toString)
+      case l: VList  => l
+    )
```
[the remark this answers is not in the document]
Lyxal: Applied in #1699

PR #1675, opened 2023-03-18
Implement the `I` element in version 3

Lyxal:

### DIFF
--- a/shared/src/main/scala/Elements.scala
+++ b/shared/src/main/scala/Elements.scala
@@ -459,7 +459,7 @@ object Elements:
         ListHelpers.makeIterable(a),
         ListHelpers.makeIterable(b)
       )
-      if a.isInstanceOf[String] && b.isInstanceOf[String] then temp.mkString("")
+      if a.isInstanceOf[String] && b.isInstanceOf[String] then temp.mkString
       else temp
     }
 
--- a/shared/src/main/scala/Elements.scala
+++ b/shared/src/main/scala/Elements.scala
@@ -448,6 +448,21 @@ object Elements:
       case (a: VFun, b) => MiscHelpers.collectUnique(a, b)
     }
 
+    val interleave: Dyad = addElem(
+      Dyad,
+      "I",
+      "Interleave",
+      List("interleave"),
+      "a: lst, b: lst -> Interleave a and b"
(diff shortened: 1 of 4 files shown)


shared/src/main/scala/Elements.scala
```diff
@@ -448,6 +448,21 @@ object Elements:
...
+      "I",
+      "Interleave",
+      List("interleave"),
+      "a: lst, b: lst -> Interleave a and b"
+    ) { case (a, b) =>
+      val temp = ListHelpers.interleave(
+        ListHelpers.makeIterable(a),
+        ListHelpers.makeIterable(b)
+      )
+      if a.isInstanceOf[String] && b.isInstanceOf[String] then temp.mkString("")
```
ysthakur, line 462:
```suggestion
      if a.isInstanceOf[String] && b.isInstanceOf[String] then temp.mkString
```
`mkString` uses `""` by default

shared/src/main/scala/ListHelpers.scala
```diff
@@ -39,6 +39,23 @@ object ListHelpers:
...
+    var mutLeft = left.toList
+    var mutRight = right.toList
+    while mutLeft.nonEmpty && mutRight.nonEmpty do
+      out += mutLeft.head
+      out += mutRight.head
+      mutLeft = mutLeft.tail
+      mutRight = mutRight.tail
+
+    out ++= mutLeft
+    out ++= mutRight
```
ysthakur, line 53:
```suggestion
    val leftIter = left.iterator
    val rightIter = right.iterator
    while leftIter.hasNext && rightIter.hasNext do
      out += leftIter.next()
      out += rightIter.next()

    out ++= leftIter
    out ++= rightIter
```
I'm not sure if those last two lines will still work with an iterator, but they can definitely be modified a bit to work

shared/src/main/scala/Elements.scala
Lyxal: I blame copilot

shared/src/main/scala/ListHelpers.scala
Lyxal: Once again, I blame copilot

ysthakur (approved):

PR #1676, opened 2023-03-18
Implement J element for version 3

Lyxal:

### DIFF
--- a/shared/src/main/scala/Elements.scala
+++ b/shared/src/main/scala/Elements.scala
@@ -486,9 +486,9 @@ object Elements:
     ) {
       case (a: VNum, b: VNum)   => MiscHelpers.eval(a.toString + b.toString)
       case (a: VVal, b: VVal)   => MiscHelpers.add(a, b)
-      case (a: VVal, b: VList)  => VList.from(a +: b)
-      case (a: VList, b: VVal)  => VList.from(a :+ b)
       case (a: VList, b: VList) => VList.from(a ++ b)
+      case (a, b: VList)  => VList.from(a +: b)
+      case (a: VList, b)  => VList.from(a :+ b)
     }
 
     val modulo: Dyad = addElem(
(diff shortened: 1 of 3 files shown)


Steffan153 (approved):

shared/src/main/scala/Elements.scala
```diff
@@ -477,6 +477,20 @@ object Elements:
...
 
+    val merge: Dyad = addElem(
+      Dyad,
+      "J",
+      "Merge",
+      List("merge"),
+      "a: lst, b: lst -> Merge a and b",
+    ) {
+      case (a: VNum, b: VNum)   => MiscHelpers.eval(a.toString + b.toString)
+      case (a: VVal, b: VVal)   => MiscHelpers.add(a, b)
```
ysthakur, line 488:
```suggestion
      case (a: String, b: String)   => MiscHelpers.add(a, b)
```
Since `(VNum, VNum)` has already been matched on. Are we sure we don't want to overload this to something else? String concatenation already exists elsewhere

shared/src/main/scala/Elements.scala
```diff
@@ -477,6 +477,20 @@ object Elements:
...
+      case (a: VVal, b: VList)  => VList.from(a +: b)
+      case (a: VList, b: VVal)  => VList.from(a :+ b)
+      case (a: VList, b: VList) => VList.from(a ++ b)
```
ysthakur, line 491:
```suggestion
      case (a: VList, b: VList) => VList.from(a ++ b)
      case (a, b: VList)  => VList.from(a +: b)
      case (a: VList, b)  => VList.from(a :+ b)
```
So that this works for functions too

shared/src/main/scala/Elements.scala
Lyxal: This is for the `VNum`, `String` case as well though. It's simply doing version 2 behaviour

ysthakur: Oh, I see.

ysthakur (approved):
I wish the `(num, str)`/`(str, num)` case were overloaded to something other than `+` but LGTM

PR #1677, opened 2023-03-18
Implement K for version 3

Lyxal:


Steffan153 (approved):
Not sure if this needs to be vectorizing

ysthakur (approved):
Vectorizing makes sense. LGTM

PR #1679, opened 2023-03-19
Make Ǔ/ǔ modular

### LINKED ISSUE #1678: Enhancement: Rotation (`Ǔ` / `ǔ`) should be modular
See [this program](https://vyxal.pythonanywhere.com/#WyIiLCIiLCLHlCIsIiIsIjVcbkFCQ0QiXQ==). It's trying to rotate `ABCD` right 5 times, but since `5 > len('ABCD')`, it just doesn't do anything. Instead, rotation should be modular, and rotating `ABCD` right 5 times should result in `DABC`.

ysthakur:
Fixes #1678

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -5493,7 +5493,8 @@ def rotate_left(lhs, rhs, ctx):
         return lhs
 
     if ts is str:
-        return lhs[rhs:] + lhs[:rhs]
+        rot = rhs % len(lhs)
+        return lhs[rot:] + lhs[:rot]
 
     @lazylist
     def gen():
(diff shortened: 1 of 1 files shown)


nayakrujul:
Not sure but you might need to check if `lhs` is not empty, otherwise you might get `ModuloByZero` error

ysthakur:
Ah, I didn't think of that. Looks like whoever implemented this already thought of that, though, and added a check a couple lines above, so it's probably safe.

nayakrujul (approved):

chunkybanana (approved):

Steffan153 (approved):

PR #1680, opened 2023-03-19
Add element S

Steffan153:


shared/src/main/scala/Elements.scala
```diff
@@ -623,6 +623,24 @@ object Elements:
...
+      "Sort ascending",
+      List(
+        "sort",
+        "sortasc",
+        "sort-asc",
+      ),
+      "a: any -> convert to list and sort ascending",
+    ) {
+      // should do something else for num overload later
+      case (a: VAny) =>
```
ysthakur, line 638: `(a: VAny)` can be replaced with `a`, since we only deal with `VAny`s, and since that would become `case a =>`, we don't even need pattern matching, so `addFull` can be used and this line can become `a =>`

shared/src/main/scala/Elements.scala
```diff
@@ -623,6 +623,24 @@ object Elements:
...
+        VList(
+          ListHelpers.makeIterable(a).sorted(MiscHelpers.compareExact(_, _))*
+        )
```
ysthakur, line 641: We should decide on a style for this sort of thing. Should `VList.from(x)` be preferred or `VList(x*)`? `VList.from` is more verbose, but I like it more because I'm not sure if `*` creates a new `ArraySeq` or something. If it does, it might not only be bad for performance but also cause an infinite lazy list to be evaluated. Gonna have to look this up.

shared/src/main/scala/Elements.scala
```diff
@@ -623,6 +623,24 @@ object Elements:
...
+      "S",
+      "Sort ascending",
+      List(
+        "sort",
+        "sortasc",
+        "sort-asc",
+      ),
+      "a: any -> convert to list and sort ascending",
+    ) {
+      // should do something else for num overload later
```
Lyxal, line 637: I think sort returning a sorted list of digits is fine

shared/src/test/scala/ElementTests.scala
```diff
@@ -538,6 +538,23 @@ class ElementTests extends VyxalTests:
     }
   }
 
+  describe("Element S") {
+    testMulti("S")(
+      List[VAny]("qzqadbA;z") -> VList(";Aabdqqzz".map(_.toString)*),
```
Lyxal, line 543: Sorting a string should return a string

Lyxal (changes requested):

shared/src/main/scala/Elements.scala
Steffan153: That's useless though

Lyxal: Well I don't see what else something called "sort" could do to a number that still makes some idiomatic sense.

Steffan153: That's not very useful either.

PR #1681, opened 2023-03-21
Add a group indices element + overlapping pairs element

### LINKED ISSUE #1394: Add element to group indices by their corresponding values
Jelly `Ġ`.

Lyxal:
Replaces "zip self"

Closes #1394


Lyxal:
> This isn't actually very useful. Make this a digraph

[that's why I asked whether it should be a digraph 3 months ago](https://github.com/Vyxal/Vyxal/issues/1394#issuecomment-1368342566). I just assumed that because jelly had it as a monograph, it should be one here too.

Steffan153:
We dont have to copy jelly

Steffan153 (approved):

Issue #14, opened 2021-04-09
Vyxal Corpus


2x-1:
Similar to [this](https://github.com/DennisMitchell/jellylanguage/issues/91) issue opened in the Jelly repo and [this](https://github.com/barbuz/Husk/issues/14) issue opened in the Husk repo, I decided to run Lynn's method on 05AB1E answers (with a few modifications).

[SEDE query](https://data.stackexchange.com/codegolf/query/1395643/vyxal-corpus) used to get all Vyxal answers on CGCC. [Code used to analyze the data.](https://gist.github.com/2x-1/7679095832149ea67b6298337bc68e28) [Final results](https://gist.github.com/2x-1/5a8fb0c71b4421cb7fb79f7427761ff8)

Lyxal:
Results as of 23/06/2021:

```
Digraphs:
  16 +,
  16 \\
  16 ,)
  16 \|
  14 *\
  13 \_
  13 \/
  12 ++
  11 =[
  11 øm
  10 ð*
  10 `\
  10 (n
  10 nε
   9 |:
   8 ?(
   8 :q
   8 3*
   8 d+
   8 $+
   8 n-
   8 ⁰n
   8 p,
   7 ,(
   7 n›
   7 ƛ\
   7 $i
   7 _*
   7 \꘍
   6 `:

Trigraphs:
   7 \_*
   7 \\꘍
   6 `:q
   6 *\|
   6 \|:
   5 ⁰nε
   5 ð*\
   5 \/꘍
   5 +,)
   4 +++
   4 :q$
   4 q$+
   4 \|+
   4 ›\\
   4 `$i
   4 ꘍₴n
   4 |:„
   4 :„+
   4 „+p
   4 +p,
   3 a-a
   3 ∑₀Ḋ
   3 `øṙ
   3 $i;
   3 \\`
   3 ⇧\_
   3 +,(
   3 (\|
   3 ;Ṅ,
   3 *d+

Quadgraphs:
   4 :q$+
   4 ›\\꘍
   4 \|:„
   4 |:„+
   4 :„+p
   4 „+p,
   3 ⇧\_*
   3 *\|+
   3 \\꘍₴
   3 ꘍₴nε
   3 *\|:
   2 `:qp
   2 `:q+
   2 :q+?
   2 q+?O
   2 a-a-
   2 -a-a
   2  I'm
   2 I'm 
   2 12Ḋ⌐
   2 2Ḋ⌐₄
[...]

PR #1690, opened 2023-03-25
Make `@` vectorised lengths

Lyxal:
Based on an observation by @jfioasd in #14


Steffan153 (approved):
Now we have a free string and number overload too

Issue #1486, opened 2022-11-10
Number overload to `g` and `G`


Steffan153:
Either operate on digits (which is arguably useless) or do something else

chunkybanana:
I was considering making them sufficiently_large_number(n) (nth's grahams number, lifetime of worm [n], something like that). We'd never need a sane upper bound again!

Lyxal:
didn't mean to close

PR #1692, opened 2023-03-28
Make `G` and `g` NOPs on Numbers

Lyxal:
Postpones #1486

This is temporary while we wait for something to come up that would be useful


Steffan153:
Sorry? Do you think this is useful?

PR #1701, opened 2023-04-02
Make negative bit shifts not error

nayakrujul:
When shifting a negative amount, it will now shift the other way

For example, [this](https://vyxal.pythonanywhere.com/#WyIiLCIiLCI1TuKGsyIsIiIsIjEwMCJd) should now give 3200 (`100 << 5`) rather than trying `100 >> -5` and failing.


nayakrujul:
Oh yeah, I'll do that now

nayakrujul:
I messed up, let me fix that

nayakrujul:
Should be ok now

PR #1707, opened 2023-04-04
Update `R` so it can take implicit input

### LINKED ISSUE #1706: Bug: Implicit inputs not working with the `R` command
When using `R`, implicit inputs don't work.

See [this program](https://vyxal.pythonanywhere.com/#WyIiLCIiLCJSIiwiIiwiMlxuMTIzIl0=):

Code:
```py
R
```

Inputs:
```py
123
2
```

It should get `123` in base `2`, but instead it's saying *"list index out of range"*.

I think it's because of [this line in `elements.py`](https://github.com/Vyxal/Vyxal/blob/main/vyxal/elements.py#L115). It seems to be trying to pop two items from the stack, but the stack is empty.

nayakrujul:
Closes #1706 (I think)

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -110,14 +110,15 @@ elements: dict[str, tuple[str, int]] = {
     "Q": process_element("exit()", 0),
     "R": (
         """
-rhs, lhs = pop(stack, 2, ctx)
-ts = vy_type(rhs), vy_type(lhs)
-if ts == (NUMBER_TYPE, NUMBER_TYPE):
+lhs = pop(stack, 1, ctx)
+ts = vy_type(lhs)
+if ts == NUMBER_TYPE:
+    rhs = pop(stack, 1, ctx)
     stack.append(string_base_convert(lhs, rhs, ctx))
-elif types.FunctionType in ts:
+elif ts == types.FunctionType:
+    rhs = pop(stack, 1, ctx)
     stack.append(vy_reduce(lhs, rhs, ctx))
 else:
-    stack.append(rhs)
     stack.append(vectorise(reverse, lhs, ctx=ctx))
 """,
         2,
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -110,15 +110,15 @@ elements: dict[str, tuple[str, int]] = {
     "Q": process_element("exit()", 0),
     "R": (
         """
-ts = (vy_type(stack[-1]),) if len(stack) < 2 else (vy_type(stack[-2]), vy_type(stack[-1]))
+rhs, lhs = pop(stack, 2, ctx)
+ts = vy_type(rhs), vy_type(lhs)
 if ts == (NUMBER_TYPE, NUMBER_TYPE):
-    rhs, lhs = pop(stack, 2, ctx)
     stack.append(string_base_convert(lhs, rhs, ctx))
 elif types.FunctionType in ts:
-    rhs, lhs = pop(stack, 2, ctx)
     stack.append(vy_reduce(lhs, rhs, ctx))
 else:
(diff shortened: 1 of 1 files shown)


nayakrujul:
Oh no it didn't work

vyxal/elements.py
```diff
@@ -110,15 +110,15 @@ def process_element(
     "Q": process_element("exit()", 0),
     "R": (
         """
-ts = (vy_type(stack[-1]),) if len(stack) < 2 else (vy_type(stack[-2]), vy_type(stack[-1]))
+rhs, lhs = pop(stack, 2, ctx)
```
Lyxal, line 113: I'm not particularly keen on doing it this way because taking two inputs moves the input cycle by an extra input if only one value is needed.

Lyxal (changes requested):
Either take one and check if it's a number or move the input index back one in the monadic case.

vyxal/elements.py
Lyxal: `ctx.inputs[-1][1] -= 1`. But that's only if input was needed. Obviously don't move it back if no input was taken.

Note that the indexing might be wrong. Check the specs in context.py (I'll make sure to do so myself in my next review)

nayakrujul: But it pushes that input to the stack anyway so it won't matter right?

Lyxal: It pushes it to the stack, but the counter that keeps track of which implicit input is next is also incremented.

nayakrujul:
The problem with that is that if the function is on `rhs` it won't work

Lyxal:
You're definitely on to something here - this is how I would have gone about it too

nayakrujul:
Are `rhs` and `lhs` the wrong way around or not?

Lyxal:
> Are `rhs` and `lhs` the wrong way around or not? Because now it's failing a test it didn't fail before.

The failed test is due to the reversibility of the function overload ((any, fun) should do the same as (fun, any))

nayakrujul:
> The problem with that is that if the function is on `lhs` it won't work

Yeah I don't know how to fix that

Lyxal:
I think I have an idea. Give me a minute or two and I'll push something

Lyxal:
There we go. Good as new!

Lyxal:
I'm just going to give this a quick test in regards to input wrapping. I won't be too long

nayakrujul:
```python
    ctx.inputs[-1][1] -= count_input_popped + 1
```

Why is the `+ 1` there? If there were no inputs popped shouldn't it leave the input cycle alone?

Lyxal:
That's what I'm trying to figure out

nayakrujul:
Would this work?

```python
    if count_input_popped > 1: ctx.inputs[-1][1] -= 1
```

You only need to change it if there were two inputs popped, right?

Lyxal (approved):

nayakrujul:
Wait but even if there were no inputs popped it still decrements `ctx.inputs[-1][1]`?

Lyxal:
dah frick I knew I was missing something.

nayakrujul:
> Would this work?
> 
> ```python
>     if count_input_popped > 1: ctx.inputs[-1][1] -= 1
> ```
> 
> You only need to change it if there were two inputs popped, right?

^^^

Lyxal:
No.

Consider the case of having something on the stack and taking an input.

If `R` only uses the top, the input index still needs decrementing.

nayakrujul:
Oh, so this would work:

```python
    if count_input_popped > 0: ctx.inputs[-1][1] -= 1
```

Lyxal:
#1709

PR #1717, opened 2023-04-14
Version 3: Implement G element (maximum)

Lyxal:

### DIFF
--- a/shared/src/main/scala/Elements.scala
+++ b/shared/src/main/scala/Elements.scala
@@ -533,7 +533,7 @@ object Elements:
       "a: non-lst, b: non-lst -> Maximum of a and b",
       "a: lst, b: fun -> Call b infinitely with items of a as starting values"
     ) { ctx ?=>
-      val top = ctx.pop();
+      val top = ctx.pop()
       top match
         case a: VList =>
           ctx.push(ListHelpers.maximum(a))
--- a/shared/src/main/scala/Elements.scala
+++ b/shared/src/main/scala/Elements.scala
@@ -524,6 +524,32 @@ object Elements:
         ListHelpers.map(a, ListHelpers.makeIterable(b, Some(true)))
     }
 
+    val maximum = addDirect(
+      "G",
+      "Monadic Maximum | Dyadic Maximum | Generate From Function | Vectorised Maximum",
+      List("max", "maximum", "generator"),
+      Some(2),
+      "a: lst -> Maximum of a",
+      "a: non-lst, b: non-lst -> Maximum of a and b",
+      "a: lst, b: fun -> Call b infinitely with items of a as starting values"
+    ) { ctx ?=>
+      val top = ctx.pop();
+      top match
+        case a: VList =>
+          ctx.push(ListHelpers.maximum(a))
+        case _ =>
+          val next = ctx.pop()
+          (top, next) match
+            case (a: VVal, b: VVal) => ctx.push(MiscHelpers.dyadicMaximum(a, b))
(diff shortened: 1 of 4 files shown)


GingerIndustries:
nice number

nayakrujul (approved):

Steffan153 (approved):

shared/src/main/scala/Elements.scala
```diff
@@ -524,6 +524,34 @@ object Elements:
...
+    val maximum = addDirect(
+      "G",
+      "Monadic Maximum | Dyadic Maximum | Generate From Function | Vectorised Maximum",
+      List("max", "maximum", "generator"),
+      Some(2),
+      "a: lst -> Maximum of a",
+      "a: non-lst, b: non-lst -> Maximum of a and b",
+      "a: lst, b: fun -> Call b infinitely with items of a as starting values"
+    ) { ctx ?=>
+      val top = ctx.pop();
```
Steffan153, line 536:
```suggestion
      val top = ctx.pop()
```

shared/src/main/scala/ListHelpers.scala
```diff
@@ -107,6 +130,14 @@ object ListHelpers:
 
   end map
 
+  def maximum(iterable: VList)(using ctx: Context): VAny =
+    iterable match
+      case VList() => VList()
+      case VList(list*) =>
+        list.reduce { (a, b) =>
+          if MiscHelpers.compareExact(a, b) > 0 then a else b
+        }
```
Steffan153, line 139: Can you do something like
```
list.min(Ordering.fromLessThan[(VAny, VAny)](MiscHelpers.compareExact(_, _)))
```

shared/src/main/scala/ListHelpers.scala
```diff
@@ -211,4 +242,11 @@ object ListHelpers:
 
     parts.toSeq
   end split
+
+  def vectorisedMaximum(iterable: VList, b: VVal): VList =
+    VList.from(iterable.map { a =>
+      (a: @unchecked) match
+        case a: VList => vectorisedMaximum(a, b)
+        case a: VVal  => MiscHelpers.dyadicMaximum(a, b)
+    })
```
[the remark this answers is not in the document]
Lyxal: `iterable` might contain further lists, which aren't instances of `VVal`

shared/src/main/scala/Elements.scala
```diff
@@ -524,6 +524,34 @@ object Elements:
...
+      top match
+        case a: VList =>
+          ctx.push(ListHelpers.maximum(a))
+        case _ =>
+          val next = ctx.pop()
+          (top, next) match
+            case (a: VVal, b: VVal) => ctx.push(MiscHelpers.dyadicMaximum(a, b))
+            case (a: VList, b: VFun) =>
```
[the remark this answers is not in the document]
Lyxal: You're right. And in testing I had to put the function as top anyway. I blame copilot for making me think reversibility was needed :p

shared/src/main/scala/ListHelpers.scala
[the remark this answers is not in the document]
Lyxal: I don't know if it would work type-wise

shared/src/main/scala/ListHelpers.scala
Lyxal: Maybe, but I prefer the style of the reduce-by better.

Steffan153: Yes, but are you sure this isn't faster?

PR #1718, opened 2023-04-14
Update .gitattributes (v2)

GingerIndustries:
Seems like the paths for those files changed


GingerIndustries:
I don't know if this is actually a problem for v2, but it definitely is for v3

Lyxal (changes requested):
They haven't moved on the version 2 branch

PR #1719, opened 2023-04-14
Update .gitattributes (v3)

GingerIndustries:
Same thing as other PR

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,3 @@
-documents/elements.md linguist-generated=true
 documents/elements.txt linguist-generated=true
 pages/dictionary.js linguist-generated=true
 static/parsed_yaml.js linguist-generated=true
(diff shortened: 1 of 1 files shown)


Lyxal (changes requested):
`documentation/elements.txt`

There is no md version yet

.gitattributes
```diff
@@ -1,5 +1,4 @@
-documents/knowledge/elements.md linguist-generated=true
-documents/knowledge/elements.txt linguist-generated=true
+documents/elements.txt linguist-generated=true
```
Lyxal, line 1:
```suggestion
documentation/elements.txt linguist-generated=true
```

Lyxal (changes requested):

.gitattributes
Lyxal: https://github.com/Vyxal/Vyxal/blob/version-3/documentation/elements.txt

Lyxal (approved):

PR #1720, opened 2023-04-21
Get a start on implementing fracbyte functionality.

Lyxal:
[We're entering the fracbyte market](https://github.com/AndrovT/Vycoder).

This PR incorporates @AndrovT's brilliant `vycoder` library into the language. Ported to scala as [Vyncode](https://github.com/Vyxal/Vyncode)

How:

- The `!` flag reads a string of raw bits. This might have to be changed for technical scoring reasons, but who knows.
- The `=` flag reads a program as usual, but prints the bit string.

For the website, the result of compiling vyxal/vyncode as scala.js is used

For the offline interpreter, the jar is used.


static/main.js
```diff
@@ -627,7 +639,7 @@ function shareOptions(shareType) {
             output = `[Vyxal, ${len} byte${"s".repeat(code.length != 1)}${utfable ? '' : ' (UTF-8)'}](${url})`
             break
         case "post-template":
-            output = `# [Vyxal](https://github.com/Vyxal/Vyxal)${flagAppendage} ${len} byte${"s".repeat(len != 1)}${utfable ? '' : ' (UTF-8)'}
+            output = `#[Vyxal](https://github.com/Vyxal/Vyxal)${flagAppendage} ${len} byte${"s".repeat(len != 1)}${utfable ? '' : ' (UTF-8)'}
```
Steffan153, line 642:
```suggestion
            output = `# [Vyxal](https://github.com/Vyxal/Vyxal)${flagAppendage} ${len} byte${"s".repeat(len != 1)}${utfable ? '' : ' (UTF-8)'}
```

GingerIndustries:
Don't merge this yet!

nayakrujul:
@GingerIndustries ik, I read the chat

Steffan153 (commented):
Huh? Vycoder already exists in python, why not just use the python?

Steffan153:
For JS you could use something like https://github.com/flexxui/pscript, or just add an API endpoint to Django and use that

Lyxal:
> For JS you could use something like https://github.com/flexxui/pscript, or just add an API endpoint to Django and use that

The library doesn't seem to support imports nor do conversion of all numbers to big ints (numbers involved in the encoding/decoding will exceed js number limits).

And I'd rather not have an api to the server because it'd be making a lot of requests on a server that's already questionably fragile.

Lyxal:
Having it in scala also means there's less work to do when it eventually gets applied to version 3

Steffan153 (approved):
Fine, fine, fine. But I don't like the idea of a bloated site.

Lyxal:
so turns out decoding is a little broken. I'll fix that and see if adding some recent answers to the corpus improves anything.

Lyxal:
Decoding now works good online (it was a bug with not resetting class fields when calling Initalise). I'll play around with the corpus later

PR #1725, opened 2023-04-30
Fix up a few calls to `max` with an empty list

nayakrujul:
Basically there were a few places where `max` was being called with something which could have been an empty list (which would give an error).

Yes, I did test all of them on the online interpreter, and they were all erroring if you gave them empty lists. No, I'm not giving you all the links. No, I don't have time to do the same with `min` right now.

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -528,10 +528,10 @@ def align_left(lhs, ctx):
     (str) -> left-aligned string
     (lst) -> left-align lines
     """
-    
+
     if not lhs:
         return []
-    
+
     ts = vy_type(lhs)
 
     if ts == str:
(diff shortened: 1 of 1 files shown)


Lyxal (approved):
Looks good to me 👍

nayakrujul:
@lyxal I'll do the same with `min` later today

Lyxal:
Cool cool. No pressure or rush to do so

nayakrujul:
@Lyxal I couldn't actually find any for `min` which didn't already have a catch for the empty list, so... well done (I guess?)

Lyxal:
They don't call me Lyxal Golfing for nothing

PR #1737, opened 2023-05-02
Don't defer the other popper.

Lyxal:


Steffan153:
Please no. Our site is slow enough as it is.

PR #1738, opened 2023-05-03
Add prefixes element

Lyxal:

### DIFF
--- a/shared/src/main/scala/ListHelpers.scala
+++ b/shared/src/main/scala/ListHelpers.scala
@@ -216,6 +216,14 @@ object ListHelpers:
 
   end sortBy
 
+  def prefixes(iterable: VList)(using ctx: Context): VList =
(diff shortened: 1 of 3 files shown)


shared/src/main/scala/ListHelpers.scala
```diff
@@ -216,6 +216,14 @@ object ListHelpers:
 
   end sortBy
 
+  def prefixes(iterable: VList)(using ctx: Context): VList =
+    return VList(
+      iterable.indices
+        .map { i =>
+          VList(iterable.take(i + 1)*)
+        }*
+    )
```
Steffan153, line 225: `VList.from(iterable.inits.reverse.tail)`

Steffan153: Actually you probably need to map `VList.from` too

PR #1741, opened 2023-05-05
Add: L for Length, Q for Quit and T for Triple | Update: Tests for P | Change Grouping Lambda Modifier Characters

Lyxal:


nayakrujul (approved):

nayakrujul (approved):

shared/src/main/scala/ListHelpers.scala
```diff
@@ -245,6 +245,40 @@ object ListHelpers:
...
+            Some((Some(matrix(c).index(r)), c + 1))
+          else if filler.isDefined then
+            Some((Some(filler.getOrElse(VNum(0))), c + 1))
+          else Some((None, c + 1))
+        }
+      VList.from(temp.flatten)
+
+    val out: LazyList[VAny] = LazyList
+      .unfold(0) { r =>
+        if !matrix.exists(_.isDefinedAt(r)) then None
```
Steffan153, line 272: This doesn't work on an infinite matrix. Also it's bad preformance to check that on every row

Steffan153: Check out 05ab1e's approach
https://github.com/Adriandmen/05AB1E/blob/master/lib/commands/matrix_commands.ex#L30-L39
It's pretty good

Steffan153: `matrix.forall(_.isEmpty)` This probably hangs for infinite lists, right?

05ab1e's works fine on infinites, as it just filters out empty ones and checks if a list is empty. If it's infinite, it is never empty, so it keeps on infinitely generating.

Lyxal: > This doesn't work on an infinite matrix. Also it's bad preformance to check that on every row

It should, because it's a port of the version 2 algorithm which allegedly works on infinite lists

Steffan153: > allegedly

`exists` will hang on an infinite list if none are empty

Steffan153: Or maybe it does work, I didn't actually test the scala version. The v2 works https://vyxal.pythonanywhere.com/?v=1#WyIiLCIiLCLDnuKInsabw57iiJ47XG5cbuKIqVxuXG4xMOG6jsabMTDhuo47IiwiIiwiIl0=

Steffan153: Yep it hangs https://tio.run/##yy9OTMpM/f/f9PBcTYXD8w7v@P8fAA

Steffan153: If you think about it, it's not possible, because the trilionth element of the infinite list might have another element. I'd say add a digraph for infinite lists that goes off the length of the first list in the matrix.

Lyxal: I'm fine with the digraph option

PR #1742, opened 2023-05-06
Use a Newly Generated FontAwesome Kit

Lyxal:
I realised today that I've completely lost access to the account that has the old kit. So I made a new one just in case anything ever goes wrong with it.


nayakrujul (approved):

Steffan153:
We should be using inline SVGs instead of loading bloated JS anyway but whatever.

PR #1761, opened 2023-05-25
Add non-modular indexing

Lyxal:
I swear there's been cases where it's been a problem to not have non-modular indexing

### DIFF
--- a/documents/knowledge/elements.md
+++ b/documents/knowledge/elements.md
@@ -3820,6 +3820,14 @@ Similar to set union, but with duplicates allowed.
 
 - lst a, lst b: `multiset union of a and b`
 -------------------------------
+## `` Þḭ `` (Non-Modular Index)
+
+Get the nth item of a list, but without modular arithmetic. Negative indices are allowed.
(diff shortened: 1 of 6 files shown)


Steffan153 (changes requested):
Erroring is extremely unuseful in code golf. You should return something else like `0`.

Lyxal:
> Erroring is extremely unuseful in code golf. You should return something else like `0`.

That would be a different overload then. I want something that _does_ error. Even if it's not useful.

Lyxal:
I also wanted something that explicitly disrupts control flow/easily tells you if the index was out of bounds. Returning 0 if out of bounds doesn't easily tell you if the index was out of bounds because you may have gotten a 0 out of the list.

Lyxal:
if ind > length: return "invalid input" or something like that

Lyxal:
It may not be the most useful or golfy thing, but it's something I want to be a thing in version 2

Steffan153 (approved):
Fine

PR #1789, opened 2023-06-19
Add command for basic parsing of HTML using BeautifulSoup

PlaceReporter99:


PlaceReporter99:
Well, I need help resolving dependency issues…

Steffan153 (dismissed):
Not the eval. That will be another security vulnerability.

vyxal/elements.py
```diff
@@ -443,6 +444,10 @@ def process_element(
     "k□": process_element("[[0,1],[1,0],[0,-1],[-1,0]]", 0),
     "kṘ": process_element('"IVXLCDM"', 0),
     "k•": process_element('["qwertyuiop","asdfghjkl","zxcvbnm"]', 0),
+    "¨P": process_element(
+        'in = pop(stack, 2, ctx) ; stack.append(eval(f"bs({in[1]},"html.parser").{in[0]}"))',
```
Lyxal, line 448: This currently returns a `bs4.BeautifulSoup` object, which will can't be used by anything - the only types used in Vyxal are numbers (a combination of `int and any `Sympy` types), strings, lists/`LazyList`s and functions. The HTML parser would need to return something usable by programs.

Lyxal (changes requested):

vyxal/elements.py
PlaceReporter99: It will be converted to string before returning.

vyxal/elements.py
```diff
@@ -445,7 +445,7 @@ def process_element(
     "kṘ": process_element('"IVXLCDM"', 0),
     "k•": process_element('["qwertyuiop","asdfghjkl","zxcvbnm"]', 0),
     "¨P": process_element(
-        'in = pop(stack, 2, ctx) ; stack.append(eval(f"bs({in[1]},"html.parser").{in[0]}"))',
+        'in = pop(stack, 2, ctx) ; stack.append(str(eval(f"bs({in[1]},"html.parser").{in[0]}")))',
```
Lyxal, line 448: Just converting to string returns the original HTML with slightly less newlines/somewhat neater formatting. It doesn't return any parsed information at all. You would need to return a list of all items in the HTML document.

Lyxal (changes requested):

vyxal/elements.py
```diff
@@ -443,6 +444,10 @@ def process_element(
     "k□": process_element("[[0,1],[1,0],[0,-1],[-1,0]]", 0),
     "kṘ": process_element('"IVXLCDM"', 0),
     "k•": process_element('["qwertyuiop","asdfghjkl","zxcvbnm"]', 0),
+    "¨P": process_element(
```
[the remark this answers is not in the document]
PlaceReporter99: Fixed.

vyxal/elements.py
PlaceReporter99: Now returns list with actual parse data.

PlaceReporter99:
> Not the eval. That will be another security vulnerability.

fixed

PlaceReporter99:
Is this ready to merge?

Lyxal:
> Is this ready to merge?

@PlaceReporter99 I was wondering if you found the changes I made to the PR were what you had in mind.

PlaceReporter99:
Well I was thinking where you specify the HTML document and the parse syntax and it returns that section of the html document, like this:
<code><code>\`HTML Doc\`\`body.div.form.p\`¨P</code> (Vyxal) --> <code>stack.append(flattenlist(list(BeautifulSoup("HTML Doc","html.parser").find_all("body").find_all("div").find_all("form").find_all("p"))))</code> (Pseudo-Python)</code>
Basically extracts all paragraphs that are in a form that are in a div that are in a body.

Lyxal (approved):

PR #1799, opened 2023-07-01
Add function to pause execution for a specified time

PlaceReporter99:
now use `(num) ¨w` to wait num seconds!


nayakrujul (commented):
Looks good, but do you want to add a check to see if it's a number so that `time.sleep` doesn't error?

PlaceReporter99:
at some point

Lyxal (changes requested):
I think adding a check is a good idea

PlaceReporter99:
> I think adding a check is a good idea

@Lyxal @nayakrujul  fixed

vyxal/elements.py
```diff
@@ -443,6 +444,7 @@ def process_element(
     "k□": process_element("[[0,1],[1,0],[0,-1],[-1,0]]", 0),
     "kṘ": process_element('"IVXLCDM"', 0),
     "k•": process_element('["qwertyuiop","asdfghjkl","zxcvbnm"]', 0),
+    "¨w": ("time.sleep((_:=pop(stack, 1, ctx)) if type(_)==type(1) else 0)", 1),
```
Lyxal, line 447: This shouldn't push 0. How about `lhs = pop(stack, 1, ctx)\nif vy_type(lhs) == NUMBER_TYPE: time.sleep(int(lhs))`

Lyxal (changes requested):

vyxal/elements.py
PlaceReporter99: It waits 0

Lyxal (approved):

nayakrujul (approved):

PR #1803, opened 2023-07-03
Use `vy_type` for checking against `NUMBER_TYPE`

Lyxal:
`NUMBER_TYPE` is actually a string. `vy_type` returns `NUMBER_TYPE` for all floats, ints, and sympy types. `isinstance` doesn't know about `NUMBER_TYPE` at all.


ysthakur (approved):

nayakrujul:
Sorry, that's my bad

PR #1824, opened 2023-07-15
reorder base_27_alphabet letters to match frequency.

PlaceReporter99:
[See here for letter frequency source.](https://en.wikipedia.org/wiki/Letter_frequency#:~:text=According%20to%20Lewand%2C%20arranged%20from%20most%20to%20least%20common%20in%20appearance%2C%20the%20letters%20are%3A%20etaoinshrdlcumwfgypbvkjxqz.)

### DIFF
--- a/vyxal/encoding.py
+++ b/vyxal/encoding.py
@@ -47,4 +47,4 @@ for char in string.printable:
 codepage_number_compress = codepage.replace("»", "")
 codepage_string_compress = codepage.replace("«", "")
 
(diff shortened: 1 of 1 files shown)


PlaceReporter99:
@ysthakur it has the more frequent letters sooner, so the number will be less and ultimately will take less data to store. If you want, we can PR this into v3 instead.

PlaceReporter99:
just wanted to fork it

PR #1825, opened 2023-07-15
Allow auto-closing strings

ysthakur:
When executing a program with a string with no closing quote, the interpreter seems to hang. Bug discovered by @nayakrujul. Quick patch since the big PR won't be merged for a while


nayakrujul (approved):
LGTM 👍

PR #1832, opened 2023-07-18
Add ¨" element and allow (str,list) as args for β

PlaceReporter99:
Much easier to implement digraphs in the string compression utility!

### DIFF
--- a/documents/knowledge/elements.md
+++ b/documents/knowledge/elements.md
@@ -1105,6 +1105,8 @@ Convert a number from a custom base to base 10
 
 - any a, num b: `a to base 10 from number base b, treating list items / string items as digits`
 - str a, str b: `a to base 10 from custom string base b, replacing values in a with their index in b and converting to base 10`
+- lst a, lst b: `a to base 10 from custom list base b, replacing values in a with their index in b and converting to base 10`
+- str a, lst b: `a to base 10 from custom list base b, parsing a into a list only containing the items in b, then replacing values in a with their index in b and converting to base 10`
 -------------------------------
 ## `` τ `` (From Base Ten / To Custom Base)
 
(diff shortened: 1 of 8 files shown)


PlaceReporter99:
Do I add the test to the element test file or create a separate file @Lyxal

Lyxal:
To the yaml file - make sure to add a description of what the new element does.

Lyxal:
elements.yaml specifically. Don't add it to the test file manually because that file is automatically generated

PlaceReporter99:
@Lyxal finished

Lyxal:
I think I might extend this PR to also make the new digraph parse json structures. That's a feature that's been suggested before and could actually be pretty cool.

PlaceReporter99:
@Lyxal i think something went wrong with one of your checks

PlaceReporter99:
@Lyxal i think `øJ` already parses json.

Lyxal:
> @Lyxal i think `øJ` already parses json.

Ah, so it does.

Lyxal (approved):

Lyxal (changes requested):
The test case `["abcaabac",["a","bc","ac","b"]] : ["a","bc","a","a","b","ac"]` doesn't make sense - the first `bc` will always be treated as `b` instead of `bc` because `re.match` gets the shortest match.

That's why the test was not finishing. It was in an infinite loop.

PlaceReporter99:
@Lyxal running it in the ide, the problem is the last `c` isn't matched.

PlaceReporter99:
Well it’s returning the correct value for that now but it’s claiming invalid types were given…

Lyxal:
I needed to rename the branch for git purposes. This PR is now #1833

PR #1833, opened 2023-07-20
Add element to parse string using alphabet

Lyxal:
I had to resort to having tuples containing the index of the item.


PlaceReporter99:
Dammit! I forgot the ctx!

PlaceReporter99 (approved):
I would obviously say yes since I came up with this idea.

PlaceReporter99:
wonder why branches get auto-deleted...

PR #1834, opened 2023-07-22
Add SHA256 hash function

PlaceReporter99:
I'm surprised Vyxal does not have this yet!

### DIFF
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -5862,7 +5862,7 @@ def separate_runl_encode(lhs, ctx: Context):
 @element("ø%", 1)
 def sha256_hash(lhs, ctx):
     """Element ø%
-    (any) -> hash a using SHA256
+    (any) -> hash a value using SHA256 (after converting to string)
     """
     return sha256(str(lhs).encode()).hexdigest()
 
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -5858,6 +5858,7 @@ def separate_runl_encode(lhs, ctx: Context):
     ctx.stacks[-1].append(items)
     return lengths
 
+
 @element("ø%", 1)
 def sha256_hash(lhs, ctx):
     """Element ø%
--- a/vyxal/elements.py
+++ b/vyxal/elements.py
@@ -450,7 +450,6 @@ else:
         "lhs = pop(stack, 1, ctx); vy_type(lhs) == NUMBER_TYPE and time.sleep(lhs)",
         1,
     ),
-    "ø%": process_element("sha256(lhs.encode()).hexdigest()", 1),
 }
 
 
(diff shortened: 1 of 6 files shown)


nayakrujul (approved):
Cool. Looks good.

vyxal/elements.py
```diff
@@ -449,6 +450,7 @@ def process_element(
         "lhs = pop(stack, 1, ctx); vy_type(lhs) == NUMBER_TYPE and time.sleep(lhs)",
         1,
     ),
+    "ø%": process_element("sha256(lhs.encode()).hexdigest()", 1),
```
ysthakur, line 453: I'd prefer if this were implemented as a function with the `@element` decorator, something like
```python
@element("ø%", 1)
def hash(lhs, ctx):
    """Element ø%
    ...
    """
    if isinstance(lhs, str):
        return sha256(lhs.encode()).hexdigest()
    else:
        <raise some kind of error saying lhs must be a string>
```

ysthakur:
This is a good PR, it's nice that you added a test. What's the use case for it, though? Do you have any answers where SHA-256 would've been useful?

vyxal/elements.py
```diff
@@ -5858,6 +5859,14 @@ def separate_runl_encode(lhs, ctx: Context):
     return lengths
 
 
+@element("ø%", 1)
+def sha256_hash(lhs, ctx):
+    """Element ø%
+    ...
```
ysthakur, line 5865: Make sure to put an actual element description here

vyxal/elements.py
```diff
@@ -5858,6 +5859,14 @@ def separate_runl_encode(lhs, ctx: Context):
...
+    """
+    return sha256(str(lhs).encode()).hexdigest()
```
ysthakur, line 5867: A check for whether `lhs` is a string or not would be helpful, so users don't get weird errors saying `encode` can't be called on a `list` or whatever

ysthakur (changes requested):

vyxal/elements.py
nayakrujul: Yeah, just realised

vyxal/elements.py
nayakrujul: It converts to `str` anyway:

> `str(lhs).encode()`

vyxal/elements.py
```diff
@@ -5858,6 +5859,14 @@ def separate_runl_encode(lhs, ctx: Context):
     return lengths
 
 
+@element("ø%", 1)
+def sha256_hash(lhs, ctx):
+    """Element ø%
+    (any) -> hash a using SHA256
```
ysthakur, line 5865:
```suggestion
    (str) -> hash a string using SHA256
```
This only works on strings, not any value (I guess you could do `str(lhs).encode()`, but I don't see much point in bothering)

vyxal/elements.py
ysthakur: Ah, my bad, didn't notice you did `str(lhs)`

vyxal/elements.py
nayakrujul: It *does* convert to string. `str(lhs).encode()` is in the code already.

ysthakur, line 5865:
```suggestion
    (any) -> hash a value using SHA256 (after converting to string)
```

ysthakur (approved):
I'm not sure when this would be helpful but doesn't hurt to have it. Good work PlaceReporter99 and nayakrujul!

Lyxal (approved):
Very good

PR #1842, opened 2023-07-30
Make vyncode not override the debug box

### LINKED ISSUE #1841: Vyncode bitstring overwrites time out message (and probably other messages too)
When using the = flag, time out messages are overwritten. Perhaps the bitstring should be appeneded to the debug box

Lyxal:
Closes #1841


PlaceReporter99 (approved):
Looks good

PR #1846, opened 2023-07-31
Improve documentation for input

Lyxal:

### DIFF
--- a/documents/knowledge/elements.yaml
+++ b/documents/knowledge/elements.yaml
@@ -659,7 +659,7 @@
 - element: "?"
   name: Input
   arity: 0
-  description: Get the next input from the input source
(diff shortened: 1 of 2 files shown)


Peter-developer01 (approved):
LGTM!

documents/knowledge/elements.yaml
```diff
@@ -659,7 +659,7 @@
 - element: "?"
   name: Input
   arity: 0
-  description: Get the next input from the input source
+  description: Get and evaluate the next input (readline) from the input source.
```
PlaceReporter99, line 662: I think you should add that the global canvas and array can be affected by this.

PlaceReporter99 (changes requested):

nayakrujul (approved):

PR #1872, opened 2023-08-18
Remove some instances of VVal

ysthakur:
Just removing some instances of VVal so it looks cleaner. also removed vdata and vatom.


nayakrujul (approved):
I'm probably not the best person to ask (I only started learning Scala today), but LGTM :+1:

PR #1894, opened 2023-08-23
Add some stuff to the debugger

ysthakur:
Adding some stuff to the debugger:

- [X] Resume command
- [X] Breakpoints
- [X] Ability to evaluate code while you're in the debugger


Lyxal (approved):
The fact there are debug tests is horrifying.

We making it out the lack of tooling that exists with most golflangs with this one

PR #1897, opened 2023-08-23
refactor: Merge addElem and addVect

ysthakur:
@nayakrujul suggested that `addElem` take a `vectorises` argument like `addFull` for convenience. This PR merges `addElem` and `addVect` into a single `addPart` helper that takes a `vectorises` argument. This'll make `Elements.scala` longer, but it's one less helper to remember. There's like a billion changes in this PR, so enjoy reviewing!


Lyxal (approved):
Lgtm.


(I reviewed the whole thing. No clue what you two were complaining about by calling it hard to review. That just sounds like a skill issue on y'all's part :p)

PR #1906, opened 2023-08-26
Format all files in workflow

ysthakur:
Previously, it was only formatting the shared and JVM sources (not including the tests). Now it should format all Scala files in the project.


nayakrujul (approved):
LGTM :+1:

PR #1907, opened 2023-08-26
Diagonalize sublists

ysthakur:
Currently, the sublists element won't enumerate every sublist even given unlimited time, so this PR uses `mergeInfLists` to make sure every sublist is enumerated eventually (like cartesian product). Unlike cartesian product, I don't think people usually rely on the exact order of the sublists, so I think this change is probably fine?


nayakrujul (approved):
:+1:

PR #1910, opened 2023-08-27
Add [0..a) and [1..a]

nayakrujul:


shared/src/vyxal/Elements.scala
```diff
@@ -1515,6 +1513,38 @@ object Elements:
       case a: VNum => a ** 3
       case a: String => VList.from(a.grouped(3).toSeq)
     },
+    addPart(
+      Monad,
+      "ɾ",
+      "Inclusive One Range | Uppercase",
+      List("1..n", "one->n", "one-range", "to-upper", "upper", "uppercase"),
```
Lyxal, line 1520: I didn't think `.` was valid in literate mode keywords.

Lyxal (approved):
Cool beans homie. Cool beans.

shared/src/vyxal/Elements.scala
nayakrujul: Umm... well it didn't give me any errors :p

Lyxal: Well of course not

There's not tests or checks for that

nayakrujul: Fine. I'll change it

PR #1918, opened 2023-09-02
Add more words to dictionaries

Lyxal:
I think Jelly sometimes has shorter strings when the dictionaries are smaller.


Lyxal:
> Why is sss_ompress.py in pages and not dict_scripts?

Because I needed it there for testing the dictionaries

PR #1937, opened 2023-10-04
Version 3 Better Recursion mod

Lyxal:
This implements

- Global recursion (already done)
- Recursion in a non-grouping modifier calls the current call-stack function (bug ridden)

What I'm using for testing recursion in modifiers is:

```
[[1,2,3], [4,5,6]] {str "[" contains? ? v: recurse sum end} call
```

Not very useful (literally just flatten then sum), but it should give `21`. It does not, at the time of writing.


Lyxal:
I uh may have broken the tests. I'll fix that tomorrow.

PR #1944, opened 2023-10-13
Add assignment and insertion

Lyxal:


Lyxal:
TODO: Make it so that it fills with 0s if index not in list

PR #1953, opened 2023-10-25
Add a lot of things to version 3

### LINKED ISSUE #1885: [v3] Make help flag output to the output box in the website
Apparently it currently outputs to the console. Would be helpful to output to the output box instead, like v2. Or maybe there could be a popup. We could expose the help text in the JS `Vyxal` object so that the fancy Vue website can make a tab for it or something.

Lyxal:
closes #1885


GingerIndustries:
Finally got the bot to autolabel this

Lyxal:
Todo: Add tests

shared/src/vyxal/Context.scala
```diff
@@ -46,18 +46,23 @@ class Context private (
     * inputs, read a line of input from stdin.
     */
   def pop(): VAny =
-    if useStack then return getTopCxt().pop()
+    if useStack then
+      val temp = parent.getOrElse(getTopCxt()).pop()
```
GingerIndustries, line 50: variable assignment why

GingerIndustries (approved):

ysthakur (changes requested):
Commented in the wrong place but there's no need to make the CLI parser public: https://github.com/Vyxal/Vyxal/commit/b800777f23fc65769342709d9a8c997ee056d2cc#r131282031

shared/src/vyxal/Modifiers.scala
```diff
@@ -414,5 +502,108 @@ object Modifiers:
...
+            astToLambda(ast1, ast1.arity.getOrElse(-1)),
+            astToLambda(ast2, ast2.arity.getOrElse(-1)),
+            AST.Command("#|para-apply"),
+          )
+      },
+    "∦" ->
+      Modifier(
+        "Parallel Apply and Wrap",
+        """|Parallel apply two elements to the top of the stack
+             |and wrap the result in a list
```
ysthakur, line 560: This isn't important at all, but could you align the `|`s here and below?

shared/src/vyxal/Context.scala
```diff
@@ -176,6 +181,33 @@ class Context private (
     parent match
       case Some(p) => p.getTopCxt()
       case None => this
+
+  def rotateLeft: Unit =
+    if isStackEmpty then push(pop())
+    else
+      val temp = stack.remove(0)
+      stack += temp
```
ysthakur, line 189:
```suggestion
      stack += stack.remove(0)
```

shared/src/vyxal/Context.scala
```diff
@@ -176,6 +181,33 @@ class Context private (
...
+
+  def rotateRight: Unit =
+    val temp = pop()
+    stack.insert(0, temp)
```
ysthakur, line 193:
```suggestion
    stack.insert(0, pop())
```

shared/src/vyxal/Context.scala
```diff
@@ -176,6 +181,33 @@ class Context private (
...
+      mut.ArrayBuffer().addAll(stack),
+      _ctxVarPrimary,
+      _ctxVarSecondary,
+      ctxArgs,
+      vars.clone(),
+      inputs,
+      (parent match
+        case Some(p) => Some(p.copy)
+        case None => None
+      ),
```
ysthakur, line 206:
```suggestion
      parent.map(_.copy),
```

shared/src/vyxal/Elements.scala
```diff
@@ -2263,6 +2313,159 @@ object Elements:
...
+      None,
+      "*a, f -> scanl each column. Use the modifier instead.",
+    ) { ctx ?=>
+      val f = ctx.pop()
+      val arg = ListHelpers
+        .makeIterable(ctx.pop())
+        .map(x => ListHelpers.makeIterable(x))
+      f match
+        case fun: VFun => ctx.push(
+            VList.from(
```
ysthakur, line 2380: Indexing can be slow on big linked lists. I think we could try something like

```scala
LazyList.unfold(arg) { rows =>
  val rem = rows.filter(_.nonEmpty)
  Option.when(rem.nonEmpty) {
    val col = rem.map(_.head)
    val nextRows = rem.map(_.tail)
    (MiscHelpers.scanl(col, fun), nextRows)
  }
}
```

This is completely untested, though, and I have a splitting headache right now so don't blame me if it doesn't work

shared/src/vyxal/Elements.scala
```diff
@@ -2386,10 +2589,130 @@ object Elements:
...
+    },
+    addDirect(
+      "\\",
+      "Dump",
+      List("dump"),
+      Some(1),
+      "a: any -> dump all values on the stack",
+    ) { ctx ?=>
+      val a = ListHelpers.makeIterable(ctx.pop())
+      for v <- a do ctx.push(v)
```
ysthakur, line 2637:
```suggestion
      ListHelpers.makeIterable(ctx.pop()).foreach(ctx.push)
```
Why type lot token when few token do trick

shared/src/vyxal/Elements.scala
```diff
@@ -2386,10 +2589,130 @@ object Elements:
...
+      Monad,
+      "Π",
+      "Product",
+      List("product", "prod"),
+      false,
+      "a: lst -> product of a",
+    ) {
+      case a: VList => ListHelpers.product(a)
+      case a: String => ListHelpers.product(ListHelpers.makeIterable(a))
+      case a: VNum => ListHelpers.product(ListHelpers.makeIterable(a))
```
ysthakur, line 2662: Is this just factorial?

shared/src/vyxal/MiscHelpers.scala
```diff
@@ -199,4 +203,18 @@ object MiscHelpers:
       case (a: VList, b: VList) => VList.from((a, b).zipped.map(subtract(_, _)))
       case (a, b) => ???
 
+  def untilNoChange(function: VFun, value: VAny)(using Context): VAny =
```
ysthakur, line 206:
```suggestion
  def untilNoChange(function: VFun, value: VAny)(using Context): VList =
```
Just making it more specific (unless making it a `VList` messes things up elsewhere, in which case, ignore this)

Lyxal:
> Commented in the wrong place but there's no need to make the CLI parser public: https://github.com/Vyxal/Vyxal/commit/b800777f23fc65769342709d9a8c997ee056d2cc#r131282031

How else is the js gonna get the help usage?

shared/src/vyxal/Elements.scala
Lyxal: It's neither.

`ListHelpers.makeIterable(a)` returns a list of digits by default, not a range.

ysthakur:
@Lyxal see the comment I linked

shared/src/vyxal/Elements.scala
```diff
@@ -193,6 +194,12 @@ object Elements:
           scribe.warn(s"Could not invert matrix $l")
           l
         }
+      case l: VList => l.vmap(item =>
```
ysthakur, line 197: You can use {case directly instead of item =>item match

shared/src/vyxal/Elements.scala
```diff
@@ -2386,10 +2586,129 @@ object Elements:
...
+    },
+    addPart(
+      Monad,
+      "⌊",
+      "Floor",
+      List("floor"),
+      true,
+      "a: num -> floor(a)",
+    ) {
+      case a: VNum => VNum(spire.math.floor(a.underlying.real))
```
ysthakur, line 2708: Does a.floor not work?

ysthakur:
This is turning into a really big PR so I think it's best if it's merged soon before more stuff is added to it

shared/src/vyxal/MiscHelpers.scala
```diff
@@ -199,4 +203,18 @@ object MiscHelpers:
...
+    val res = LazyList.unfold(value) { curr =>
+      val next = function(curr)
+      if next == prev then None
+      else
+        prev = next
+        Some(next -> next)
+    }
+    VList.from(value #:: res)
+
+  def zipWith(left: VList, right: VList, function: VFun)(using Context): VList =
```
ysthakur, line 217: Since there's already a method for zipping `VList`s that works on Scala functions, I think we can add methods to `VFun` for converting to Scala functions (`.asFn1`, `.asFn2`, ...). That'd avoid duplicating code for Scala functions and `VFun`s. Thoughts?

ysthakur, line 217: Since there's already a method for zipping `VList`s that works on Scala functions, I think we can add methods to `VFun` for converting to Scala functions (`.asFn1`, `.asFn2`, ...). That'd avoid duplicating code for Scala functions and `VFun`s. Thoughts? @Lyxal

shared/src/vyxal/Elements.scala
Lyxal: It was originally for consistency because there was no `VNum.ceil`

Lyxal:
> This is turning into a really big PR so I think it's best if it's merged soon before more stuff is added to it

I don't plan on adding any more things. It's ready for merge pending reviews.

shared/src/vyxal/MiscHelpers.scala
Lyxal: That might be a bit tricky as `VFun` would need all possible `asFn<arity>` functions. Also, how would `asFnNeg1` work?

ysthakur: We wouldn't make an `asFnNeg1`. It's only so that `VFun`s can be used with methods that want a `VAny => Context ?=> VAny` or `(VAny, VAny) => Context ?=> VAny`. `asFn1` and `asFn2` should be adequate, maybe `asFn3` at most.

shared/src/vyxal/Elements.scala
```diff
@@ -193,6 +194,11 @@ object Elements:
           scribe.warn(s"Could not invert matrix $l")
           l
         }
+      case l: VList => l.vmap({
+          case n: VNum => VNum(NumberHelpers.toBinary(n).size)
+          case default => default
+        })
```
ysthakur, line 200:
```suggestion
      case l: VList => l.vmap {
          case n: VNum => VNum(NumberHelpers.toBinary(n).size)
          case default => default
        }
```
You can actually omit the parentheses. If you only have one argument to pass, Scala lets you use `{}` instead of `({})`.

shared/src/vyxal/MiscHelpers.scala
```diff
@@ -186,6 +187,9 @@ object MiscHelpers:
     vyPrint(x)
     vyPrint("\n")
 
+  def scanl(iterable: VList, function: VFun)(using ctx: Context): VList =
+    VList.from(ListHelpers.prefixes(iterable).map(reduce(_, function)))
```
ysthakur, line 191:
```suggestion
    if iterable.isEmpty then iterable
    else
      iterable.tail.scanLeft(iterable.head)(function(_, _))
```
Doing `prefixes` and reducing each prefix separately will be inefficient. If it turns out `scanLeft` on `VList`s requires evaluating them, which I don't think it does, but if that is the case, we can always modify `scanLeft` to be lazier

shared/src/vyxal/MiscHelpers.scala
```diff
@@ -199,4 +203,18 @@ object MiscHelpers:
...
+      val next = function(curr)
+      if next == prev then None
+      else
+        prev = next
+        Some(next -> next)
+    }
+    VList.from(value #:: res)
+
+  def zipWith(left: VList, right: VList, function: VFun)(using Context): VList =
+    VList.from(left.zip(right).map((a, b) => function(a, b)))
```
ysthakur, line 218: This should probably work:
```suggestion
    left.zipWith(right) { (a, b) => function(a, b) }
```
If we do end up adding `.asFn2` later, we could write `left.zipWith(right)(function.asFn2)`

ysthakur:
> > This is turning into a really big PR so I think it's best if it's merged soon before more stuff is added to it
> 
> I don't plan on adding any more things. It's ready for merge pending reviews.

Awesome, this PR looks pretty good to me, minus the minor issues I mentioned above. After resolving those, feel free to dismiss my previous rejection and merge (I think Ginger's approval should allow you to do that)

shared/src/vyxal/MiscHelpers.scala
```diff
@@ -186,6 +187,11 @@ object MiscHelpers:
     vyPrint(x)
     vyPrint("\n")
 
+  def scanl(iterable: VList, function: VFun)(using ctx: Context): VList =
+    if iterable.isEmpty then iterable
+    else
+      iterable.tail.scanLeft(iterable.head)(function(_, _))
```
ysthakur, line 193:
```suggestion
      VList.from(iterable.tail.scanLeft(iterable.head)(function(_, _)))
```

ysthakur (approved):

PR #1963, opened 2023-10-31
So uh apparently `<backtick>^_` were in the wrong order

Lyxal:


Lyxal:
In order to not break the thousands of permalinks, the new codepage is enabled with a `c=1` URL parameter. This makes it opt-in. All new permalinks will be generated with `c=1`, but old links will not have the `c=1` copied in.

TODO: Make that more obvious in a different PR once this fix is pushed through. I want to have this go through before styling things so that it's fixed ASAP.

GingerIndustries (approved):

PR #1964, opened 2023-10-31
Make it so that `Ǔ` and `ǔ` are considered monadic to modifiers.

Lyxal:


Lyxal:
For example, you might want to vectorise rotation by one unit. I believe that would be more frequently wanted than vectorised rotate by n units.

ysthakur (approved):
LGTM

ysthakur:
A test or two would be nice

Steffan153 (approved):

PR #1971, opened 2023-11-07
Make Scala.js emit module instead of script

ysthakur:
To be able to use the Scopt help text, we need module support. Unfortunately, modules are a huge pain to get working. This currently doesn't work. It looks like the line in the web worker where it imports `vyxal.js` makes it so that the web worker doesn't respond? If I replace `"./vyxal.js"` with `"./os.module.js"`, then it does at least print an error message.

### DIFF
--- a/js/src/vyxal/JSVyxal.scala
+++ b/js/src/vyxal/JSVyxal.scala
@@ -20,30 +20,36 @@ object JSVyxal:
       printFunc: js.Function1[String, Unit],
   ): Unit =
     // todo take functions to print to custom stdout and stderr
-    if code.contains('h') then
-      val resultPromise: js.Promise[String] = js.dynamicImport {
-        HelpText().getHelpText
-      }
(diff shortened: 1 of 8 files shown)


js/src/vyxal/JSVyxal.scala
```diff
@@ -17,28 +20,37 @@ object JSVyxal:
...
+      )
+      Interpreter.execute(code)(using ctx)
   end execute
 
+  @JSExport
+  def printHelpText(printFunc: js.Function1[String, Unit]): Unit =
+    val resultPromise: js.Promise[String] = js.dynamicImport {
+      HelpText().getHelpText
+    }
+    for result <- resultPromise.toFuture do printFunc(result)
```
ysthakur, line 52: If you want to make it shorter:
```suggestion
    js.dynamicImport { HelpText().getHelpText }.then { text => printFunc(text) }
```

ysthakur (commented):
Merge at will

PR #1973, opened 2023-11-11
Add a markdown table generator for elements as well as 🌮

### LINKED ISSUE #1975: Add 🌮 command to version 3
Like cookie, but taco

Lyxal:
Closes #1975


jvm/src/vyxal/gen/GenerateDocs.scala
```diff
@@ -57,4 +59,46 @@ private object GenerateDocs:
     SugarMap.trigraphs
       .map { case (key, value) => s"$key -> $value" }
       .mkString("", "\n", "\n")
+
+  def elementTable(): String =
+    val header = "| Symbol | Name | Keywords | Arity | Vectorises | Overloads |"
+    val divider = "| --- | --- | --- | --- | --- | --- |"
+    val contents = StringBuilder()
+    val formatOverload = (overload: String) =>
```
ysthakur, line 67:
```suggestion
    def formatOverload(overload: String) =
```
Makes the intent more clear (I thought at first that you needed to pass `formatOverload` elsewhere). If you want to use it like a helper, it's better to make it a normal nested method rather than making a function object

jvm/src/vyxal/gen/GenerateDocs.scala
```diff
@@ -57,4 +59,46 @@ private object GenerateDocs:
...
+      val (args, description) = overload.splitAt(overload.indexOf("->"))
+      val newDesc = description
+        .replace("|", "\\|")
+        .replace("->", "")
+        .stripLeading()
+        .stripTrailing()
+      if args.stripTrailing() == "" then s"`$newDesc`"
+      else
+        s"`${args.stripTrailing().replace("|", "\\|").replace("->", "")}` => `$newDesc`"
+    val addRow = (elem: Element) =>
```
ysthakur, line 77:
```suggestion
    def addRow(elem: Element) =
```

jvm/src/vyxal/gen/GenerateDocs.scala
ysthakur, line 67:
```suggestion
    def formatOverload(overload: String) =
```

jvm/src/vyxal/gen/GenerateDocs.scala
ysthakur, line 77: Why not put this directly in the foreach?
```suggestion
    def addRow(elem: Element) =
```

jvm/src/vyxal/gen/GenerateDocs.scala
```diff
@@ -5,15 +5,17 @@ import vyxal.parsing.Lexer
...
 
   def elements(): String =
     val sb = StringBuilder()
     Elements.elements.values.toSeq
       .sortBy { elem =>
         // Have to use tuple in case of digraphs
         (
-          Lexer.Codepage.indexOf(elem.symbol.charAt(0)),
+          Lexer.Codepage.indexOf(elem.symbol.charAt(0)) +
+            (if "#∆øÞ".contains(elem.symbol.charAt(0)) then 200 else 0),
```
ysthakur, line 18: Is it safe to hardcode 200 and 400?

Lyxal: Yeah. It's just a value to make the value of digraphs higher than other elements. That makes them display at the bottom of the table/elements.txt file. That 200 should be 400 though because hash happens to be less than 256 when added with 200

jvm/src/vyxal/gen/GenerateDocs.scala
Lyxal: Because I wanted to write it neatly and separated from the for each. I wanted to give cleaner looking code a try.

Lyxal: Hence the lambda. I didn't want to use a def, because it's a temp function

jvm/src/vyxal/gen/GenerateDocs.scala
Lyxal: See the other conversation for why I wanted to intentionally use a lambda.

jvm/src/vyxal/gen/GenerateDocs.scala
Lyxal: not  a problrm anymore because it's in the foreach now

ysthakur:
Why the taco?

Lyxal:
image

Lyxal:
This PR is ready for review btw. I'm not adding any new features

jvm/src/vyxal/gen/GenerateDocs.scala
```diff
@@ -57,4 +59,96 @@ private object GenerateDocs:
...
+      .zip(Modifiers.modifiers.values)
+      .toSeq
+      .sortBy((modi, _) =>
+        (
+          Lexer.Codepage.indexOf(modi.charAt(0)) +
+            (if "#∆øÞ".contains(modi.charAt(0)) then 400 else 0),
+          Lexer.Codepage.indexOf(modi.substring(1)),
+        )
+      )
+      .foreach(modi =>
```
ysthakur, line 125: I'd much rather we did `.foreach { case ... =>` here instead of `modi._1` below. It's not immediately clear what `_1` is, so a variable name would help

jvm/src/vyxal/gen/GenerateDocs.scala
```diff
@@ -57,4 +59,96 @@ private object GenerateDocs:
...
+        var trigraph = ""
+        SugarMap.trigraphs
+          .collect { case (tri, s) if s == modi._1 => tri }
+          .foreach { tri => trigraph = tri }
+        contents ++=
+          s"| `${"\\".repeat(if modi._1 == "`" then 1 else 0) +
```
ysthakur, line 131: Might want to make variables outside the interpolated string, and then do `$foo` inside the string. These `${}` blocks are getting a bit big

ysthakur:
I like it, especially the column for trigraphs. Just a couple minor style things

jvm/src/vyxal/gen/GenerateDocs.scala
Lyxal: Shhh don't tell anyone but this was copiloted

jvm/src/vyxal/gen/GenerateDocs.scala
Lyxal: Will do

Lyxal:
pre-commit.ci autofix

ysthakur (approved):

PR #1977, opened 2023-11-15
Add an endpoint to get the version

Lyxal:


Lyxal:
A quick and easy PR, exactly what it says

ysthakur (approved):
I wonder if we could get the version from a git tag. For later, though

Lyxal:
> I wonder if we could get the version from a git tag. For later, though

I looked into this at the time and it requires the github API. And that requires a token, which doesn't really fit well in a static site.

PR #1981, opened 2023-11-17
The last PR before version 3

lyxal:
When this is merged, v3 will be released. Therefore, don't merge it yet :p

Things changed include:

- New codepage files
- A sort of introduction file
- `v` and the superscript v have switched roles
- a method for reading a file as raw bytes has been added
- version retrieval has been moved to the interpreter
- some elements have had their name changed slightly to account for typos or accidental duplication
- tests have been updated to reflect these changes
- the default value of `n` is now the same as v2
- `\`` is now length of stack. Superscript t is the dump modifier.


shared/src/vyxal/Interpreter.scala
```diff
@@ -8,6 +8,7 @@ import scala.collection.mutable.ListBuffer
 import scala.collection.mutable as mut
 
 object Interpreter:
+  def version = "3.0.0"
```
ysthakur, line 11: At some point, we may be able to use https://github.com/sbt/sbt-release instead of hard-coding the version number. That's for later, though, let's not delay v3 further

ysthakur (approved):
LGTM, great work

PR #1988, opened 2023-11-22
Implement a basic set of flags for version 3

### LINKED ISSUE #1987: Explicit output doesn't disable implicit output
Ruhroh

lyxal:
Closes #1987

### DIFF
--- a/shared/src/vyxal/CLI.scala
+++ b/shared/src/vyxal/CLI.scala
@@ -234,31 +234,15 @@ object CLI:
         "Make implicit range generation end at n-1 instead of n",
       ),
       flag('Ṁ', "range-programmery", "Equivalent to having both m and M flags"),
-      flag('v', "vyxal-enc", "Use Vyxal encoding for input file"),
-      flag(
-        'a',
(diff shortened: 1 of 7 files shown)


lyxal:
Okay so what I've tried to do is make it so that an empty program will print the input.

I removed the `if the stack is empty` check before printing.

But now it's erroring. To be investigated.

ysthakur:
What's the error? It doesn't get the next input if the stack is empty?

lyxal:
> What's the error? It doesn't get the next input if the stack is empty?

You can see the error in the test workflow run.

ysthakur:
> You can see the error in the test workflow run.

I think that's just because it now uses `ctx.pop()` instead of `ctx.peek`. Not sure if pop's behavior makes it more desirable, though

lyxal:
> > You can see the error in the test workflow run.
> 
> I think that's just because it now uses `ctx.pop()` instead of `ctx.peek`. Not sure if pop's behavior makes it more desirable, though

This is intentional

PR #2004, opened 2023-12-03
Make ² and ⁻ work with lists

Fmbalbuena:


lyxal (changes requested):
These are supposed to vectorise by default. Making them act on the entire list is not desired behaviour.

ysthakur:
@Fmbalbuena Specifically, the implementations need to use `addPartVect` instead of `addPart` (to vectorize as lyxal said). No other changes are needed

AMiller42:
@ysthakur If these are supposed to vectorise by default, then I don't believe any changes are needed.

[Try it Online!](https://vyxal.github.io/latest.html#WyIiLCIiLCIjW1wiYWJjZFwifFwiZWZnaFwifDY5fDQyMCNdwrIiLCIiLCIiLCIzLjIuMCJd)

ysthakur:
Oh, interesting, I thought addPart didn't handle that

ysthakur:
Closing since the current behavior already matches the intended behavior

PR #2012, opened 2023-12-08
The cooler execute

gingershaped:
Add a JS function to pass more detailed settings to the interpreter

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -44,3 +44,11 @@ jobs:
       - uses: coursier/cache-action@v6
       - name: Compile
         run: sbt vyxalJS/compile
+      - name: Link
+        run: sbt vyxalJS/fastLinkJS
+      - name: Upload JS
+        uses: actions/upload-artifact@v3.1.3
+        with:
+          name: js
+          path: js/target/scala-3.3.0/vyxal-fastopt/
+          
--- a/js/src/vyxal/JSVyxal.scala
+++ b/js/src/vyxal/JSVyxal.scala
@@ -12,6 +12,7 @@ object HelpText:
   @JSExport
   def getHelpText(): String = CLI.helpText
 
+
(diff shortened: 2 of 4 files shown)


js/src/vyxal/JSVyxal.scala
```diff
@@ -54,6 +54,45 @@ object JSVyxal:
...
+      endPrintMode: EndPrintMode = EndPrintMode.Default,
+      defaultValue: String = "0",
+      rangify: Boolean = false,
+      rangeStart: Int = 1,
+      rangeOffset: Int = 0,
+      literate: Boolean = false,
+  ): Unit =
+    val inputList =
+      inputs.map(MiscHelpers.eval(_)(using Context())).toSeq.reverse
+    val settings = Settings(
```
AMiller42, line 67: Maybe I'm just not understanding, but what's the benefit of this over parsing a list of flags like in the old execute?

ex. `val settings = Settings(online = true).withFlags(flags.toList)`

gingershaped: It maps better to the UI design I have in mind; beyond that, it gives the frontend more power to configure Vyxal as it wishes.

.github/workflows/test.yaml
```diff
@@ -44,3 +44,11 @@ jobs:
       - uses: coursier/cache-action@v6
       - name: Compile
         run: sbt vyxalJS/compile
+      - name: Link
+        run: sbt vyxalJS/fastLinkJS
```
lyxal, line 48: Why not `fullLinkJS`?

gingershaped: `fastLinkJS` is faster, and these aren't production builds; they're mostly for checking that the code does what you want without compiling it yourself

lyxal (changes requested):
Testing to see if the bot posts this

Issue #2017, opened 2023-12-15
Fuzzed Issues


AMiller42:
Issues found by the fuzzer.

- [x] Some modifiers (`ᵃᴴ`) error on elements implemented with `addDirect`
- [x] Vectorise (`ᵛ`) errors on nilads
- [x] Rotate (`ṙṘ`) errors on empty strings
- [x] Indexed inputs (`⁰¹`) error if not enough inputs are given
- [x] String Cast to Int (`⌊`) errors on non-numeric strings
- [x] Overlapping Slices (`o`) errors on negative numbers

AMiller42:
Side note: If I understand correctly, I think a lot of the `addDirect` elements could be `addFull` instead, not sure if that should be changed.

ysthakur:
> Side note: If I understand correctly, I think a lot of the `addDirect` elements could be `addFull` instead, not sure if that should be changed.

Yes addFull would be preferabe to addDirect, mpre information

AMiller42:
I was incorrect about the nature of the first mentioned issue. The `ᴴ` modifier was erroring due to typecasting and arity issues, and the `ᵃ` modifier appears to be erroring because the `ȧ` element doesn't work on strings. It just happens that all of the elements I used to test were `addDirect` elements, and at the time I didn't fully understand what that meant.

PR #2021, opened 2023-12-16
Easy fuzzy issues

AMiller42:
Fixing some of the smaller issues from #2017. I foresee the other couple of issues taking a bit longer to fix, but these ones can be merged now.


AMiller42:
Apparently there's also a change to the fuzzer I forgor 💀

PR #2024, opened 2023-12-20
v3: Make fuzz task

ysthakur:
This PR adds a `fuzz` task to the JVM module in Mill so you can do `./mill jvm.fuzz`, and removes the `--fuzz` flag from the normal JVM main method. This means that the fuzzing branch can be merged into the main version-3 branch without fuzzing being exposed to users


lyxal (approved):
🥣

lyxal:
Merge when you're ready

PR #2026, opened 2023-12-20
Death to sbt

ysthakur:
Removes every sbt-related file, as well as documentation about sbt. Long live Mill!


lyxal:
No merging until it's figured out why mill doesn't like showing input taken from readline

PR #2035, opened 2023-12-29
V3 add classes

lyxal:
Mfw Vyxal oop


documentation/Advanced Features/Objects and Extensions.md
```diff
@@ -0,0 +1,269 @@
...
+code. This is not a bad thing, as it makes the implementation of objects much
+simpler.
+
+## Defining Objects
+
+To define an object:
+
+```
+#:O ObjectName | #$restrictedMember #=privateMember <data> #$restrictedMemberWithValue
+<data> #=privateMemberWithValue #!publicMember}
```
ysthakur:
```suggestion
#:O ObjectName |
  #$restrictedMember
  #=privateMember
  <data> #$restrictedMemberWithValue
  <data> #=privateMemberWithValue
  #!publicMember}
```
Not a big deal (you don't have to accept this) but it might be clearer (I was confused at first because I thought the first `<data>` belonged to the first `#=privateMember` before realizing this is a stack-based language)

documentation/Advanced Features/Objects and Extensions.md
```diff
@@ -0,0 +1,269 @@
...
+has struct-like objects. These objects contain only data, and no methods. To add
+methods to an object, you can use extension methods (covered later).
+
+There is also no inheritance, meaning that composition is the only way to reuse
+code. This is not a bad thing, as it makes the implementation of objects much
+simpler.
+
+## Defining Objects
+
+To define an object:
```
ysthakur: The terminology has been confusing me a bit. What you define with `#:O` is a struct, right? Not a prototype like JS? Because calling it an object implies it's a JS-like object that acts as a prototype for other objects, rather than a C-struct-like template for making objects. I feel like it'd be better to refer to say `#:O StructName` rather than `#:O ObjectName`, because `struct` is something a lot of us are familiar with, and is basically the same concept, as far as I can tell

documentation/Advanced Features/Objects and Extensions.md
```diff
@@ -0,0 +1,269 @@
...
+
+To set a member of an object:
+
+```
+## Assuming the object is on the stack  
+"memberName" <value> ŀ
+```
+
+This sets the value of the member to the value on the stack. If the member is a
+restriced or private member, this will fail, unless inside an extension method.
```
ysthakur:
```suggestion
restricted or private member, this will fail, unless inside an extension method.

```

Maybe also say that extension methods will be covered later, and link to them

documentation/Advanced Features/Objects and Extensions.md
```diff
@@ -0,0 +1,269 @@
...
+an extension method that is only available within a certain scope/context.
+* The types of the arguments are checked at runtime. If the types do not match,
+an error will be thrown.
+* Like custom elements, the implementation branch is where the actual code for the element or modifier is written. When called, the method will pop all arguments from the stack
+and execute the implementation as if it were a lambda. The implementation branch can be any valid Vyxal code.
+
+### Order of Extension Method Matching
+
+When a symbol is called, first the interpreter will check for an extension method
+with the appropriate types of items on the stack. If found, it will execute the
```
ysthakur: Might be worth mentioning the phrases "multimethods" or "multiple dispatch" somewhere, to help people who already understand those concepts (as far as I can tell, "multiple dispatch" is an accurate way of describing the type-based overloading Vyxal does). Maybe also throw in a bit about *why* we do type-based overloading (better to have the language do it for you than do it manually, safer than languages like Python which let you do whatever, blah blah blah). Maybe mention Julia?

documentation/Advanced Features/Objects and Extensions.md
```diff
@@ -0,0 +1,269 @@
...
+  #$mp
+}
+
+#$Map Ė
+"key1" "value1" #:@put
+```
+
+This example is the same as the previous example, except that it uses an extension
+method instead of a lambda. Calling `put` with any other value for `mp` that is
+not a `Map` object will not execute the extension method.
```
ysthakur:
```suggestion
not a `Map` object will throw an error since no `put` method has been
defined for that value's type.

```

documentation/Advanced Features/Objects and Extensions.md
```diff
@@ -0,0 +1,269 @@
...
+  argA as typeA,
+  argB as typeB,
+  ...
+  argN as typeN
+does
+    implementation
+}
+```
+
+_Note that any branch keyword can be used in place of `does`, `as`, `given` and `,`._
```
ysthakur:
```suggestion
> [!note]
> Any branch keyword can be used in place of `does`, `as`, `given` and `,`._

```
This isn't necessary but it's kinda cool. Here's what it looks like:

> [!note]
> Any branch keyword can be used in place of `does`, `as`, `given` and `,`._

shared/src/vyxal/Elements.scala
```diff
@@ -187,6 +179,14 @@ object Elements:
             case _ => throw InvalidListOverloadException("Ạ", b, "Number")
         if a.isInstanceOf[String] then temp.mkString
         else temp
+      case (a, b: VList, c) =>
+        var temp = ListHelpers.makeIterable(a)
+        for i <- ListHelpers.makeIterable(b) do
+          i match
+            case ind: VNum => temp = ListHelpers.assign(temp, ind, c)
+            case _ => throw InvalidListOverloadException("Ạ", b, "Number")
```
ysthakur:
```suggestion
        val temp = ListHelpers.makeIterable(b).foldLeft(ListHelpers.makeIterable(a)) {
          case (temp, ind: VNum) => ListHelpers.assign(temp, ind, c)
          case _ => throw InvalidListOverloadException("Ạ", b, "Number")
        }
```
No idea if it works but it probably does

shared/src/vyxal/Elements.scala
```diff
@@ -1080,6 +1085,13 @@ object Elements:
               temp = ListHelpers.insert(temp, index, elem)
             case _ => throw InvalidListOverloadException("Ị", b, "Number")
         temp
+      case (a, b: VList, c) =>
+        var temp = ListHelpers.makeIterable(a)
+        for i <- b.reverse do
+          i match
+            case index: VNum => temp = ListHelpers.insert(temp, index, c)
+            case _ => throw InvalidListOverloadException("Ị", b, "Number")
+        temp
```
ysthakur:
```suggestion
        b.foldRight(ListHelpers.makeIterable(a)) {
          case (index: VNum, acc) => ListHelpers.insert(acc, index, c)
          case _ => throw InvalidListOverloadException("Ị", b, "Number")
        }
```

shared/src/vyxal/Elements.scala
```diff
@@ -2525,6 +2537,17 @@ object Elements:
         val temp =
           ListHelpers.transliterate(ListHelpers.makeIterable(a), b, c).mkString
         if VNum.NumRegex.matches(temp) then VNum(temp) else temp
+      case (a: VObject, b: String, c) => MiscHelpers.setObjectMember(a, b, c)
+      case (a: VObject, b: VList, c) =>
+        var obj = a.copy()
```
ysthakur: Is it necessary to copy it right here if `setObjectMember` won't mutate it anyway?

shared/src/vyxal/Elements.scala
```diff
@@ -2525,6 +2537,17 @@ object Elements:
...
+        for i <- b.lst.indices do
```
ysthakur: Why iterate over the indices and not the list itself?

shared/src/vyxal/Exceptions.scala
```diff
@@ -113,6 +113,35 @@ class UnimplementedOverloadException(element: String, args: Seq[VAny])
 class VyxalRecursionException()
     extends VyxalRuntimeException("Too many recursions")
 
+class IterificationOfNonIterableException(value: VAny)
+    extends VyxalRuntimeException(s"Cannot iterify $value")
+
+class BadArgumentException(message: String, arg: VAny)
+    extends VyxalRuntimeException(s"$message received bad argument: $arg")
+
+/** Class related exceptions */
```
ysthakur, line 122: We could make a `class VyxalClassException(msg: String) extends VyxalRuntimeException(msg)`. Not sure if we'll ever need to catch every kind of class-related exception though

shared/src/vyxal/Globals.scala
```diff
@@ -25,6 +25,13 @@ case class Globals(
     String,
     CustomDefinition,
   ] = Map()
+  var classes: Map[
+    String,
+    CustomClass,
+  ] = Map()
```
ysthakur: This formatting looks off. Did scalafmt split this up onto multiple lines? If it's broken, we might want to file a bug report

shared/src/vyxal/Globals.scala
```diff
@@ -36,7 +43,7 @@ class Inputs(origInputs: Seq[VAny] = Seq.empty):
 
   /** Uses an array for constant access, not for mutating items */
   private var currInputs = origArr
-  private var allInputs = origInputs.reverse
+  private val allInputs = origInputs.reverse
```
ysthakur, line 39: This is the best day of my life

shared/src/vyxal/Interpreter.scala
```diff
@@ -104,16 +106,30 @@ object Interpreter:
...
             case Some(implementation) =>
               val lam = VFun.fromLambda(implementation.asInstanceOf[AST.Lambda])
               if implementation.arity.getOrElse(0) == -1 then executeFn(lam)
               else ctx.push(executeFn(lam))
-            case None => throw VyxalYikesException(
-                s"Symbol $cmd has no impl. Should have already been caught :skull:"
-              )
-        else
+              executed = true
+            case None => ()
```
ysthakur, line 131: Can't this if block go into the `case None =>` above? And then the `if !executed` block below can go into the `case None =>` for this one

shared/src/vyxal/Interpreter.scala
```diff
@@ -279,7 +297,7 @@ object Interpreter:
       vars: mut.Map[String, VAny] = mut.Map(),
   )(using ctx: Context): VAny =
     val VFun(_, arity, params, origCtx, lambda, _) = fn
-    var originallyFunction = false
+    val originallyFunction = false
```
ysthakur, line 300: 🎉🎊🎉🥳

shared/src/vyxal/Interpreter.scala
```diff
@@ -354,4 +372,43 @@ object Interpreter:
     scribe.trace(s"Result of executing function: $res")
     res
   end executeFn
+
+  def createObject(con: VConstructor)(using ctx: Context): VObject =
+    val fields = ctx.globals.classes(con.name).fields
+    ctx.privatable += con.name
+    val assignedFields = mut.Map[String, (Visibility, VAny)]()
```
ysthakur, line 379: I'd prefer a `Field` case class over `(Visibility, VAny)`, but I don't feel too strongly about it, so feel free to resolve this comment

shared/src/vyxal/Context.scala
```diff
@@ -40,6 +41,7 @@ class Context private (
     val testMode: Boolean = false,
     val useStack: Boolean = false,
     var recursion: Int = 0,
+    val privatable: ListBuffer[String] = ListBuffer(),
```
ysthakur, line 47: Could you add info on what `privatable` is to the doc comment above?

shared/src/vyxal/ListHelpers.scala
```diff
@@ -402,7 +402,7 @@ object ListHelpers:
     val size = iterable.length
     val pieceSize = (size / pieces).floor
     var remaining = iterable
-    var out = ListBuffer.empty[VList]
+    val out = ListBuffer.empty[VList]
```
ysthakur, line 405: 🚀

shared/src/vyxal/VAny.scala
```diff
@@ -126,7 +128,17 @@ extension (self: VAny)
...
+      case o: VObject => true
 end extension
 
+case class VConstructor(
+    name: String
+)
+
+case class VObject(
+    className: String,
+    fields: mut.Map[String, (Visibility, VAny)],
```
ysthakur: Given that setting a field always makes a new copy of the object, it'd probably be better to make this an immutable map instead

shared/src/vyxal/MiscHelpers.scala
```diff
@@ -160,6 +169,38 @@ object MiscHelpers:
...
+  def setObjectMember(obj: VObject, name: String, value: VAny)(using
+      ctx: Context
+  ): VObject =
+    val (visibility, _) = obj.fields
+      .getOrElse(name, throw FieldNotFoundException(obj.className, name))
+
+    val objName = obj.className
+    val newObj = obj.copy()
+    visibility match
+      case Visibility.Public => newObj.fields(name) = (visibility, value)
```
ysthakur: > [!warning]
> `.copy()` only does a shallow copy, not a deep copy. This means that `newObj.fields` is the same object as the old `obj.fields` (unless you've overridden it somewhere, but I didn't see that). Make `fields` an immutable map, and create a new `fields` map here instead of setting a single value. Also, we need to add tests to make sure setting a field only returns a new object but doesn't mutate existing objects

That didn't need an admonition but they're so cool

documentation/Advanced Features/Objects and Extensions.md
lyxal: You're right that it's more like a struct, but I wanted to avoid calling it that so as to not create confusion with the terminology used for control flow structures. Maybe there's a better term available that I don't know of, like record.

ysthakur: Record sounds good to me. I think record might imply all members are public to some, but whatever

shared/src/vyxal/Elements.scala
lyxal: when you put it that way, no

shared/src/vyxal/Elements.scala
lyxal: Because I need the indices

shared/src/vyxal/Exceptions.scala
lyxal: I think it's fine the way it is - only the specific exceptions I defined are used.

shared/src/vyxal/Elements.scala
ysthakur: Doesn't this work?
```scala
for elem <- b do
  obj = MiscHelpers.setObjectMember(obj, elem.toString, c)
```

shared/src/vyxal/Globals.scala
lyxal: copilot probably

shared/src/vyxal/Globals.scala
lyxal: no.

lyxal: oh wait this isn't a suggestion it's a comment on a change

ysthakur: Did you assume I was asking you to change a `var` to a `val` lol

shared/src/vyxal/Interpreter.scala
```diff
@@ -354,4 +372,43 @@ object Interpreter:
...
+      val fieldVal =
+        if predef.isDefined then
+          Interpreter.executeFn(
+            VFun
+              .fromLambda(predef.get.asInstanceOf[AST.Lambda])
+              .copy(
+                arity = 0
+              )
+          )
+        else ctx.pop()
```
ysthakur:
```suggestion
        predef match
          case Some(predef) =>
            Interpreter.executeFn(
              VFun
                .fromLambda(predef.asInstanceOf[AST.Lambda])
                .copy(
                  arity = 0
                )
            )
          case None => ctx.pop()
```

shared/src/vyxal/parsing/Lexer.scala
```diff
@@ -186,7 +188,8 @@ object Lexer:
 
     // Now, bind the move right tokens to the next token
 
-    var bound = ListBuffer[LitToken | (LitToken, Int)]()
+    val bound: ListBuffer[LitToken | (LitToken, Int)] =
+      ListBuffer[LitToken | (LitToken, Int)]()
```
ysthakur:
```suggestion
    val bound = ListBuffer[LitToken | (LitToken, Int)]()
```
The type annotation *should* be unnecessary, could you see if this works?

shared/src/vyxal/parsing/Parser.scala
```diff
@@ -3,6 +3,7 @@ package vyxal
 import scala.language.strictEquality
```
ysthakur, line 3: The package statement at the top is wrong, it should be `package vyxal.parsing`, but we can fix that later sincei t'd involve some (minor) refactoring

ysthakur (approved):
Just one thing: if any of the members have initializers, why wrap those in lambdas? Why not leave them as expressions? The interpreter can directly execute `AST`s too, right?

PR #2037, opened 2024-01-05
Set XOR: don't uniquify the result

RubenVerg:


ysthakur (approved):
LGTM, thanks

@lyxal: the code itself looks good, so feel free to merge if the behavior is desired

PR #2039, opened 2024-01-06
Set Union: don't uniquify the result

RubenVerg:
Changes behavior to be like that of APL [Union](https://aplwiki.com/wiki/Union)


ysthakur (approved):
LGTM, thanks

PR #2040, opened 2024-01-06
v3: Make set intersection act like APL

ysthakur:
Makes set intersection not uniquify, like APL


shared/src/vyxal/VList.scala
```diff
@@ -182,12 +182,6 @@ class VList private (val lst: Seq[VAny])
         true
     })
 
-  @targetName("setDiff")
-  def -(other: VList): VList =
```
ysthakur, line 186: Unrelated to this PR, but it turns out `filterNot` can do the job slightly more concisely and it's only used in one place (credit to Ruben Verg for the change)

lyxal (approved):

PR #2042, opened 2024-01-07
The good update pr

lyxal:
Nowhere near ready. This is for the workflows.


lyxal:
Nuh uh

So the plan is to put the lexers from here into a 3.4.2 release and then do the smart big brain changes fresh from there.

PR #2043, opened 2024-01-07
Fix list ordering `compare`

RubenVerg:
I left the behavior where lists are first compared by length, and only if they have the same length are they compared by contents.


mathcat4 (approved):
lgtm

shared/test/resources/tests.yaml
```diff
@@ -330,6 +330,8 @@
       in: [[8, 9, "acb", "abc", 1, 9, 2, [1, 2]]],
       out: [1, 2, 8, 9, 9, [1, 2], "abc", "acb"],
     }
+  - { in: [[[1, 2, 3], [1, 2, 4]]], out: [[1, 2, 3], [1, 2, 4]] }
```
ysthakur:
```suggestion
  - { in: [[[1, 2, 4], [1, 2, 3]]], out: [[1, 2, 3], [1, 2, 4]] }
```
Just so we know that something is actually happening, in case `compare` somehow returns 0 and the result is only correct because the sort is stable

gingershaped (approved):
very gaming

PR #2044, opened 2024-01-07
Fix complex numbers

RubenVerg:
~~Breaks `F` with complex inputs (and potentially also other things that aren't tested with complex nums? testing with complex nums seems to be sporadic), not sure how to fix it because I don't really understand how it works.~~ Changed the test (was told it was wrong)

~~Complex `%` afaik could be defined but it's not as easy as the other operations are.~~ Actually turns out to be surprisingly easy, was a minor change to not discard the complex part of the divison result

For ordering complex numbers I chose lexicographical ordering (compare the real parts, if they are equal compare the imaginary parts)


lyxal (approved):
I see complex number operations were never implemented properly.

ysthakur (approved):
Nice job. We (okay, mostly my fault) were kinda lazy when implementing operations on complex numbers and decided it was simpler to use the real parts for everything instead

PR #2045, opened 2024-01-07
Avoid stack overflow when compiling lexers

ysthakur:
I think all the chained `|`s in `SBCSLexer.token` and `LiterateLexer.singleToken` were blowing the stack when compiling, because the macro for handling `|` involves a quote within a splice within a quote, and I couldn't find any other sources of a possible stack overflow (aside from `StringIn`, but all our usages of `StringIn` involve a small number of small strings, nothing to make a really deep trie)

### DIFF
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -23,9 +23,6 @@ concurrency:
   group: "pages"
   cancel-in-progress: true
 
-env:
-  JAVA_OPTS: ${{ vars.JAVA_OPTS }}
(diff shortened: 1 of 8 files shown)


lyxal (changes requested):
Well we might as well put the variable tokens into their own parser too because they can be factored out

ysthakur:
Sorry @lyxal, I'm merging without an approval because the test workflow on https://github.com/Vyxal/Vyxal/pull/2046 is failing. We can make changes to the lexer later

PR #2047, opened 2024-01-08
Export lexing to the JavaScript module

RubenVerg:
Because `Lexer`.`lexLiterate` turns `Token`.`value` into its SBCS counterpart, the exported `vyxal`/`Vyxal`.`lexLiterate` does the same. Original input can however be trivially reconstructed using the `startOffset` and `endOffset` properties on `JSToken`


ysthakur (approved):
LGTM, thanks

PR #2050, opened 2024-01-09
Change list evaluation to collecting all pushed items on a new stack

RubenVerg:

### DIFF
--- a/shared/src/vyxal/Interpreter.scala
+++ b/shared/src/vyxal/Interpreter.scala
@@ -100,11 +100,10 @@ object Interpreter:
       case AST.CompressedString(value, _) =>
         ctx.push(StringHelpers.decompress252String(value))
       case AST.Lst(elems, _) =>
-        val list = collection.mutable.ListBuffer.empty[VAny]
-        for elem <- elems do
-          execute(elem)(using ctx.makeChild())
-          list += ctx.pop()
-        ctx.push(VList.from(list.toList))
+        val context = ctx.copy
+        context.clear()
+        for elem <- elems do execute(elem)(using context)
+        ctx.push(VList.from(context.getStack))
       case AST.Command(cmd, _, overwriteable) =>
         var executed = false
         if overwriteable && ctx.globals.extensions.contains(cmd) then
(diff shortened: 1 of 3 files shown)


lyxal (changes requested):
Make sure to add tests in InterpreterTests

lyxal (approved):

lyxal:
I'll wait for user to have a check before merging

shared/src/vyxal/Interpreter.scala
```diff
@@ -100,11 +100,10 @@ object Interpreter:
       case AST.CompressedString(value, _) =>
         ctx.push(StringHelpers.decompress252String(value))
       case AST.Lst(elems, _) =>
-        val list = collection.mutable.ListBuffer.empty[VAny]
-        for elem <- elems do
-          execute(elem)(using ctx.makeChild())
-          list += ctx.pop()
-        ctx.push(VList.from(list.toList))
+        val context = ctx.copy
+        context.clear()
```
ysthakur, line 104: Maybe we could add a `stack` parameter to `Context#copy` to avoid having to clear later?

ysthakur:
Can list elements still access what was on the stack before the list literal started executing? If not, maybe we could do something closer to the old code, except record the length of the stack before each element starts executing so we can pop the appropriate number of elements? But that would get messed up if you popped something that was on the stack already and pushed something else (e.g. `3 #[ 2 + #]`).

Maybe if we could make a child context that only uses the parent context's when its own stack runs out of elements? But that's not how child contexts work right now

RubenVerg:
@ysthakur lxyal and I agreed that allowing access to the outer stack leads to code that's hard to decide what should happen such as your example, so we opted for using a separate, empty-by-default stack. Maybe an alternative could be a parent stack that's read-only, so for example `3 #[1 + #]` would use the `3` but not pop it from the stack. That however seems much harder to implement, and it doesn't seem that useful. If you want to access things from the outer stack, use the primitives that build lists

ysthakur (approved):
Ah ok, makes sense then. LGTM

PR #2051, opened 2024-01-09
Implement trigonometric & hyperbolic functions

RubenVerg:
~~Some of the new tests don't pass; I believe this is due to an issue with Spire, tracked at typelevel/spire#1313. Once the issue is fixed I'll update the Spire dependency and the PR can be approved.~~ I changed the functions to not depend on the bugged functions.


shared/src/vyxal/VNum.scala
```diff
@@ -55,11 +55,31 @@ class VNum private (val underlying: Complex[Real]) extends Ordered[VNum]:
...
   def to(end: VNum, step: VNum = 1): VList =
     VList.from(Inclusive(this, end, step))
 
+  def sin: VNum = underlying.sin
+  def cos: VNum = underlying.cos
+  def tan: VNum = underlying.tan
+  def asin: VNum = underlying.asin
+  def acos: VNum = underlying.acos
+  def atan: VNum = underlying.atan
+  // def atan2(rhs: VNum): VNum = spire.math.atan2(underlying, rhs.underlying)
```
ysthakur, line 66: Should probably leave a "TODO" here along with a link to https://github.com/typelevel/spire/issues/1313 so we can come back to this when the Spire bug is fixed

shared/src/vyxal/VNum.scala
```diff
@@ -55,11 +55,31 @@ class VNum private (val underlying: Complex[Real]) extends Ordered[VNum]:
...
+  def atan2(rhs: VNum): VNum =
+    // atan2(a, b) = -i * ln((b + i*a)/sqrt(a^2 + b^2)), according to WolframAlpha
+    Complex[Real](0, -1) *
+      spire.math.log(
+        (rhs.underlying + underlying * Complex[Real](0, 1)) /
+          spire.math.sqrt(underlying ** 2 + rhs.underlying ** 2)
+      )
+  def sinh: VNum = underlying.sinh
+  def cosh: VNum = underlying.cosh
+  // def tanh: VNum = underlying.tanh
```
ysthakur, line 77: Also leave a todo here

shared/test/resources/tests.yaml
```diff
@@ -1501,3 +1501,141 @@
 "Þi":
   - {in: [[1,[2,3]],[1,0]], out: 2}
   - {in: [["xyzabc"], [0,4]], out: "b"}
+
+"∆s":
+  - in:
+    - - 3.1415926535897932384626433832795028841971693993751058209749445923 # π
```
ysthakur, line 1530: Do the tests fail if you only use, say, 7 digits? Although how long we make our numbers doesn't matter much either way

shared/test/resources/tests.yaml
```diff
@@ -1501,3 +1501,141 @@
...
+  - in:
+    - - 1
+      - -1
+      - !num 0ı1
+      - !num 1ı1
+    out:
+      - 0
+      - 3.1415926535897932384626433832795028841971693993751058209749445923
+      - 1.5707963267948966192313216916397514420985846996875529104874722961
+      - 0.7853981633974483096156608458198757210492923498437764552437361480
```
ysthakur:
```suggestion
      - 0.7853981633974483096156608458198757210492923498437764552437361480

```

ysthakur (commented):
LGTM, just add "TODO" to `atan2` and `tanh` so we can come back to them later. Those bugs are unfortunate but nice job working around them

ysthakur (approved):
asdfasdf

PR #2053, opened 2024-01-09
dont fix a thing that wasnt just added again

pacmanboss256:

### DIFF
--- a/shared/src/vyxal/VNum.scala
+++ b/shared/src/vyxal/VNum.scala
@@ -25,9 +25,9 @@ class VNum private (val underlying: Complex[Real]) extends Ordered[VNum]:
   def isValidInt: Boolean = underlying.real.isValidInt
 
   /** Whether there is an imaginary and a real part */
-  def isComplex: Boolean = (underlying.imag != 0 && underlying.real != 0)
+  def isComplex: Boolean = ((underlying.imag != 0 && underlying.real != 0) || (underlying.imag == 0 && underlying.real == 0))
   /** Whether there is only an imaginary part */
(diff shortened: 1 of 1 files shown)


ysthakur (commented):
I feel like both of those should return true for (0, 0)

a-a-GiTHuB-a-a:
why

shared/src/vyxal/VNum.scala
```diff
@@ -24,8 +24,10 @@ class VNum private (val underlying: Complex[Real]) extends Ordered[VNum]:
   /** Whether the real part is small enough to be converted to an `Int` */
   def isValidInt: Boolean = underlying.real.isValidInt
 
-  /** Whether there is an imaginary part */
-  def isComplex: Boolean = underlying.imag != 0
+  /** Whether there is an imaginary and a real part */
+  def isComplex: Boolean = ((underlying.imag != 0 && underlying.real != 0) || (underlying.imag == 0 && underlying.real == 0))
+  /** Whether there is only an imaginary part */
+  def isImaginary: Boolean = (underlying.real == 0)
```
ysthakur: How about `isReal` and `isImaginary`, since those would complement each other? `isComplex` would then be `a.isReal ^ a.isImaginary`
```suggestion
  /** Whether there is only a real part */
  def isReal: Boolean = underlying.isReal
  /** Whether there is only an imaginary part */
  def isImaginary: Boolean = underlying.isImaginary
```
Spire already has those implemented too:

https://github.com/typelevel/spire/blob/ac844d3138e7881f753fc2b89113092d89208567/core/src/main/scala/spire/math/Complex.scala#L138-L139

ysthakur (approved):
LGTM, thanks for the changs

PR #2055, opened 2024-01-10
Implement pretty printing and proper stringification for records, constructors and lambdas.

RubenVerg:


shared/src/vyxal/VAny.scala
```diff
@@ -134,11 +139,21 @@ end extension
...
+  override def toString = s"$name constructor"
 
 case class VObject(
     className: String,
     fields: Map[String, (Visibility, VAny)],
-)
+):
+  override def toString =
+    val fs = fields.map {
+      case (name, (vis, value)) => s"$value ${vis match
```
ysthakur: I'd kinda prefer to have the match outside the string, assigned to a variable `prefix` or something

lyxal (changes requested):
This only impacts when `toString` is called - printing a record doesn't call to string. This'll need to be changed.

RubenVerg:
@lyxal fixed, thanks for spotting that

lyxal:
For records, maybe something like:

```
<Name> { 
  !Public: value,
  $Restricted: value,
  =Private: value
}
```

For example,

```
<Map> {
  $keys: ["k1", "k2", "k3"],
  $values: ["v1", "v2", "v3"]
}
```

RubenVerg:
@lyxal with the newlines and indentation? that's gonna become an issue when there are maps nested inside maps, it's gonna look wrong

lyxal:
How about outermost map has indentation, and inner maps are flat

ysthakur:
I feel like we should either go all the way with flatness or all the way with indentation. Maybe `toString` could be completely flat, but when displaying in the REPL, we pretty-print everything (including lists?) with indentation? And in the web interpreter, it could show objects like the web console shows JS objects, with dropdowns and stuff

lyxal (approved):

shared/src/vyxal/AST.scala
```diff
@@ -184,6 +190,12 @@ enum Visibility derives CanEqual:
   case Private
   case Restricted
 
+  def sigil: String =
+    this match
```
ysthakur: You could also make this a field by adding it to visibility,s constru tor

ysthakur:
```scala
enum Visibility(val sigil: String) derives CanEqual:
  case Public extends Visibility("!")
  case Private extends Visibility("=")
  case Restricted extends Visibility("$")
```

ysthakur (approved):
LGTM, this should make working in the REPL more pleasant. I was going to suggest changing the interpolated strings to not be multiline by extracting the expressions into variables, but tbh, it doesn't matter much

PR #2057, opened 2024-01-10
Fix REPL stops printing after a print element is ran

RubenVerg:
This fixes a bug where after executing a print element in a REPL, all the next input lines don't get the automatically-print-the-top-of-the-stack behavior.


ysthakur (approved):
LGTM. I was going to suggest setting it to false inside `Interpreter.execute`, but that may mess with executing functions at the end before you print the results

PR #2058, opened 2024-01-11
rename drop to drop/zero slice from

TheSecondComing123:

### DIFF
--- a/shared/src/vyxal/Elements.scala
+++ b/shared/src/vyxal/Elements.scala
@@ -1071,8 +1071,8 @@ object Elements:
     addPart(
       Dyad,
       "İ",
-      "Drop | Collect While Unique | Complex Number",
-      List("drop", "collect-while-unique", "complex"),
+      "Zero Slice From | Collect While Unique | Complex Number",
+      List("zero-slice-from", "slice-from", "from", "drop-from", "collect-while-unique", "complex"),
       false,
       "a: num, b: num -> a.real + b.real * i",
       "a: str|lst, b: num -> a[b:]",
--- a/shared/src/vyxal/Elements.scala
+++ b/shared/src/vyxal/Elements.scala
@@ -1071,8 +1071,8 @@ object Elements:
     addPart(
       Dyad,
       "İ",
-      "Zero Slice From | Collect While Unique | Complex Number",
-      List("zero-slice-from", "slice-from", "from", "drop-from", "collect-while-unique", "complex"),
+      "Drop/Zero Slice From | Collect While Unique | Complex Number",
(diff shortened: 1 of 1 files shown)


shared/src/vyxal/Elements.scala
```diff
@@ -1071,8 +1071,8 @@ object Elements:
     addPart(
       Dyad,
       "İ",
-      "Drop | Collect While Unique | Complex Number",
-      List("drop", "collect-while-unique", "complex"),
+      "Zero Slice From | Collect While Unique | Complex Number",
+      List("zero-slice-from", "slice-from", "from", "drop-from", "collect-while-unique", "complex"),
```
ysthakur:
```suggestion
      "Drop | Collect While Unique | Complex Number",
      List("zero-slice-from", "slice-from", "drop", "collect-while-unique", "complex"),
```
The `zero-slice-from` and `slice-from` keywords are a good idea, but `drop` is also a common name for this operation (also, I've never seen anyone call it `drop-from`)

ysthakur (changes requested):
Thanks for the PR! Just a tiny change (you can use the "Commit suggestion" button to apply it)

shared/src/vyxal/Elements.scala
TheSecondComing123: Personally, I've never heard of `drop`. I like the keywords, but maybe we could call it "Zero Slice From/Drop"?

RubenVerg: Drop is definitely the most common name, for example [Scala](https://www.scala-lang.org/api/current/scala/collection/IterableOnceOps.html#drop-aba), [Haskell](https://hackage.haskell.org/package/base-4.19.0.0/docs/Prelude.html#v:drop), and [APL](https://apl.wiki/Drop) all call it "drop".

TheSecondComing123: @RubenVerg Alright, I've changed it to Drop/Zero Slice From.

ysthakur (approved):
Calling it just Drop would've been too but I guess adding `/Zero Slice From` doesn't hurt. Nice job!

TheSecondComing123:
Wait, there's still a problem. Some of the keywords aren't implemented in literate mode (e. g. `zero-slice-from` turns into `zero-slice-|`.

ysthakur:
What error are you getting? I tried locally and it's lexed correctly

TheSecondComing123:
@ysthakur image

It's not exactly erroring, but the output isn't correct either.

TheSecondComing123:
@ysthakur  Oh woah it suddenly worked I must have tested it wrong.

ysthakur:
Awesome, good to hear

PR #2061, opened 2024-01-12
Edge complex fix

pacmanboss256:
fix edges for complex numbers


shared/src/vyxal/Elements.scala
```diff
@@ -1010,10 +1010,12 @@ object Elements:
       false,
       "a: lst -> [a[0], a[-1]]",
       "a: str -> [a[0], a[-1]]",
-    ) { a =>
-      val iterable = ListHelpers.makeIterable(a)
-      if iterable.isEmpty then VList.from(Seq.empty)
-      else VList(iterable.head, iterable.last)
+    ) {
+      case a: VNum if !(a.isImaginary || a.isReal) => VList(a.real, a.imag)
```
ysthakur: Hold on, if I have a list of complex numbers, some of which have 0 as either their real or imaginary component, then calling edges on each element in that list will have wildly different behavior for some of those numbers. Maybe you could just do this for all `VNum`s, whether or not they have nonzero real and imaginary components?

shared/src/vyxal/Elements.scala
```diff
@@ -1010,10 +1010,12 @@ object Elements:
       false,
       "a: lst -> [a[0], a[-1]]",
       "a: str -> [a[0], a[-1]]",
```
ysthakur, line 1031: You're adding new behavior, so be sure to add an overload here, something like `"a: num -> blahb lah"`

shared/src/vyxal/Elements.scala
lyxal: @ysthakur I don't think that's a problem - for a real number, getting the first and last digits is sometimes useful. But that doesn't make as much sense for complex numbers, where the imaginary part can be completely unrelated to the real part.

ysthakur: Right, but suppose you have a program that generates a list of complex numbers and you want to get the real and imaginary part of each. Only problem is, some of those complex numbers happen to have 0 as the real component. Maybe something like `[0i1, 1i2, 2i3]`. What you expect is `[[0, 1], [1, 2], [2, 3]]`, but what you get instead is `[[0, 0], [1, 2], [2, 3]]`. To get around that, you would need to manually handle the case when the real or imaginary component is 0

pacmanboss256: @ysthakur you would still get 0,1 but it would break for 0+bi where b>9

shared/src/vyxal/Elements.scala
pacmanboss256: fixed this, potential documentation standards for complex numbers?

ysthakur: I can't think of anything specific. Maybe `[a.real, a.imag]`?

shared/src/vyxal/Elements.scala
pacmanboss256: fixed this

ysthakur: It still won't work if your list of complex numbers happens to have a number whose imaginary component is 0, like `2i0`. I'd rather make this get the real and imaginary components no matter what. If someone wants the first and last digits (which seems less useful, but correct me if I'm wrong), they can get the digits and then call ends on that (only 1 more byte)

pacmanboss256: my plan is to add an overload anyways for a complex number that splits it into parts because thats more useful tbb

lyxal (approved):

PR #2063, opened 2024-01-13
Make trigraphs in digraphs work

ysthakur:
Now digraphs whose first or second character is a trigraph should work


lyxal (changes requested):
Where tests

ysthakur:
@lyxal Test added

lyxal (approved):

PR #2064, opened 2024-01-14
Add radians-to-degrees and degrees-to-radians elements

RubenVerg:


lyxal (approved):

lyxal:
In the future, make sure not to add tests in two PRs to the same area - it causes merge conflicts otherwise.

PR #2067, opened 2024-01-15
Add support for non-integer and negative arguments to `ċ`

RubenVerg:
Also changed a test that said 3 choose 3 should be 0, as it is usually defined to be 1.


lyxal (approved):

lyxal:
Merge when you're ready

PR #2069, opened 2024-01-16
Add `reshape` function & `ÞR` element

RubenVerg:


shared/src/vyxal/ListHelpers.scala
```diff
@@ -854,6 +854,21 @@ object ListHelpers:
     current
   end reduce
 
+  def reshape(iterable: VList, shape: Seq[VNum]): VAny =
+    var index = -1
```
ysthakur: Why not use a `iterable.iterator` here? If the backing list is a linked list, indexing into it over and over again could be inefficient. I'm assuming an iterator would be more performant. When you hit the end of the iterator, you could make a new iterator since you have the original `VList` available

shared/src/vyxal/Elements.scala
```diff
@@ -3928,6 +3928,20 @@ object Elements:
     ) {
       case a: VNum => a / VNum(spire.math.Real.pi) * VNum(180)
     },
+    addPart(
+      Dyad,
+      "ÞR",
+      "Reshape",
+      List("reshape"),
+      false,
+      "a: lst, b: lst[num] => a reshaped to shape b",
```
ysthakur, line 3937: Just to be clear, this only works if `a` is a flat list, right? If I want to reshape a matrix, I'll have to flatten it first and then reshape it?

shared/src/vyxal/Elements.scala
```diff
@@ -3928,6 +3928,20 @@ object Elements:
...
+    ) {
+      case (a, b: VList) if b.forall(_.isInstanceOf[VNum]) =>
+        ListHelpers
+          .reshape(ListHelpers.makeIterable(a), b.map(_.asInstanceOf[VNum]))
```
ysthakur: I think maybe we could accept any kind of lists instead of having a guard, and only check the shape later so we could tell the user which particular element was not a number
```suggestion
      case (a, b: VList) =>
        val shape = b.map {
          case n: VNum => n.toInt
          case x => throw element x is not an int expectedo nly ints for reshape or something
        }.toList // To force it, idk if there's a better way to force generically (LazyList does have a `force` method but it's not on Seq)
        ListHelpers
          .reshape(ListHelpers.makeIterable(a), shape)
```

shared/src/vyxal/ListHelpers.scala
```diff
@@ -854,6 +854,21 @@ object ListHelpers:
...
+      index += 1
+      if !iterable.isDefinedAt(index) then
+        index =
+          0 // avoid %= iterable.length to hopefully work with infinite lists
+      iterable(index)
+    def go(shape: Seq[Int]): VAny =
+      if shape.isEmpty then iterable(0)
+      else if shape.length == 1 then
+        VList.from((0 until shape.head).map { _ => nextElement() })
+      else VList.from((0 until shape(0)).map { _ => go(shape.tail) })
```
ysthakur: `VList.fill` should work here (or at least, `VList.from(Seq.fill(shape.head)(...))`)

shared/src/vyxal/Elements.scala
RubenVerg: that is correct, not sure how to specify in the docs

ysthakur: Maybe something like `a (assumed flat) reshaped to shape b`? But then people might think you can't use it on nested lists at all

ysthakur (approved):
Doc comment's not a huge deal so I'm approving, merge when you like. Good work

PR #2071, opened 2024-01-18
Allow forward references to custom elements

ysthakur:
This PR adds another phase to the parser that finds extensions before the command grouping and modifier stuff is done so that forward references to them can be made.

Not done, also need to test


ysthakur:
This is useless since https://github.com/Vyxal/Vyxal/pull/2042 will redo the parser and allow forward references anyway

PR #2072, opened 2024-01-18
Add APL-style (multidimensional) ranges

RubenVerg:


shared/src/vyxal/Elements.scala
```diff
@@ -2492,19 +2492,23 @@ object Elements:
...
-      true,
       "a: num -> [1..a]",
+      "a: lst[num] -> apl-style iota from 1 to a",
       "a: str -> a.upper()",
     ) {
       case a: VNum => NumberHelpers.range(1, a)
-      case a: String => a.toUpperCase
+      case a: VList if a.forall(_.isInstanceOf[VNum]) =>
+        NumberHelpers.range(1, a.map(_.asInstanceOf[VNum]))
+      case a => (Monad.vectorise("ɾ") {
```
ysthakur: This makes a new function every time. I'd recommend putting something like this in `StringHelpers`:

```scala
val uppercase = Monad.vectorise("ɾ") {
  case a: String => a.toUppercase
}
```

ysthakur: In fact, you don't even need to manually vectorise. Since you have that guard up there to only catch `VList`s that only have numbers, you're all good. Just make this last case `case a: String => a.toUpperCase` rather than `case a => ...` (and go back to `addPart`, I think)

shared/src/vyxal/Elements.scala
```diff
@@ -2516,12 +2520,16 @@ object Elements:
...
-      true,
       "a: num -> [0..a)",
+      "a: lst[num] -> apl-style iota from 0 until a",
       "a: str -> a.lower()",
     ) {
       case a: VNum => NumberHelpers.range(0, a - 1)
-      case a: String => a.toLowerCase
+      case a: VList if a.forall(_.isInstanceOf[VNum]) =>
+        NumberHelpers.range(0, a.map(_.asInstanceOf[VNum] - 1))
+      case a => (Monad.vectorise("ʀ") {
```
ysthakur: Same here, no need to manually vectorise

shared/src/vyxal/Elements.scala
```diff
@@ -3173,14 +3181,18 @@ object Elements:
...
+      "a: lst[num] -> apl-style iota from 0 to a",
       "a: str -> is a lowercase?",
     ) {
       case a: VNum => NumberHelpers.range(0, a)
-      case a: String =>
-        if a.length == 1 then a.forall(_.isLower)
-        else VList.from(a.map(x => VNum(x.isLower)))
+      case a: VList if a.forall(_.isInstanceOf[VNum]) =>
+        NumberHelpers.range(0, a.map(_.asInstanceOf[VNum]))
+      case a => (Monad.vectorise("ʀ") {
```
ysthakur: asi;jdfpw8yq3awsejd;asdlfasdfasfd

shared/src/vyxal/Elements.scala
```diff
@@ -3941,8 +3953,7 @@ object Elements:
           case n: VNum => n
           case other => throw BadArgumentException("reshape", other)
```
ysthakur: This might possibly make the user think that `other` was the sole argument to reshape. Maybe we should have a more specific exception, like `s"bad argument to reshape, shape must be a number but got $shape"`? Not important, though, and also unrelated to this issue

shared/src/vyxal/NumberHelpers.scala
```diff
@@ -257,6 +257,12 @@ object NumberHelpers:
     val step = if start < end then 1 else -1
     start.to(end, step = step)
 
+  def range(start: VNum, ends: Seq[VNum])(using Context): VList =
+    val ranges = ends.map(start.to(_))
+    ListHelpers
+      .reshape(ListHelpers.cartesianProductMulti(ranges), ranges.map(_.length))
+      .asInstanceOf[VList]
```
ysthakur, line 265: Is `.asInstanceOf` safe to use here? You could either make `reshape`'s return type `VList` or wrap this in `VList.from`:
```suggestion
    VList.from(
      ListHelpers
        .reshape(ListHelpers.cartesianProductMulti(ranges), ranges.map(_.length))
    )
```

shared/test/resources/tests.yaml
```diff
@@ -1182,11 +1182,15 @@
 
 "ɾ":
   - {in: [10], out: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]}
+  - {in: [[3, 2]], out: [[[1, 1], [1, 2]], [[2, 1], [2, 2]], [[3, 1], [3, 2]]]}
```
ysthakur, line 1186: I'd like to see an example like `in: [["foo", 10, [2, 2]]]` to make sure the vectorisation works properly

ysthakur (changes requested):
DENIED

shared/src/vyxal/NumberHelpers.scala
RubenVerg: Reshape returns a scalar (not a `VList`) if the shape is an empty list. I suppose this might not be safe if range is called with an empty list but that could just be an error (though I guess a more clear error might be useful)

RubenVerg: alright, I added an error message, now `.asInstanceOf[VList]` is safe

ysthakur (approved):

PR #2092, opened 2024-03-05
Make Set Intersection work on strings

RubenVerg:


ysthakur (approved):
LGTM, but how about making it work for `(String, VList)` too?

lyxal (approved):

Steffan153 (approved):

PR #2093, opened 2024-03-19
Implement letter to number digraph

mathcat4:
title.
I'm not that good at scala, so parts of the code might not be good scala practices. (does scala have a PIP or smth)


RubenVerg (changes requested):
I would not use "ASCII magic" for both computations and use indexing into/from strings instead. Personally I think this sort of ASCII-based calculations makes the code much less clear to a reader.

shared/src/vyxal/Elements.scala
```diff
@@ -4049,6 +4049,31 @@ object Elements:
...
+        "letter-number-swap",
+        "number-letter-swap",
+        "a1-swap",
+      ),
+      true,
+      "a: str -> the index of a in the alphabet (one-indexed)",
+      "a: int -> the a-th letter of the alphabet (one-indexed)",
+    ) {
+      case a: VNum =>
+        "abcdefghijklmnopqrstuvwxyz" (math.floorMod((a - 1).toInt, 26)).toString
```
ysthakur: I'd prefer converting to an `Int` at the end, in case the number is too big to fit inside an `Int`. Maybe something like this?
```suggestion
        "abcdefghijklmnopqrstuvwxyz" (((a - 1).floor % 26).toInt).toString
```
Also, `<string literal> (<arg>).<method>` is kinda hard to read. I'd suggest using `.charAt` because that feels clearer than `.apply`, or better, extract `"abc...xyz"` to a `val` inside `StringHelpers`

shared/src/vyxal/Elements.scala
```diff
@@ -4049,6 +4049,31 @@ object Elements:
...
+      case a: String if a.length == 1 =>
+        VNum("abcdefghijklmnopqrstuvwxyz".indexOf(a.head.toLower) + 1)
+      case a: String => VList.from(
+          a.map(char =>
+            VNum("abcdefghijklmnopqrstuvwxyz".indexOf(char.toLower) + 1)
+          ).toList
+        )
```
ysthakur: There's a teeny bit of code duplication here. Maybe try something like this?
```suggestion
      case a: String =>
        val inds =
          a.map(char =>
            VNum("abcdefghijklmnopqrstuvwxyz".indexOf(char.toLower) + 1)
          )
        if inds.length == 1 then inds.head else VList.from(inds)
```

shared/test/resources/tests.yaml
```diff
@@ -1808,3 +1808,10 @@
 "ÞỊ":
   - {in: [[0, 1, 0, 1]], out: [1, 3] }
   - {in: [5], out: [0]}
+
+"øA":
+  - {in: [7], out: "g"}
+  - {in: [-69], out: "i"}
+  - {in: ["Y"], out: 25}
+  - {in: ["vyxal"], out: [22, 25, 24, 1, 12]}
+  - {in: [["xyz", 9, "h"]], out: [[24, 25, 26], "i", 8]}
```
ysthakur: Would be nice to test the empty string

shared/test/resources/tests.yaml
```diff
@@ -1808,3 +1808,10 @@
 "ÞỊ":
   - {in: [[0, 1, 0, 1]], out: [1, 3] }
   - {in: [5], out: [0]}
+
+"øA":
+  - {in: [7], out: "g"}
```
ysthakur, line 1813: Would be nice to also test a number greater than 2^32, i.e., something that doesn't fit in `Int`.

ysthakur (changes requested):

ysthakur:
@mathcat4 Thanks for the PR. If you want a style guide, [here's](https://docs.scala-lang.org/style/) one, but your code already looks pretty good to me overall

shared/src/vyxal/Elements.scala
```diff
@@ -4049,6 +4049,29 @@ object Elements:
...
+        "letter-number-swap",
+        "number-letter-swap",
+        "a1-swap",
+      ),
+      true,
+      "a: str -> the index of a in the alphabet (one-indexed)",
+      "a: int -> the a-th letter of the alphabet (one-indexed)",
+    ) {
+      case a: VNum =>
+        "abcdefghijklmnopqrstuvwxyz" (math.floorMod((a - 1).toInt, 26)).toString
```
lyxal:
```scala
MiscHelpers.index("abc...xyz", a - 1)
```

lyxal (changes requested):

mathcat4:
Testing for a very large number was taking forever to run, so I added a %26. I don't know why it was so slow, I'm guessing it's because of something in "ListHelpers".

shared/src/vyxal/Elements.scala
```diff
@@ -4049,6 +4049,29 @@ object Elements:
...
+        "letter-number-swap",
+        "number-letter-swap",
+        "a1-swap",
+      ),
+      true,
+      "a: str -> the index of a in the alphabet (one-indexed)",
+      "a: int -> the a-th letter of the alphabet (one-indexed)",
+    ) {
+      case a: VNum =>
+        MiscHelpers.index("abcdefghijklmnopqrstuvwxyz", (a - 1) % 26)
```
ysthakur: If you're using %, you no longer need MiscHelpers.index. just use String#charAt

ysthakur (commented):
Could you also make abcd...xyz a value inside stringhelpers instrad of typing it out?

shared/src/vyxal/Elements.scala
ysthakur: Although % might not handle negative numbers right. I think there mught be a different method on VNum that does actual modulo

PR #2094, opened 2024-03-20
fix grid neighbor diagonal issues

pacmanboss256:


ysthakur:
<s>Unrelated to this PR, but could you make `neighbourDirections` a `val` instead of a `def`. As far as I can tell, making it a method is unnecessary</s>

Edit: Actually, I'd like to see if I can refactor `neighbourDirections` entirely, making it curried. I'll make that a separate PR

ysthakur (approved):
LGTM, thanks for the fix

PR #2103, opened 2024-05-24
v3: Implement complex parts (∆ṗ) and polar parts (∆Ṗ)

ysthakur:
Implemented ∆ṗ for getting real and imaginary parts of a complex number and ∆Ṗ for getting polar coordinates of a complex number (radius and angle (radians))


shared/src/vyxal/Elements.scala
RubenVerg, line 1: `VNum` has an `arg` method, no need to use atan2

shared/test/resources/tests.yaml
RubenVerg, line 1: What does `W` next to the element name do?

RubenVerg (changes requested):

shared/test/resources/tests.yaml
ysthakur: It's to wrap the stack so we can check the real part/absolute value. Is there a less hacky way to do it?

RubenVerg: ah, i see. yeah i guess it makes sense, i don't think there's a better way with the way tests are structured

shared/src/vyxal/Elements.scala
ysthakur: Ah nice, didn't see that

lyxal (approved):

PR #2106, opened 2024-05-26
v3: Don't use reflection to generate stuff

### LINKED ISSUE #2105: Deploy workflow is failing
The deploy workflow is [failing](https://github.com/Vyxal/Vyxal/actions/runs/9226998122/job/25387998078) with an error saying that `upickle.default.write` can't be found.

I suspect it's because of https://github.com/Vyxal/Vyxal/pull/2101, where @scala-steward updated the Scala Toolkit version (@scala-steward, don't worry, you're a good bot and we still love you. You could'n't've foreseen this). Both Mill and Scala Toolkit depend on upickle, and maybe the minor update to Scala Toolkit made it so that it now has a different version of upickle, causing weirdness.

ysthakur:
Fixes #2105

Here's a similar issue someone else had: https://github.com/com-lihaoyi/mill/issues/2985. I tried using the fix given there, but it seems to refer to stuff we don't have, or I guess I didn't understand it? Anyway, I gave up on that.

Then I did some reflection, after which I decided I didn't much like reflection. So I made it so that `GenerateDocs` and friends are main methods that can be run from build.sc. Also, I used `T {}` instead of `T.sources`, which is probably unidiomatic but I think we're already not using `T.sources` the way it's supposed to be used (I think we're supposed to put files in the directory Mill tells us to put them).

In the process, I merged `GenerateTheseusData` and `GenerateKeyboard` into `GenerateTheseus`


lyxal (commented):
Does this generate the same style of parsed yaml for the keyboard?

ysthakur:
@lyxal Yes, I did just push a commit that makes the `GenerateTheseus.scala` more idiomatic, but I didn't change the actual logic at all

ysthakur:
Oh, I think the diff may be because I used a HashMap instead of a mutable.Map in one place

lyxal (approved):

PR #2141, opened 2025-01-14
Codepage Shuffle v3.5 edition

lyxal:
Turns out I was 5 codepage slots over the 256 limit. So:

- Remove `reduce lambda` to save a codepage slot (I was planning on removing it at some point anyway)
- Remove `empty string` constant (was `⑨`) to save another codepage slot
- Demote `✒` (`print without newline`) to be a digraph (`#,`) and remove the monograph to save a third codepage slot
- Demote `ᖶ` (`if-else modifier`) to be a digraph modifier (`#⍰`) and remove the monograph for saving no. 4
- Remove `⎃` (`flatten and join on empty string`) to save the final needed codepage slot.

The missing characters were `⊞γ‹›Ṫ`


lyxal:
Also, adds `×÷~` to codepage

PR #2154, opened 2025-05-19
Fix link in README.md

froth:
This was treated as a relative link and now points to the right place


lyxal:
Hi there hello,

Thanks for the link fix, always good to catch stray links like that.

PR #2164, opened 2025-07-17
Improve all of the keywords

pacmanboss256:


shared/src/vyxal/elements/ElementInformation.scala
```diff
@@ -2403,7 +2416,12 @@ object ElementInformation:
     ),
     AddElement(
       symbol = "•",
-      keywords = Seq("dot-product", "bijective-base", "first-predicate-index"),
+      keywords = Seq(
+        "dot-product",
+        "string-repat",
```
themoonisacheese: typo, not surfaced as dupe of `string-repeat`. suggest `string-repeat-concat`?

pacmanboss256: fixed, that name is a bit awkward but there isn't a better option which is not way too long

themoonisacheese (approved):

pacmanboss256:
if this gets merged the version number should be updated to 3.8.1

pacmanboss256:
merging this once i update it for the new stuff and make a few adjustments

gingershaped (approved):

PR #2168, opened 2025-07-27
add some k constants

pacmanboss256:


pacmanboss256:
oh it also fixed `kD`

PR #2170, opened 2025-07-31
change roll to shift

pacmanboss256:


lyxal (approved):
Huh, this really was just a name change

PR #2194, opened 2025-08-26
Convert register to a hybrid with v2 global array

pacmanboss256:
Register now acts as a hybrid between the current form and a global array. 

Breaking changes include `` ` `` changing from stack length to register peek, and adding `` #` `` for stack length. Register is completely overhauled, with `£` now not only pushing to the register, but takes a function and applies it to the first element. Might add a digraph to apply a function to all elements, but thats less important. `R` flag can be set to peek the stack when setting the register rather than popping.


shared/src/vyxal/elements/ElementInformation.scala
```diff
@@ -1814,13 +1814,18 @@ object ElementInformation:
...
       Options(),
       Overload(
         name = "Set Register",
-        args = Seq("any"),
-        description = "Set the register to #1",
+        args = Seq("nsl"),
+        description = "Push #1 to the register",
+      ),
+       Overload(
+        name = "Apply to Register",
```
lyxal: Don't forget to update this

ysthakur:
> `R` flag can be set to peek the stack when setting the register rather than popping.

I'm not a huge fan of using a flag to change behavior from popping to peeking. This *is* a golflang, but Vyxal is supposed to be readable too, and you should be able to look at a piece of code and know what it does without having to look at the flags too. Flags should be more for things like formatting.

Also, Vyxal already has a reputation for being cheaty with its flags, I don't want to make that worse

---

Other changes sound good.

shared/src/vyxal/RegisterHelpers.scala
ysthakur, line 1: Huh, is the Scalafmt workflow not working? It should be enforcing 2 spaces rather than tabs (I also saw a stray space in `Globals.scala`)

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,64 @@
+package vyxal
+
+import vyxal.conversions.{*, given}
+
+import scala.collection.mutable.ArrayBuffer
+
+
+object RegisterHelpers:
+
+	val register = Context().globals.register
```
ysthakur: I guess the register is shared by all contexts, but I don't think this is what you want.

This will create a whole new context and get the register from that. That context and that register, as far as I can tell, will be completely unrelated to any contexts created for the user's actual program. On top of that, because this is a static variable, that `ArrayBuffer` will be used by all contexts in all programs that the interpreter executes.

(please correct me if I'm wrong, it's been a long time since I last looked at this repository so I've forgotten how things work)

I think what you want is something like this:
```scala
def register(using ctx: Context): ArrayBuffer[VAny] = ctx.globals.register
def clear(using Context) = register.length == 0
def length(using Context) = VNum(register.length)
```

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,64 @@
...
+		then
+			val c = Interpreter.executeFn(fn, args = Seq(register(idx.toInt)))
+			register.update(idx.toInt, c)
+
+
+	/** Map a function over the entire register */
+	def map(fn: VFun)(using ctx: Context): Unit =
+		for x <- 0 to register.length - 1 do
+			if register(x).isInstanceOf[VPhysical] then applyFn(fn, VNum(x)) 
+			else register(x)
```
ysthakur: The doc comment says one thing, but the function body is doing something else entirely. Shouldn't it be something like this?
```suggestion
	/** Map a function over the entire register */
	def map(fn: VFun)(using ctx: Context): Unit =
		register.forEach(x => applyFn(fn, x))
```
Although I would say this is *applying* a function to the entire register, not *mapping* a function over it. I'm assuming it's more useful to not only apply a function with side effects to the register but to actually modify the register. For that, you could use [`register.mapInPlace(applyFn(fn, _))`](https://www.scala-lang.org/api/current/scala/collection/mutable/ArrayBuffer.html#mapInPlace-fffff965)

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,64 @@
...
+	def length: VNum = VNum(register.length)
+
+	def reverseRegister(): Unit = 
+		val rev = register.reverse
+		for i <- 0 to register.length - 1 do
+			register.update(i, rev(i))
+		
+	/** Apply a monadic function to an item in the register */	
+	def applyFn(fn: VFun, idx: VNum)(using ctx: Context): Unit =
+		if register.last.isInstanceOf[VPhysical]
```
ysthakur: Why only apply functions to non-functions? Applying a higher-order function to a function stored in the register seems like something users might need

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,64 @@
...
+	def clear(): Unit = register.clearAndShrink(1)
+
+	def isEmpty: Boolean = register.length == 0
+
+	def length: VNum = VNum(register.length)
+
+	def reverseRegister(): Unit = 
+		val rev = register.reverse
+		for i <- 0 to register.length - 1 do
+			register.update(i, rev(i))
```
ysthakur: Nitpick: I don't know why, but creating a new ArrayBuffer just to copy from it annoys me a little. I'd prefer something like this:
```suggestion
		for i <- 0 until register.length / 2 do
			val temp = register(i)
			register(i) = register(register.length - 1 - i)
			register(register.length - 1 - i) = temp
```
This is mostly a style thing, so feel free to ignore the suggestion

---

However, if you do ignore the above suggestion, do consider this one: Use `0 until register.length` instead of `0 to register.length - 1` (they do the same thing). It's just a style thing but I think it's reasonable to prefer `until` here

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,64 @@
...
+			if register(x).isInstanceOf[VPhysical] then applyFn(fn, VNum(x)) 
+			else register(x)
+
+
+
+	def push(a: VAny)(using Context): Unit =
+		register.append(a)
+
+	def pop(j: VNum = 1, allVals: Boolean = false, peek: Boolean = false)(using ctx: Context): VAny =
+		val n = if allVals then register.length else j.toInt
```
ysthakur: Given that `allVals` can kinda "overwrite" `j`, I think I would prefer it if we removed the `allVals` parameter from this method entirely. Then we could make a different method for just the `allVals = true` case:

```scala
def popAll(peek: Boolean = false)(using Context): VAny = pop(register.length, peek = peek)
```

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,64 @@
...
+		if register.isEmpty then VNum(0)
+		else
+			val top = register.takeRight(n)
+			if !peek then 
+				register.dropRightInPlace(n)
+				register.trimToSize()
```
ysthakur: Do you have reason to believe that leaving the register at its old capacity would be a problem? If not, I don't think there's any reason to decrease capacity. If anything, it could slightly decrease performance if someone's constantly popping items from and pushing items onto the register.

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,64 @@
...
+			if top.length == 1 && !allVals then 
+				val c = top.head
+				c match
+					case c: VFun => Interpreter.executeFn(c)
```
ysthakur: Is this useful behavior? Would someone want to store functions onto the register and pop them off without executing them?

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,64 @@
...
+					case _ => c
+			else top.toList.filterNot(_.isInstanceOf[VFun])
```
ysthakur: Why filter out the functions? Users might want to get their functions off the register

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,64 @@
...
+	def pop(j: VNum = 1, allVals: Boolean = false, peek: Boolean = false)(using ctx: Context): VAny =
+		val n = if allVals then register.length else j.toInt
+		if register.isEmpty then VNum(0)
+		else
+			val top = register.takeRight(n)
+			if !peek then 
+				register.dropRightInPlace(n)
+				register.trimToSize()
+			if top.length == 1 && !allVals then 
+				val c = top.head
```
ysthakur: This is just a style thing, but `c` isn't a very descriptive variable name, and you could probably get away with not even making a variable and just using `top.head` directly below instead. `top.head` isn't a heavy operation, it's just indexing into an array.

shared/src/vyxal/elements/ElementInformation.scala
```diff
@@ -1844,7 +1844,118 @@ object ElementInformation:
...
+    ),
+    AddElement(
+      symbol = "Þ⍨",
+      keywords = Seq("register-dump", "dump-register"),
+      arity = 0,
+      Options(),
+      Overload(
+        name = "Dump Register",
+        args = Seq(),
+        description = "Pop each item in the register to the stack",
```
ysthakur: Is the order documented somewhere? It's fine if it's not documented right here, but it should be written down somewhere how the register works.

Does the first element of the register get popped and pushed onto the stack first, or does the last element get pushed first?

shared/src/vyxal/elements/ElementInformation.scala
```diff
@@ -1844,7 +1844,118 @@ object ElementInformation:
...
+      ),
+    ),
+    AddElement(
+      symbol = "Þ¥",
+      keywords = Seq("pop-all-register", "wrap-register-pop"),
+      arity = 0,
+      Options(),
+      Overload(
+        name = "Pop Register as list",
```
ysthakur, line 1878: Nitpick: the fact that this is named `Pop register as list` makes it seem like the dual of `Push register as list` below (which it's not). I really don't know of a better name, but it would help to be more explicit about the fact that this is the same as the other command, except this one clears the register after wrapping in a list.

shared/src/vyxal/elements/ElementInformation.scala
```diff
@@ -1844,7 +1844,118 @@ object ElementInformation:
...
+    ),
+    AddElement(
+      symbol = "Þ`",
+      keywords = Seq("register-length", "reg-len"),
+      arity = 0,
+      Options(),
+      Overload(
+        name = "Register Length",
+        args = Seq(),
+        description = "Push the length of the register to the stack",
```
ysthakur: I think you can just say "Length of the register." It's a stack-based language, so it's implied that it pushes to the stack

shared/src/vyxal/elements/ElementInformation.scala
```diff
@@ -1844,7 +1844,118 @@ object ElementInformation:
...
+        description = "Get item in the register at index #1",
+      ),
+    ),
+    AddElement(
+      symbol = "Þ⊖",
+      keywords = Seq("reg-pop-n", "register-pop-n"),
+      arity = 1,
+      Options(),
+      Overload(
+        name = "Pop n from Register",
```
ysthakur, line 1944: Same comment here, the name makes it feel like a dual of `Push n from Register` even though it's not

shared/src/vyxal/elements/ElementInformation.scala
```diff
@@ -3082,6 +3198,11 @@ object ElementInformation:
         args = Seq("lst[num|str]"),
         description = "Flattens each item in #1",
       ),
+      Overload(
+        name = "Register Map",
+        args = Seq("fun"),
+        description = "Map #1 over the register",
```
ysthakur: Is this supposed to just apply the function to every element in the register without modifying the register itself? Or is it supposed to do `register = register.map(#1)` (replace every element in the register by applying the function to it)?

shared/src/vyxal/elements/Elements.scala
```diff
@@ -739,14 +739,60 @@ object Elements:
...
-        summon[Context].globals.register = pop()
+      direct(1) {
+        val a = pop()
+        RegisterHelpers.push(a)
+        if summon[Context].settings.registerPeek then push(a)
+      },
+    "¥" -> 
+      direct(0) {
+        push(RegisterHelpers.pop(peek=true))
+      },
```
ysthakur: This is just a style thing, but I'd prefer to use `direct` as little as possible. iirc if you have an element defined like `direct(0) { push(foo) }`, you can use `niladify { ctx ?=> push(foo) }`. This and a lot of other elements below could use `niladify` instead

ysthakur: One thing about `niladify` that's annoying is that there's two overloads, one that takes a `VAny` and one that takes a `Context ?=> VAny`. And I think without `ctx ?=>`, the compiler can't tell you mean the second overload.

To fix that, you could name the two overloads different things. The first could be `constant` or something, and the second overload could remain `niladify`

shared/src/vyxal/elements/Elements.scala
```diff
@@ -1151,9 +1197,15 @@ object Elements:
...
-      case lst: VList => VList(lst.map(item => ListHelpers.sum(item.itr)))
+    "Ͼ" ->
+      direct(1) {
+        val top = pop()
+        top match
+          case num: VNum => push(" " * num.toInt) // hallelujah
+          case lst: VList => push(VList(lst.map(item => ListHelpers.sum(item.itr))))
+          case f: VFun if f.arity == 1 =>
+            RegisterHelpers.applyFn(f, RegisterHelpers.length - 1)
+          case _ => throw UnsupportedOverloadException("Ͼ", "Function arity !=1")
```
ysthakur: This will also throw an exception if it gets a string, in which case the function arity message doesn't make sense

shared/src/vyxal/elements/Elements.scala
```diff
@@ -1151,9 +1197,15 @@ object Elements:
...
-      case num: VNum => " " * num.toInt // hallelujah
-      case lst: VList => VList(lst.map(item => ListHelpers.sum(item.itr)))
+    "Ͼ" ->
+      direct(1) {
+        val top = pop()
+        top match
+          case num: VNum => push(" " * num.toInt) // hallelujah
+          case lst: VList => push(VList(lst.map(item => ListHelpers.sum(item.itr))))
+          case f: VFun if f.arity == 1 =>
+            RegisterHelpers.applyFn(f, RegisterHelpers.length - 1)
```
ysthakur: We don't want to push the result of the function onto the stack?

shared/test/src/vyxal/RegisterTests.scala
```diff
@@ -0,0 +1,93 @@
...
+      )
+      it("Should push zero if empty") {
+        RegisterHelpers.clear()
+        given ctx: Context = Context(testMode = true)
+        ctx.push(5)
+        Interpreter.execute("£")
+        val checkpoint = Checkpoint()
+        checkpoint { assert(ctx.isStackEmpty) }
+        Interpreter.execute("``")
+        checkpoint { assertResult(Seq[VAny](0,5))(Seq(ctx.pop(), ctx.pop())) }
```
ysthakur: These could always be 2 different tests

ysthakur (commented):
It's been a while since I've contributed to this codebase or golfed, so forgive me if my comments are completely off.

Also, I only skimmed the tests, did not review them thoroughly at all

shared/src/vyxal/RegisterHelpers.scala
pacmanboss256: pop can take some number n which is implemented as a digraph elsewhere to pop n args

shared/src/vyxal/RegisterHelpers.scala
pacmanboss256: i don't know how register indexing works if theres an empty slot

shared/src/vyxal/elements/Elements.scala
pacmanboss256: this is a port of `ᴿ` from v3 legacy

shared/src/vyxal/RegisterHelpers.scala
ysthakur: I'm aware. I'm saying you should keep this method but remove `allVals` from it, and then create a new method that's specifically for popping/peeking all values

shared/src/vyxal/RegisterHelpers.scala
ysthakur: What do you mean by empty slot?

shared/src/vyxal/elements/Elements.scala
ysthakur: Okay, I'll assume there was a good reason for that

shared/src/vyxal/elements/Elements.scala
```diff
@@ -748,39 +748,14 @@ object Elements:
...
-      },
-    "Þ`" ->
-      direct(0) {
-        push(RegisterHelpers.length)
-      },
-    "Þ_" ->
-      direct(0) {
-        RegisterHelpers.clear()
-      },
+    "¥" -> niladify(ctx ?=> RegisterHelpers.pop(peek=true)),
```
ysthakur: Are you able to do this now that `niladify` doesn't have a second overload now, btw? Or is the compiler too dumb to understand?
```suggestion
    "¥" -> niladify { RegisterHelpers.pop(peek=true) },
```

shared/src/vyxal/RegisterHelpers.scala
pacmanboss256: like would indexing be messed up if I pop something and reverse it?

shared/src/vyxal/RegisterHelpers.scala
```diff
@@ -0,0 +1,57 @@
...
+		if register.last.isInstanceOf[VPhysical]
+		then
+			val c = Interpreter.executeFn(fn, args = Seq(register(idx.toInt)))
+			register.update(idx.toInt, c)
+
+
+	/** Apply a function to all elements in the register */
+	def map(fn: VFun)(using ctx: Context): Unit =
+		for x <- 0 until register.length do
+			applyFn(fn, VNum(x))
```
ysthakur: This is now applying the function to indices, not elements. Did you mean
```suggestion
		for x <- register do
			applyFn(fn, x)
```
?

shared/src/vyxal/RegisterHelpers.scala
ysthakur: I don't see why it would. `trimToSize` affects the *capacity*, not the *size* of the `ArrayBuffer`. When you do `register.trimToSize()`, it allocates a new array of length `register.size`, then copies all the elements to that new array. This has absolutely no effect on the *size* of the register. The user has no clue that `register.trimToSize()` was called to decrease the capacity of the register. It only changes the length of the array backing the `ArrayBuffer`. So it's totally fine if you don't mess with the capacity of an `ArrayBuffer` after operations like `dropRightInPlace`

shared/src/vyxal/RegisterHelpers.scala
pacmanboss256: applyFn takes an an index

pacmanboss256:
> > `R` flag can be set to peek the stack when setting the register rather than popping.
> 
> I'm not a huge fan of using a flag to change behavior from popping to peeking. This _is_ a golflang, but Vyxal is supposed to be readable too, and you should be able to look at a piece of code and know what it does without having to look at the flags too. Flags should be more for things like formatting.
> 
> Also, Vyxal already has a reputation for being cheaty with its flags, I don't want to make that worse
> 
> Other changes sound good.

This exclusively changes the behavior of `£` to not consume whatever it pops.

shared/src/vyxal/RegisterHelpers.scala
ysthakur: Oh, I'm such an idiot

ysthakur (approved):
Thanks for quickly updating based on feedback, LGTM now. Could you just run the formatter before pushing?

lyxal (approved):

PR #2198, opened 2025-09-01
Implement Nullads

pacmanboss256:


shared/src/vyxal/Context.scala
```diff
@@ -57,6 +57,7 @@ class Context private (
     */
   def pop(): VAny =
     if useStack && parent.isDefined then return parent.getOrElse(this).pop()
```
ysthakur: Hey wait a sec, this should be `parent.get.pop()`. `parent.isDefined` ensures that we have a parent (also, calling `this.pop()` here would result in an infinite loop, so that's the wrong thing to do anyway). Totally unrelated to this PR though

shared/src/vyxal/Context.scala
```diff
@@ -57,6 +57,7 @@ class Context private (
...
+    stack.filterInPlace(!_.isInstanceOf[VNull])
```
ysthakur: Why do this when peeking and popping? Why not only filter out `VNull`s when pushing?

shared/src/vyxal/Context.scala
```diff
@@ -105,7 +107,9 @@ class Context private (
     else stack.toList.reverse ::: globals.inputs.peek(n - stack.length)
 
   /** Push items onto the stack. The first argument will be pushed first. */
-  def push(items: VAny*): Unit =
+  def push(itms: VAny*): Unit =
+    val items = itms.filterNot(_.isInstanceOf[VNull])
+    if items.isEmpty || items.isInstanceOf[VNull] then return
```
ysthakur: `items.isInstanceOf[VNull]` will never be true. `items` is a `Seq[VAny]`. You're already filtering out nulls when creating `items`

```suggestion
    if items.isEmpty then return
```

ysthakur: Honestly you don't even need to bother with this early return behavior, unless you have reason to believe this optimization will bring about a noticeable performance boost. I'm fine either way though

shared/src/vyxal/VAny.scala
```diff
@@ -110,6 +115,10 @@ object conversions:
   given Conversion[Boolean, VNum] = b => if b then 1 else 0
 end conversions
 
+case class VNull(n: VAny) extends VAny
```
ysthakur: Can we remove `VNull` now that you have `NullMonad/Dyad`?

pacmanboss256: sure

PR #2201, opened 2025-09-03
Use parent.get over parent.getOrElse(this)

ysthakur:
A few methods in `Context.scala` had code like this:
```scala
def foo() =
  if useStack && parent.isDefined then
    return parent.getOrElse(this).superMethod()
```

The `getOrElse` was useless, because `parent.isDefined` already made sure that `parent.get` would return a value.

There was also one method that didn't have the `parent.isDefined` check, meaning that it could have gone into an infinite loop if `useStack` were true.

### DIFF
--- a/shared/src/vyxal/Context.scala
+++ b/shared/src/vyxal/Context.scala
@@ -56,7 +56,7 @@ class Context private (
     * inputs, read a line of input from stdin.
     */
   def pop(): VAny =
-    if useStack && parent.isDefined then return parent.getOrElse(this).pop()
+    if useStack && parent.isDefined then return parent.get.pop()
     val elem: VAny =
       if stack.nonEmpty then stack.remove(stack.size - 1)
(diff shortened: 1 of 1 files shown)


ysthakur:
The Scalafix workflow is failing because of a thing on main, you can ignore that

shared/src/vyxal/Context.scala
```diff
@@ -79,12 +79,12 @@ class Context private (
     * start of the list.
     */
   def pop(n: Int): Seq[VAny] =
-    if useStack then return parent.getOrElse(this).pop(n)
+    if useStack && parent.isDefined then return parent.get.pop(n)
```
ysthakur, line 82: You might say that this is a change that should be tested, especially since we don't even know if an infinite loop is actually possible here (no one's reported such a bug). However, I would respond that I am tired and could not care less

lyxal (approved):

PR #2211, opened 2025-09-21
Add a parameter to explicitly print the version

Yewzir:
This flag is functional in most applications and is often useful for external scripts that explicitly request versions.


pacmanboss256:
`v` already exists as a flag, will need to use --version and also make it an offline only flag because the interpreter gives you the version anyways

PR #2218, opened 2025-11-28
Round to arbitrary number of decimal places

themoonisacheese:
ℳ can round to decimal places, but always returns 0 if #2 >9. this creates a fallback where the string representation of #1 will be returned instead, helping for challenges such as "print 1000 digits of pi"

image


lyxal (approved):
Based

PR #2230, opened 2026-07-02
Add cantor list collapsing thingy

### LINKED ISSUE #2228: Add Cantor Pairing Function / Method of encoding list of integers into single integer
**Describe the feature**

The [Cantor Pairing function](https://en.wikipedia.org/wiki/Pairing_function#Cantor_pairing_function) is a function that takes two natural `x,y` and returns `(x+y+1)*(x+y)/2 + y`. This always returns a unique natural number for each possible value of `x` and `y`. This can also be inverted, and if this is implemented, the inverse also should be implemented.

**Details**

This can also be used to encode a list of natural numbers into a natural number, by first placing the length of the list at the start, and then iteratively applying the function on the first two elements of the list (and then adding the result back on to the list) until you get the single number. The reason why the length needs to be added is because said number could theoretically encode any form of list.

This results in smaller numbers than using prime factorization in most cases, making it smaller if you want to compress the number.

[This person](https://codegolf.stackexchange.com/questions/267048/tips-for-golfing-in-vyxal-3#comment584967_267052) wanted to compress lists of numbers (and strings as well.)

PlaceReporter99:
May possibly close #2228


shared/src/vyxal/elements/Elements.scala
```diff
@@ -2153,31 +2153,28 @@ object Elements:
...
-    addPart("øA", Monad, true) {
-      case a: VNum =>
-        "abcdefghijklmnopqrstuvwxyz".charAt(((a - 1) % 26).toInt).toString
-      case VStr(a) =>
-        val inds = a.map(char =>
-          VNum("abcdefghijklmnopqrstuvwxyz".indexOf(char.toLower) + 1)
-        )
-        if inds.length == 1 then inds.head else VList(inds)
-    },
+      addPart ("øA", Monad, true) {
```
PlaceReporter99: What!?! Whitespace!?!

PlaceReporter99:
I need to fix that

PlaceReporter99:
```
|The following import might make progress towards fixing the problem:
Error:       |
Error:       |  import sourcecode.Text.generate
```
“You’re so bad I’ll try and force you to vibe code it!”

PlaceReporter99:
We're definitely going to squash this into a single commit

PlaceReporter99:
Fair enough, i need to document it, but hey, it compiles!

PlaceReporter99:
This definitely needs to be squashed

lyxal (approved):

PR #2233, opened 2026-07-14
Implement SHA256 from scratch

PlaceReporter99:
Worked in python, so should work if everything is ported correctly


PlaceReporter99:
Let's go!

PR #2238, opened 2026-07-26
Implement VType, user thrown exceptions, and better error handling utilities

PlaceReporter99:
Error info will now be a VObject containing the error data instead of a VStr. If you want to get the message as it was before, it is in the "message" field.


PlaceReporter99:
Next I will try and figure out how to get the context variable to hold the error. It will just be a string most likely though.

PlaceReporter99:
Weird, it's demanding input for some of the tests...

PlaceReporter99:
Gosh that was weird as hell

shared/src/vyxal/elements/ElementInformation.scala
```diff
@@ -347,7 +360,64 @@ object ElementInformation:
...
+        Overload(
+          name = "Throw exception",
+          args = Seq("any", "exc"),
+          description =
+            "Throw #2 while formatting the message string with contents of #1",
+        ),
+      ),
+      AddElement(
+        symbol = "#ꜝ",
+        keywords = Seq("assert", "error-if-falsy"),
```
themoonisacheese, line 409: we use `falsey` for describing 0 across this codebase, i'm pretty sure

shared/src/vyxal/elements/Elements.scala
```diff
@@ -1795,6 +1791,32 @@ object Elements:
         MiscHelpers.untilNoChange(predicate, initial).tail
       case (initial, predicate: VFun) =>
         MiscHelpers.untilNoChange(predicate, initial).tail
+      case (VListOf[VFun](fns), VListOf[VList](types)) =>
+        if types.forall((x: VAny) =>
+            x.asInstanceOf[VList].lst.forall((y: VAny) => y.isInstanceOf[VType])
```
themoonisacheese: lambda can be ` => _.isInstanceOf[VType]` which is cleaner. works the same way anyway.

shared/src/vyxal/elements/Elements.scala
```diff
@@ -1966,6 +1988,23 @@ object Elements:
       direct(0) {
         throw ContinueLoopException()
       },
+    "#q" ->
+      direct(2) {
+        val (b, a) = (pop(), pop())
```
themoonisacheese: assigning a,b is useless; they'll get reassigned to whatever names you use in your `case` statements. use `(pop(), pop()) match` instead

shared/src/vyxal/elements/Elements.scala
```diff
@@ -2042,19 +2081,38 @@ object Elements:
...
+                  Visibility.Restricted,
+                  VDuration(JDuration.ofMillis(t.timeElapsed)),
+                ),
+                "timeRemaining" -> (
+                  Visibility.Restricted,
+                  VDuration(JDuration.ofMillis(t.timeRemaining)),
+                ),
+                "isPaused" -> (
+                  Visibility.Restricted,
+                  (if t.paused then VNum(1) else VNum(0)),
```
themoonisacheese: replace with `VNum(t.paused)`

shared/src/vyxal/elements/Elements.scala
```diff
@@ -2042,19 +2081,38 @@ object Elements:
       push(a)
       a match
         case t: VTimer => push(
-            VList(
-              Seq(
-                VDuration(JDuration.ofMillis(t.timeElapsed)),
-                VDuration(JDuration.ofMillis(t.timeRemaining)),
-                if t.paused then VNum(1) else VNum(0),
-              )
+            VObject(
```
themoonisacheese, line 2093: any reason for this to be a custom object rather than a shallow list? i'm not saying it's not a good idea, but i'd like to hear arguments motivating this, as this is almost always less golfy.

shared/src/vyxal/elements/Elements.scala
```diff
@@ -2042,19 +2081,38 @@ object Elements:
...
       direct(Monad) {
         VDate.setDefaultZone(pop().asInstanceOf[VStr].s)
       },
+    addPart("#Y", Monad, false) {
+      case a => VType(a)
+    },
+    addPart("#ɦ", Monad, true) {
+      case VStr(a) => VType(
+          a
+        ) // We cannot overload #Y to convert a string to a type because #Y has to output the type of it (which would be VStr)
```
themoonisacheese, line 2123: then this function is only ever useful if you know what you pass it is a VStr, in which case you don't need the function at all ( i guess it's a golfier way of getting a VStr type...)

themoonisacheese (changes requested):

shared/src/vyxal/elements/Elements.scala
PlaceReporter99: Having a long list of values is a bit ambiguous as to what exactly they mean

shared/src/vyxal/elements/Elements.scala
PlaceReporter99: You forgot about a weird Scala quirk:
`case a: VStr` makes `a` a VStr
`case VStr(a)` makes `a` whatever was passed into VStr, in this case a String. VType was overloaded to convert this string to the type it names, for example, VType("VList") gets you the type VList.

shared/src/vyxal/elements/ElementInformation.scala
PlaceReporter99: We’ve barely used either, but “falsey” occured 0 times in the documentation MD file (the introduction one) and “falsy” occured once.

shared/src/vyxal/elements/Elements.scala
PlaceReporter99: Completed

shared/src/vyxal/elements/Elements.scala
PlaceReporter99: Completed

shared/src/vyxal/elements/Elements.scala
PlaceReporter99: Completed

shared/src/vyxal/elements/Elements.scala
themoonisacheese: that's a start, though this is a golfing language. "user friendly" isn't exactly high on the list of priorities, and it is notably much lowe than "can i access whether or not my timer is paused in like 3 bytes?". i'm fine with either, but it seems the motivation isn't really there when this is exactly what we document instead of hold the user's hand

themoonisacheese (approved):

shared/src/vyxal/elements/Elements.scala
PlaceReporter99: If a future update were to expose more timer data to programmers, the list might be re-ordered and something that previously would’ve gotten whether the timer is paused might get something else. With object names, you can add more and reorder them, and as long as you don’t change the names, you won’t break existing code.

themoonisacheese: up until today, this hasn't been a concern. the interpreter makes the language and if your submission works only in a specific version, that's still a valid submission. in fact, we even go to great lengths to keep old interpreter builds around and selectable.

what you're suggesting is simply put to trade bytes for guaranteed backwards compatibility, and the fact is that for a golflang, bytes saved are a higher priority than backwards compatibility. in a praclang, i would definitely agree that your way is better, but as it stands, this element would be entirely useless because accessing any part of it means spending upwards of 5 bytes.

another way for you to have your ccake and eat it, perhaps: allow for VObjects to be iterable, that way users interested in actually writing terse programs (100% of them) can use array access primitives whith the understanding that changes to the structure are breaking.

@lyxal i'd like your input on this.

PlaceReporter99: As it stands I don’t think an overload for VObject and number on the index element has been added so it could potentially index the corresponding variable’s value at that index.

PlaceReporter99: Or we could make it so that V gets a list of values and K gets a list of variable names. Only issue is what to do with private fields…

PlaceReporter99:
DO NOT MERGE as i have not implemented tests for those new overloads

PlaceReporter99:
Should be mergeable now

shared/src/vyxal/elements/Elements.scala
PlaceReporter99: Updated V and K, meaning getting timer info in the golfy way would now be just one extra byte, by adding V in the middle to get values.

PR #2239, opened 2026-08-06
Add øs (SVG polyline)

themoonisacheese:
Adds `øs` (keywords `svg`, `to-svg`, `svg-polyline`, `polyline`), a non-vectorising monad that turns coordinates into an SVG polyline with stroke black and fill none.

```
#[#[1|2#]|#[3|4#]#]øs    →  <polyline points="1,2 3,4" stroke="black" fill="none" />
#[1|2|3|4#]øs            →  <polyline points="1,2 3,4" stroke="black" fill="none" />
#[1 2j|3 4j#]øs          →  <polyline points="1,2 3,4" stroke="black" fill="none" />
```

## Overloads

**`lst`** — the coordinates are one of:

- a list of points, where a point is `[x, y]` or a lone complex number
- a flat list of real numbers in the order they show up in the SVG (`x, y, x, y, ...`); an odd length errors
- a flat list containing any complex number, in which case every number is its own point (real part x, imaginary part y), so a real number in that list sits on the x axis (`3` → `3,0`)

Pairing therefore only kicks in when every number in a flat list is real, which keeps the two flat forms unambiguous.

**`obj`** — coordinates come from the object's `points`, `coords`, or `coordinates` field (in that order), and the stroke and fill from its `stroke` and `fill` fields, defaulting to black and none:

```
record Poly => [[1,2],[3,4]] $points "red" $stroke "blue" $fill end `Poly` svg
→ <polyline points="1,2 3,4" stroke="red" fill="blue" />
```

Fields are read through `getObjectMember`, so access modifiers apply. An object with no coordinate field throws `FieldNotFoundException`. Attribute values are XML-escaped.

## Tests

12 cases in `tests.yaml` covering the point/flat/complex/empty list forms, plus `ElementTests` coverage of the object overload and the error cases (ragged points, mixed points and bare numbers, odd flat length, non-numeric and complex-in-a-pair coordinates, non-list/non-object input).

Full `jvm.test` passes apart from `Elements #O, #^, #v, #›` "unpausing and pausing should be idempotent", which is a pre-existing timing flake in the timer element from #2237 (`PT0.001S` vs `PT0S`) — it fails identically on a clean `version-3` checkout. `js.compile` and scalafmt are clean; `table.md` and `vyxal-lit.nanorc` are regenerated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


themoonisacheese:
failed test is unrelated to this, has to do with timers.

lyxal (approved):
You know what, why not?

